PR #2, opened 2020-08-16
Add links to Atom packages

alexfedoseev:


chenglou:
Hah! Does it actually work? I haven't tested on Atom

alexfedoseev:
@chenglou Auto-formatting works for me. Please, try it out to confirm.

P.S. You would need to either disable `language-reason` or set manually syntax to `Rescript` (`source.rescript`) for a file b/c both use the same `res` extensions (see https://github.com/reasonml-editor/language-reason/issues/11).

chenglou:
Thanks; I'll close this PR because I've renamed this repo to rescript-vscode since lang-server is a implementation detail that just happens to live here. I've added your link to the rescript documentation itself instead =)

PR #3, opened 2020-08-26
Move grammar to grammars folder

brnrdog:
I've been trying to add the TM grammar to the [github/linguist](https://github.com/github/linguist/pull/4929) but it fails when I point to this repository. While observing other grammars in linguist, I noticed they usually have a `grammars` or `syntaxes` folder. So I [created a repository](https://github.com/brnrdog/syntax-test) with the TM grammar inside a `grammars` folder, and this way I was able to add the grammar.

So I'm assuming by moving the TM grammar file to this folder github/linguist will be able to detect the grammar.


chenglou:
That's so weird... id this documented somewhere?

chenglou:
Thanks, done!

PR #4, opened 2020-08-29
Add MIT License

brnrdog:
I'm adding the same license as in [rescript-lang/rescript-sublime](https://github.com/rescript-lang/rescript-sublime). A license is required for [github/linguist](https://github.com/github/linguist/pull/4975) since we're using this repository for the grammar.


amiralies:
Wouldn't it better to put grammer in a separate repo?

PR #14, opened 2020-10-28
Hook up command from rescript-editor-support.

cristianoc:
Hook up `rescript-editor-support.exe` from https://github.com/rescript-lang/rescript-editor-support.
Adds support for jump-to-definition, hover, autocompletion.


cristianoc:
Currently vendoring `server/darwin/rescript-editor-support.exe `.

Issue #10, opened 2020-09-21
bs binaries paths are determined wrong in monorepos


alexfedoseev:
https://github.com/rescript-lang/rescript-vscode/blob/ba96a2fe1c603e73703bce7c25244d2508dcc334/server/src/server.ts#L291

E.g. this wouldn't work for monorepos that use yarn workspaces, b/c binaries are hoisted above the bsconfig's node_modules.

gaku-sei:
Second this, it doesn't work in monorepos 😞 

A solution in the short term could be to have a "bsb path"/"bsc path" settings maybe? It's something that could be useful anyway 😄

dan003400:
We as well use yarn workspaces and have lost formatting.

BlueHotDog:
Came here to open the same issue.
Any solution would help as we're really excited to get started using it. looks much more stable, fast. fun. you got me all excited!

gaku-sei:
@BlueHotDog you can already get the extension working with monorepos if you use this branch: https://github.com/rescript-lang/rescript-vscode/pull/16 it works fine so far 😄

dchambers:
I worked around this for the time being by adding this `package.json` config to the sub-project where I'm going to be using ReScript:

```
"workspaces": {
  "nohoist": ["bs-platform"]
},
```

and then running:

```
yarn --force
```

Alternatively you can do this globally by modifying the `workspaces` definition in your root `package.json` from something like this:

```
"workspaces": ["libs/*"],
```

to this:

```
"workspaces": {
  "packages": ["libs/*"],
  "nohoist": ["**/bs-platform"]
},
```

BlueHotDog:
We rather wait until this extension stabilizes. this issue is open for 4 months, with open Prs trying to fix this. can any maintain please help here?

dan003400:
Hey guys, check this out if you are using yarn workspaces, I have a PR in that has a current release flag that you can try for vim plug.

https://github.com/rescript-lang/vim-rescript/pull/19

IwanKaramazow:
We've fixed this and it's released =D 
See [forum](https://forum.rescript-lang.org/t/ann-rescript-vscode-1-0-4-and-vim-rescript-1-2-0-relased/877)  or the [release](https://github.com/rescript-lang/rescript-vscode/releases/tag/1.0.4)

PR #16, opened 2020-11-19
Tweak heuristics to find the project root and "build root"

IwanKaramazow:
The project root means "this is a folder containing node_modules/bs-platform/{platform}/bsc.exe".
This path needs to be correct because it drives the formatter.
The "build root" represents the nearest directory containing a "bsconfig.json" file. "bsconfig.json" can be used to locate build artefacts and diagnostics.

The previous heuristics seemed to fail in the case users where using yarn workspaces.
Yarn workspaces seem to have the following layout:
```
/root
  /node_modules
  - package.json
  - yarn.lock
  /folder1
    /src
       - MyApp.res
    /node_modules
    - package.json
    - bsconfig.json
```

The compiler seems to be located in the `node_modules` under the root and not in the `node_modules` of `folder1`.
By searching for `bscPartialPath` instead of the nearest `bsconfig.json`, we can correctly determine the root.
However the `bsconfig.json` indicates the "build root" for the compiler artefacts and diagnostics in `folder1`.

Fixes https://github.com/rescript-lang/rescript-vscode/issues/10

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -126,33 +126,33 @@ let openedFile = (fileUri: string, fileContent: string) => {
 
   stupidFileContentCache.set(filePath, fileContent);
 
-  let projectRootPath = utils.findProjectRootOfFile(filePath);
-  if (projectRootPath != null) {
-    if (!projectsFiles.has(projectRootPath)) {
-      projectsFiles.set(projectRootPath, {
+  let buildRootPath = utils.findBuildRootOfFile(filePath);
+  if (buildRootPath != null) {
+    if (!projectsFiles.has(buildRootPath)) {
+      projectsFiles.set(buildRootPath, {
         openFiles: new Set(),
         filesWithDiagnostics: new Set(),
         bsbWatcherByEditor: null,
       });
       compilerLogsWatcher.add(
-        path.join(projectRootPath, c.compilerLogPartialPath)
+        path.join(buildRootPath, c.compilerLogPartialPath)
       );
     }
-    let root = projectsFiles.get(projectRootPath)!;
+    let root = projectsFiles.get(buildRootPath)!;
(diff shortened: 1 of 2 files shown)


BlueHotDog:
I think that the logic you implemented might be a bit too naive, yarn will not hoist dependencies in a couple of cases:
- You marked a dep as nohoist
- There are different versions of dependencies.

I think a safer approach might be to:
find the project root by actually copying what yarn is doing, this looks like it:
https://github.com/square/find-yarn-workspace-root

but, we also need to check each folder on the way to see if bsc exists in its node_modules.
It's an expensive check, but we can do it only once when the extension starts.

BlueHotDog:
Actually, now looking at the code. my mistake, this seems to behave exactly as expected!

dan003400:
I have checked this branch out, ran a yarn compile, and nothing has changed, is there anything else needed to get this working?

dchambers:
I worked around this for the time being by adding this `package.json` config to the sub-project where I'm going to be using ReScript:

```
"workspaces": {
  "nohoist": ["bs-platform"]
},
```

and then running:

```
yarn --force
```

Alternatively you can do this globally by modifying the `workspaces` definition in your root `package.json` from something like this:

```
"workspaces": ["libs/*"],
```

to this:

```
"workspaces": {
  "packages": ["libs/*"],
  "nohoist": ["**/bs-platform"]
},
```

ryyppy:
I think this PR is obsolete due to #59

IwanKaramazow:
Superseded by https://github.com/rescript-lang/rescript-vscode/pull/59 and https://github.com/rescript-lang/rescript-vscode/commit/d75c6642f270a13981a4e833a3cb3c1ada7f698c

PR #28, opened 2020-11-28
renaming config to rescript

BlueHotDog:
Tiny fix. making the settings in vscode be "ReScript" instead of "Example configuration"


IwanKaramazow:
@BlueHotDog Good catch! Thanks

PR #29, opened 2020-11-28
Tweak project root heuristics

BlueHotDog:
This is based off the work of: https://github.com/rescript-lang/rescript-vscode/pull/16
I wanted to suggest a couple of conventions and organization practices as this project is expected to grow:
- I've started moving logic related to Project to "Project.ts"
- this is very personal, but i generally dislike a grab-bag files like utils.ts, they end-up being a catch all for various logic that fits no-where instead of forcing developers to think where things belong - So i extracted file related helpers to fs_helpers.

The benefit of having things like Project/File/ and various other Domain Contextes is that they're by definition easier to test.
What do you guys think? 
If you like it, i would love to proceed and organize Server.ts


IwanKaramazow:
@BlueHotDog would you be willing to set some tests up for this, before I take a look at this?

BlueHotDog:
Sure! i think it would be great to add contribution guidelines. how do i run tests? what tests should i write? do we have code-coverage?

BlueHotDog:
@IwanKaramazow hey, could you please give me a quick rundown on how tests are structured currently and what i should run to satisfy the project requirements(happy to add them myself to the readme)

IwanKaramazow:
@BlueHotDog Sorry for the late reply, turns out there are no tests in place yet. Will have to take care of that first

BlueHotDog:
NP, maybe we can manually test it and merge it meanwhile? we can't adopt it right now since we use a monorepo

BlueHotDog:
Hi guys, whats the ETA on this?
We're waiting for this PR to land to be able to use the extension on our monorepo. how can we push this?

IwanKaramazow:
@BlueHotDog taking a look at this now, can you rebase on master?

IwanKaramazow:
Also does this work locally on your setup? Server doesn't seem to start with this branch, might be something on my computer?

BlueHotDog:
@IwanKaramazow well, this was stale so a bunch of breaking changes merged to master.
Just rebased and checked locally and it seems to work. We're still blocked from using the extension until this lands. Can we please check/merge this so no more conflicts arise? this seems like a very reasonable thing to add.

IwanKaramazow:
@BlueHotDog I formalised the different cases in https://github.com/rescript-lang/rescript-vscode/pull/51
Once Hongbo gives his approval, we can go ahead and merge this.

BlueHotDog:
hey guys, any way i can help with this? really want to contribute to this project..(and also start actually using the new syntax)

chenglou:
Hey BlueHotDog! We'd like to avoid early compartmentalization because the editor's various use-cases are still not well defined, so the compartmentalization is more harmful than beneficial. For example, this would have made #59 and subsequent cleanups harder. Thankfully that one was merged first, which obsoletes most of this PR's cleanup actually (it's a good thing). Independent of this PR, one thing is that we don't want to have a notion of an editor "project". This is a concept that doesn't fit well into many's workflow (e.g. the moment you open a new tab on a file that isn't part of a project) and certainly goes against most other LSP-enabled editors' usages (though we also try to only use LSP for vscode and vim, due to LSP being not ideal in general).

Thanks for the effort though =)

chenglou:
Though we should write down this latter part somewhere more prominent.

BlueHotDog:
thanks for the elaborate response @chenglou . two questions:
- does this now support monorepo?
- I think that thinking about the editor/extension domain is a good idea, even if that thinking will be through some documents vs code. specifically for new-comers(like me) that will be able to learn the concepts without the need to understand the technical stuff(maybe worth a different issue to discuss).

chenglou:
#59 supports monorepo now (see comments at https://github.com/rescript-lang/rescript-vscode/commit/5caa099c7d3133fa8e444dd1162388a4c34f6b7b)
You'll end up with a slower formatting experience because instead of using the native bsc.exe to format, we degrade to using the JS bsc in `node_modules/.bin`. Thankfully this only affects yarn workspace or whatever. The extra formatting delay in the non-monorepo case was unacceptable to us.

PR #38, opened 2020-12-06
Add support for Reason syntax

osener:
Thanks a lot for your work on this extension! Devtools and the new syntax work great, even better than my experience with reason-language-server. I've added basic Reason support based on reason-language-server. There's still some modules in our project that we haven't migrated to Reason, and it is really handy to be able to use this new and improved VSCode tooling on our legacy modules as well.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -84,6 +84,7 @@ export function activate(context: ExtensionContext) {
 		// Register the server for plain text documents
 		documentSelector: [
 			{ scheme: 'file', language: 'rescript' },
+			{ scheme: 'file', language: 'reason' }
 		],
 		synchronize: {
 			// Notify the server about file changes to '.clientrc files contained in the workspace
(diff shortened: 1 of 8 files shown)


ryyppy:
Hey! Thanks for your effort submitting a PR.

Let me explain our rationale why we won't be able to merge this:

There is already a well known [Reason VSCode](https://marketplace.visualstudio.com/items?itemName=jaredly.reason-vscode) extension out there that kinda collides with our domain already. We want to clearly separate the boundaries of both language platforms (Reason / ReScript). It's an extra maintenance burden we can't afford (even if our Reason version is locked to 3.6, the Reason project will continue with their own v4 variation, and over time we will eventually get unrelated issues that are caused by Reason, not ReScript). Also we hit some weird issues when ppl installed both plugins during the transition phase, so we actually needed to make sure that both extensions don't overlap with their language detection.

It is recommended to run both VSCode extensions side by side to get support for both syntaxes. If you want to have the same LSP support, either submit a PR to the Reason VSCode extension to be able to talk to our rescript LSP, or use a local build of your forked ReScript VSCode extension to have mixed support for both syntaxes.

We are aware that this is inconvenient for some existing users, but we had to define a boundary on what we want to support officially, and we don't want to trade long-term maintainability with short-term benefits. 

I hope this makes sense. Cheers!

PR #54, opened 2020-12-28
Tweak bsc partial path.

IwanKaramazow:
The platform specific bsc might be in an unknown location. An example of this scheme is a variant of monorepos setup through yarn workspaces. The only guarantee we have is the `bsc` binary exposed at the root of the [project](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin)


IwanKaramazow:
Non-viable, the node overhead is too much.

Issue #55, opened 2020-12-29
Errors and Problems don't update from /lib/bs/compiler.log


marcoemrich:
this might be a duplicate of #34. I post it as a new issue since it has nothing to do with Angular or a specific project.

**Description**: VS Code does not update errors (or the lack of) from the complier-log. Neither Problem-View nor "red underlines".

Since this a very basic function, I suspect it has to do with my environment which is Node 14, Ubuntu 20.04 and VSCode 1.52. I used a new clean build of VSCode and only installed rescript-vscode 1.0.2.

**Steps**: See the the Video to reproduce 

https://user-images.githubusercontent.com/447570/103289983-c453d200-49e8-11eb-8510-d5cf51d7e0d4.mp4

**Observations:**

* The Compiler-Log is always correct
* Restarting VSCode updates the VSCode views, but it still does not update again

Thanks so much for all your hard work. Let me know, if I can do anything to help track down the issue. (I'm very dedicated, this is the only issue that prevents my team to switch to ReScript).

marcoemrich:
It gets weirder :( I tried to debug the extension. When I set a breakpoint at the beginning of "sendUpdatedDiagnostics" in the server.ts - and press "continue", it just works. As soon as I remove the breakpoint, it stops working :(

marcoemrich:
Update: I found the issue! It is a race condition with [chokidar](https://github.com/paulmillr/chokidar) and ```fs.readFileSync```. It can be fixed by provideing an option to chokidar.watch like described here https://stackoverflow.com/questions/60460751/getting-empty-string-from-fs-readfile-inside-chokidar-watchpath-file-onchang

I experimented and found out that a treshhold of 1 millisecond is enough:

```awaitWriteFinish: {stabilityThreshold: 1} ```

PR will follow!

IwanKaramazow:
Thanks for the PR =D Closing as this is now fixed: https://github.com/rescript-lang/rescript-vscode/pull/56

PR #56, opened 2020-12-30
fix watcher race condition

marcoemrich:
see issue #55
on some sytems the `.compiler.log` was not readable after
_chokidar_ called _sendUpdateDiagnostics_. Setting
_awaitWriteFinish_ as an option fixes the issue

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -108,7 +108,7 @@ let deleteProjectDiagnostics = (projectRootPath: string) => {
 let compilerLogsWatcher = chokidar
   .watch([], {
     awaitWriteFinish: {
-      stabilityThreshold: 500,
+      stabilityThreshold: 1,
     },
   })
   .on("all", (_e, changedPath) => {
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks! This heuristics is really unfortunate. And 500ms is a _lot_. I'd rather long pool than use a watcher in this case, as the watcher has other problems (e.g. not detecting the creation of the `lib/bs` directory at all). Are you sure 500ms is needed?

marcoemrich:
ups, no! I already updated to 1ms. Seems it didn't make it upstream. Sorry - I update the PR.

chenglou:
Cool, thanks! Merging. Let's hope there's no report of new odd behaviors. Watchers are scary things we're thinking of removing one day...

marcoemrich:
Thanks for Merging! Yes, watchers are really hard to use, every file system could behave differently. I'm just happy it works right now :)

PR #57, opened 2020-12-30
Add support for jsx html element highlights

TomiS:
I'm not sure if this is at all how you want to solve this. But since I spent some time toying with this approach, I might as well post it as a PR.

Some notes:
- Regex matches all html tags based on the existence of opening chevron and one of explicitly listed tag names. The opening tag then expects either closing chevron or at least one whitespace character (to avoid collisions with type arguments)
- Closing tag works similarly except that it doesn't allow any content other than the tag name. 
- The tag list is just pasted from somewhere and is almost certainly not complete. I can figure out where to find a complete list if the approach is considered good otherwise.

### DIFF
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -325,7 +325,23 @@
               "name": "entity.name.namespace"
             }
           }
-        }
+        },
+        {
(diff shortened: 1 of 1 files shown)


TomiS:
Hmm, I realized this fails with code that uses html tag names as type names.
```
type div: bool = false;
let foo: option<div> = Some(div)
```

chenglou:
Whoever's reading this: make sure you've seen my comment at https://github.com/rescript-lang/rescript-vscode/pull/8#issuecomment-764469070 first

I'm open to consider coloring lower-cased JSX tags. But not if they discolor `option<foo>`. There might be some circumventions for this. I'll think about it.

amiralies:
what about just coloring angle brackets and equal signs?
image
 
This is from vim-rescript
image

chenglou:
That could work, but requires detection the jsx block anyway

chenglou:
Btw @amiralies see https://github.com/rescript-lang/rescript-vscode/#recommendation and https://github.com/rescript-lang/rescript-vscode/pull/8#issuecomment-764469070. Your `=` should already be colored. Maybe for operators in particular, I can add another scope.

chenglou:
The highlighting (using the same ported grammar) in Sublime Text:

Screen Shot 2021-02-03 at 2 26 09 AM

Unfortunately vscode's default dark theme doesn't highlight any operator...

amiralies:
is that the same exact tm grammar mapped to submile-syntax?
I think submile-syntax is more powerful than tm grammar.

chenglou:
It's the same, though for the example above I've tweaked a few things for clarity, ignore that. But the main point is that your theme (and mine, the default vscode dark theme) don't highlight stuff like operators.

Yeah I'm familiar with the difference between the two grammars. Part of why I didn't lazy out with reusing TM grammar for Sublime is precisely because the new Sublime grammar is much nicer.

zth:
Note: https://github.com/rescript-lang/rescript-vscode/pull/367 will implement the equivalent of this. Thank you @TomiS !

PR #71, opened 2021-01-25
Windows fixes

### LINKED ISSUE #41: on windows, the error diagnostics does not work
```rescript
Js.log ("hel")

let a = 1 + 2

let u = [1,2]

let v = u + 3
```

The problem panel is empty.

lib/bs/.compiler.log is correct

```
#Start(595963434)


  We've found a bug for you!

  C:\Users\hongbozhang\git\hello\src\demo.res:10:9

  

   8 │ let u = [1,2]

   9 │ 

  10 │ let v = u + 3

  11 │ 

  

  This has type: array<int>

  Somewhere wanted: int

  

#Done(595964572)
```

johnny88:
Some fixes to the extension on windows including showing errors and running bsb.

- The error path wasn't taking into account that on windows paths often start with C:/ while looking for a ':'.
- On windows, you have to call child processes [differently](https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows).

Thanks to @mewhhaha for doing all the work!

Fixes #41


client/package-lock.json
```diff
@@ -1,264 +1,8 @@
 {
 	"name": "rescript-language-client",
 	"version": "0.0.1",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
chenglou, line 4: Humm I guess this should not be in this diff? Also the team's using a newer version of npm, thus the lockfile version difference

server/src/utils.ts
```diff
@@ -273,10 +297,11 @@ export let parseCompilerLogOutput = (
 
 	let result: filesDiagnostics = {};
 	parsedDiagnostics.forEach((parsedDiagnostic) => {
-		let [fileAndLocation, ...diagnosticMessage] = parsedDiagnostic.content;
-		let locationSeparator = fileAndLocation.indexOf(":");
-		let file = fileAndLocation.substring(2, locationSeparator);
-		let location = fileAndLocation.substring(locationSeparator + 1);
+		let [fileAndLocationLine, ...diagnosticMessage] = parsedDiagnostic.content;
+
+		let fileAndLocation = fileAndLocationLine.trim()
```
chenglou, line 302: Why the trim here?

chenglou (changes requested):
Thanks a lot! Left some comments =D

server/src/utils.ts
```diff
@@ -105,10 +105,14 @@ export let runBsbWatcherUsingValidBsbPath = (
 	bsbPath: p.DocumentUri,
 	projectRootPath: p.DocumentUri
 ) => {
-	let process = childProcess.execFile(bsbPath, ["-w"], {
-		cwd: projectRootPath,
-	});
-	return process;
+	switch (os.platform()) {
+		case "win32": return childProcess.exec(`${bsbPath}.cmd -w`, {
```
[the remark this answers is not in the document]
mewhhaha: Maybe all these switch statements I introduced should be if-else? Would that be better?

server/src/utils.ts
```diff
@@ -105,10 +105,14 @@ export let runBsbWatcherUsingValidBsbPath = (
 	bsbPath: p.DocumentUri,
 	projectRootPath: p.DocumentUri
 ) => {
-	let process = childProcess.execFile(bsbPath, ["-w"], {
-		cwd: projectRootPath,
-	});
-	return process;
+	switch (os.platform()) {
```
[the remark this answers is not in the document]
mewhhaha: Sorry, missed that! Will change to process.platform. 😄

server/src/utils.ts
mewhhaha: I saw that the lines came in with spacing in front of the path, like: '  C:\\rescript-project-template\\src\\Demo.res:1:14-20\r\r'. Would have to remove the whitespace so it's possible to slice off the drive letter and colon afterwards. I think this was done with .slice(2) before. Would you rather it be done using that method? Is it always 2 spaces?

client/package-lock.json
mewhhaha: You're right. Should restore the lock-file. :-)

server/src/utils.ts
johnny88: It was `substring(2)` but that's the issue we saw. It can be done in the original way.

I do feel like trim makes it clearer what's going on, and can be more robust if the compiler decides to change the number of spaces (like @mewhhaha seems to be implying from 'Is it always 2 spaces')

Let us know which you prefer and it's an easy fix.

server/src/utils.ts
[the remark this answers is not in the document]
mewhhaha: I changed all of them into if elses.

server/src/utils.ts
chenglou: Ah yes, the 2 spaces is guaranteed. I've put a note here just now: https://github.com/rescript-lang/rescript-vscode/blob/20841214830a3a81140f63a8973ab5d08cfcd79f/CONTRIBUTING.md#compilerlog

Whenever it's not, it's a bug that shouldn't be covered up here, but rather fixed int the compiler. We've caught a few compiler diagnostics display bugs this way actually.

In general also, we'd like to avoid trimming: https://github.com/rescript-lang/rescript-vscode/blob/20841214830a3a81140f63a8973ab5d08cfcd79f/CONTRIBUTING.md#formatting-newline

FYI trimming has caused and hidden all sorts of bugs before. Sometime the formatting adds a newline, or two, then the editor trims, adds a new line, or whatever. Lots of little shuffling.

chenglou:
Btw, what about `bsc`? https://github.com/rescript-lang/rescript-vscode/blob/20841214830a3a81140f63a8973ab5d08cfcd79f/server/src/utils.ts#L73

server/src/utils.ts
```diff
@@ -154,6 +159,28 @@ export let parseDiagnosticLocation = (location: string): Range => {
...
+		return fileAndLocation.indexOf(":")
+	}
+}
+
+let separateFileAndLocation = (fileAndLocation: string): [string, string] => {
+	let locationSeparator = findLocationSeparator(fileAndLocation)
+	let file = fileAndLocation.slice(0, locationSeparator)
+	let location = fileAndLocation.slice(locationSeparator + 1)
+	if (process.platform === "win32") {
+		return [`file:\\\\\\${file}`, location];
```
chenglou, line 176: So what's this for?

chenglou: Somewhat related: I had a note about `file:` at https://github.com/rescript-lang/rescript-vscode/blob/20841214830a3a81140f63a8973ab5d08cfcd79f/server/src/utils.ts#L20. Not sure if relevant

server/src/utils.ts
johnny88: I'll go ahead and fix this and change it back to the cutting out the 2 spaces using substring.

server/src/utils.ts
johnny88: Yeah I think so too, I'll try it and see if that gives us the same result

mewhhaha: Seems to be working on my Ubuntu and Windows 10. Do you see the current approach (Uri.file(..)) to be an issue?

johnny88: I should be what @chenglou was talking about, the utility vscode has for handling files.

chenglou:
Sorry, maybe I got ahead of myself but I thought the vscode-uri util would have been more useful; I guess not right now.
The extra dep isn't worth the one line usage; I'll remove it.

Merging now. Thanks so much! Could you test master afterward? Then I'll make a release.

mewhhaha:
I tried it on my Windows 10 machine and it worked! Thanks!

chenglou:
I've cleaned up the file and location parsing code to parse even more seriously. Now it doesn't accidentally parse the wrong `:`, so doesn't need windows-specific heuristics in that one place anymore.

Will release tomorrow.

mewhhaha:
It seems like a good idea to match on the consistent part of the incoming line, however this regex doesn't work for me as the incoming line looks like `C:\\rescript-project-template\\src\\Demo.res:1:14-20\r\r`. I'm guessing these carriage return characters are the issue.

chenglou:
You mean `\r\n` right?

mewhhaha:
I would've expected \r\n, but the debugger gives me \r\r. Is that just an issue with the debugger?

chenglou:
Ah probably because of https://github.com/rescript-lang/rescript-vscode/blob/1cd7b813e450b4346ba9ddc9900f77d2114c44c6/server/src/utils.ts#L247

mewhhaha:
Maybe splitting on /\r?\n/ would help?

chenglou:
I'm splitting on `os.EOL` now. Try that?

mewhhaha:
It doesn't work for. There's still lingering \r characters on a few of the lines. Maybe it'd help if I posted the content of what's being read from the compiler.log:

`'#Start(359482524)\r\n\r\r\n  We've found a bug for you!\r\r\n  C:\\rescript-project-template\\src\\Demo.res:1:14-20\r\r\n  \r\r\n  1 │ let x: int = "asdas"\r\r\n  \r\r\n  This has type: string\r\r\n  Somewhere wanted: int\r\r\n  \r\r\n  You can convert string to int with Belt.Int.fromString.\r\r\n  \r\r\n#Done(359482556)\r\n'`

chenglou:
That log file consists of bsb, ninja and ocaml's format module, so I'm thinking maybe there's some mix of hard-coded `\n` in there and `\r\n` from one or more of these that properly handles newline. Except... that wouldn't produce a double `\r`...?

mewhhaha:
Splitting on /\r?\r?\n/ would work, but maybe that's circumventing the issue. Seems like it's \r\n that's being translated into \r\r\n at least and it only occurs between #Start and #Done. Perhaps it looked like #Start(..)\n\r\n at some point and all newlines were prefixed with an \r, like \n\r(\r)\n.

chenglou:
Ugh, I guess I'll have to end up using `trim()` anyway...

chenglou:
Done!

mewhhaha:
I think it'd be the most maintainable way of doing it, just trim instead of slice. \r?\r?\n? just feels like someone would go in and accidentally "correct" it later on. But you could limit the trim to Windows if you want. :-)

mewhhaha:
I'll try it out!

mewhhaha:
It's working. Thank you very much! Great work!

chenglou:
Np. Don't like the circumvention that hides more bug, but it'll have to do for now =)

mewhhaha:
Yes, I understand. I think so too. The first impression for Windows users will be much better now!

PR #76, opened 2021-02-04
[Test] autocomplete for pipe-first.

cristianoc:
```res
let list = Belt.List.make(12, "abc")
```

then

```res
list->
```

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -280,7 +280,7 @@ process.on("message", (msg: m.Message) => {
           hoverProvider: binaryExists,
           definitionProvider: binaryExists,
           completionProvider: binaryExists
-            ? { triggerCharacters: ["."] }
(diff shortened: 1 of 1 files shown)


server/src/server.ts
```diff
@@ -280,7 +280,7 @@ process.on("message", (msg: m.Message) => {
           hoverProvider: binaryExists,
           definitionProvider: binaryExists,
           completionProvider: binaryExists
-            ? { triggerCharacters: ["."] }
+            ? { triggerCharacters: [".", ">"] }
```
chenglou, line 283: Lol I guess triggerCharacters doesn't accept a compound `"->"`?

cristianoc: It does not -- I too was hoping it would.

cristianoc:
Now part of https://github.com/rescript-lang/rescript-vscode/pull/94

Issue #81, opened 2021-02-06
Unify various `exec` calls


chenglou:
I'll cc @cristianoc, @mewhhaha and @john-pangalos (sorry for the ping) on this one.

Currently in the codebase there are 3 ways to call a binary:

- Async `exec`, with path and argument escaping: https://github.com/rescript-lang/rescript-vscode/blob/d1e5411e0f82cc0078e3d4a07900d521b085c9b1/server/src/RescriptEditorSupport.ts#L48
- Async `exec`, that doesn't escape path but does use the right windows bsb binary: https://github.com/rescript-lang/rescript-vscode/blob/d1e5411e0f82cc0078e3d4a07900d521b085c9b1/server/src/utils.ts#L107-L114
- Sync `execFileSync`, that might not call the right windows bsc binary: https://github.com/rescript-lang/rescript-vscode/blob/d1e5411e0f82cc0078e3d4a07900d521b085c9b1/server/src/utils.ts#L83

We should really revisit those and streamline them:
- First one:
  - [x] should be `execFile`, which avoids needing to escape the command path.
  - [x] should also be sync.
  - [x] should work on windows, if it isn't.
- Second one:
  - [ ] ~~should use `execFile` in the `win32` branch too, to avoid needing to escape path.~~ Can't use that for a batch script
- Third one
  - [x] should use the right windows bsc binary, if it isn't (where's the needed `.cmd` part? https://github.com/rescript-lang/rescript-vscode/pull/71#issuecomment-768200507).

mewhhaha:
For the second case we followed the [documentation](https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows) on NodeJS. Is it possible to get around that if the file is a script? 

What's the condition for getting the popup for starting this process? Is it by having a single .res file open in the project?

chenglou:
Oh right, yeah thanks for the reminder. I've finally understood the little complexities here. (Thanks also to @bsansouci for debugging on windows with me). d1dc096acf7ee1f013fe389f22d51b172adb16e8 fixes everything.

Thanks again!

PR #83, opened 2021-02-06
execSync -> execFileSync for RescriptEditorSupport binary invocation

chenglou:
Part of #81

### DIFF
--- a/server/src/RescriptEditorSupport.ts
+++ b/server/src/RescriptEditorSupport.ts
@@ -21,7 +21,7 @@ let findExecutable = (uri: string) => {
   } else {
     return {
       binaryPath: binaryPath,
-      filePathQuoted: '"' + filePath + '"',
+      filePath: filePath,
       cwd: projectRootPath,
     };
   }
@@ -38,7 +38,7 @@ export function runDumpCommand(msg: RequestMessage): dumpCommandResult | null {
   }
 
   let command =
-    executable.filePathQuoted +
+    executable.filePath +
     ":" +
     msg.params.position.line +
     ":" +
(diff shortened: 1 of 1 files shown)


chenglou:
@cristianoc this works, but then it looks like the hover's path isn't correctly relativized anymore:

image

I guess the binary doesn't relativize correctly when there's some quoting

cristianoc:
The argument should not be quoted anymore now that an array of arguments is passed.

server/src/RescriptEditorSupport.ts
```diff
@@ -38,16 +38,16 @@ export function runDumpCommand(msg: RequestMessage): dumpCommandResult | null {
   }
 
   let command =
-    executable.binaryPathQuoted +
-    " dump " +
     executable.filePathQuoted +
```
cristianoc, line 41: The argument should not be quoted anymore now that an array of arguments is passed.

chenglou: I'm modifying the binary right now. `execFileSync`'s arguments array's each item is considered as a space-separated command. Currently this still needs quoting since we concatenate the file path with the 2 position arguments. If we want to leverage proper escaping from `execFileSync`, we need to stop concatenating too. Aka the binary's position arguments shouldn't be passed as `filePath:line:col` but `filePath line col`.

chenglou: Nvm, you're right. This... works. We'll revisit the binary's `path:line:col` argument parsing later. Less parsing = less error

chenglou:
Thanks for the review.

PR #84, opened 2021-02-15
Fix keyword list specified in grammer

abhinandanarya06:
Some of keywords are from OCaml that are not from Rescript docs, Hence i filter out all keyword specified in docs from keyword list specified in grammer

Here is the fixes

List of keyword that were specified in grammer json but not in docs
`nonrec, val, then, fun, virtual, do, done, catch, begin, pub, or, library, class, new, finally, inherit, export, object, pri, private, import, raise, functor`

List of keyword that were specified in rescript docs but not in grammer json
`true, include, open, when, with, false, type`

### DIFF
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -15,7 +15,7 @@
     },
     "RE_KEYWORDS": {
       "name": "keyword.control",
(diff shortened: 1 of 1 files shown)


bobzhang:
LGTM. CC @chenglou

chenglou:
Thanks!

PR #90, opened 2021-03-09
Add command to create interface files

fhammerschmidt:
image

Similarly, other commands could be implemented (like convert .re to .res, or view ppxed source, etc.)

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -42,6 +42,8 @@ let projectsFiles: Map<
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: m.Message) => void = (_) => { };
 
+let createInterfaceRequest = new v.RequestType<null, string, void>("rescript-vscode.create_interface");
+
 let sendUpdatedDiagnostics = () => {
   projectsFiles.forEach(({ filesWithDiagnostics }, projectRootPath) => {
     let content = fs.readFileSync(
@@ -514,7 +516,7 @@ function onMessage(msg: m.Message) {
           }
         }
       }
-    } else if (msg.method === "custom:rescript-vscode.create_interface") {
+    } else if (msg.method === createInterfaceRequest.method) {
       let extension = path.extname(msg.params.uri);
       let filePath = fileURLToPath(msg.params.uri);
       let bscExeDir = utils.findBscExeDirOfFile(filePath);
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -554,7 +554,7 @@ function onMessage(msg: m.Message) {
 
         send(response);
       } else {
-        let cmiPartialPath = filePath.split(projDir)[1].replace(c.resExt, c.cmiExt);
+        let cmiPartialPath = utils.replaceFileExtension(filePath.split(projDir)[1], c.cmiExt);
         let cmiPath = path.join(projDir, c.compilerDirPartialPath, cmiPartialPath);
         let cmiAvailable = fs.existsSync(cmiPath);
         
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -514,6 +514,82 @@ function onMessage(msg: m.Message) {
           }
         }
       }
+    } else if (msg.method === "custom:rescript-vscode.create_interface") {
+      let extension = path.extname(msg.params.uri);
+      let filePath = fileURLToPath(msg.params.uri);
+      let bscExeDir = utils.findBscExeDirOfFile(filePath);
+      let projDir = utils.findProjectRootOfFile(filePath);
+
+      if (bscExeDir === null || projDir === null) {
+        let params: p.ShowMessageParams = {
+          type: p.MessageType.Error,
+          message: `Cannot find a nearby ${c.bscExePartialPath} to generate the interface file.`,
+        };
+
+        let response: m.NotificationMessage = {
+          jsonrpc: c.jsonrpcVersion,
+          method: "window/showMessage",
+          params: params,
+        };
+
+        process.send!(response);
+      } else if (extension !== c.resExt) {
+        let params: p.ShowMessageParams = {
+          type: p.MessageType.Error,
+          message: `Not a ${c.resExt} file. Cannot create an interface for it.`,
+        };
+
+        let response: m.NotificationMessage = {
+          jsonrpc: c.jsonrpcVersion,
+          method: "window/showMessage",
+          params: params,
+        };
+
+        process.send!(response);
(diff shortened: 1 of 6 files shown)


TheSpyder:
This will be very helpful to me, my team had to write a custom vscode task to run the generate interface command (we’re using ocaml-platform but the same will apply when we eventually switch to rescript-vscode).

a command to do the re->res conversion would also be handy!

tsnobip:
great feature @fhammerschmidt, I definitely want this!

@TheSpyder, why don't you use the built-in `ml/mli` switch in the top right corner with vscode ocaml platform? It works quite well for me.

TheSpyder:
> @TheSpyder, why don't you use the built-in `ml/mli` switch in the top right corner with vscode ocaml platform? It works quite well for me.

That only switches to an existing interface file, it can't be used to generate new interfaces?

fhammerschmidt:
@TheSpyder, it should work now, according to this [comment](https://github.com/ocamllabs/vscode-ocaml-platform/issues/381#issuecomment-735361897).

TheSpyder:
> @TheSpyder, it should work now, according to this [comment](https://github.com/ocamllabs/vscode-ocaml-platform/issues/381#issuecomment-735361897).

Hmmm it's not working for me but we're getting off topic. I'll figure out what's wrong somehow 😅

server/src/utils.ts
```diff
@@ -99,6 +99,45 @@ export let formatUsingValidBscExePath = (
...
 
+export let createInterfaceFileUsingValidBscExePath = (
+  filePath: string,
+  cmiPath: string,
+  bscExePath: p.DocumentUri
+): execResult => {
+  let mliTempFile = createFileInTempDir(c.mliExt);
+
+  try {
+    // First, create a temporary .mli file
```
chenglou, line 110: You can just create a `.res` file and this command will generate the `.resi` file

server/src/server.ts
```diff
@@ -464,6 +464,82 @@ process.on("message", (msg: m.Message) => {
           }
         }
       }
+    } else if (msg.method === "custom:rescript-vscode.create_interface") {
```
chenglou, line 467: Is there a way not to use a custom language-server command? Just a vscode plugin-specific API? Or is this the idiomatic way of making things work?

server/src/utils.ts
fhammerschmidt: Do you mean to hand the `.res` to `bsc` instead of the `.cmi`?
Because that only worked for me with files which open no other modules. The `.cmi` approach always worked for me, but it yields ml syntax.

chenglou: Oh, right...

server/src/server.ts
fhammerschmidt: I guess all the work could be done on the client side which would make the message passing obsolete.
OTOH if such a process would take longer, I guess it would be better to have it run in a separate process which would be the server.

server/src/utils.ts
chenglou: Done now: https://github.com/rescript-lang/rescript-compiler/issues/5023#issuecomment-803730108

fhammerschmidt: Yeah, that's a lot better: b59e7d030e5349ed3a7d4e8c5a651ac81d36a0f0

server/src/server.ts
fhammerschmidt: Did the idiomatic approach now, and created a request type `createInterfaceRequest` for it.

fhammerschmidt:
Ok, rebased the PR to current master and made it work with the new rescript package as well.

fhammerschmidt:
@chenglou Would love to merge this!

server/src/server.ts
```diff
@@ -514,6 +520,82 @@ function onMessage(msg: m.Message) {
...
+        let response: m.NotificationMessage = {
+          jsonrpc: c.jsonrpcVersion,
+          method: "window/showMessage",
+          params: params,
+        };
+
+        send(response);
+      } else {
+        let cmiPartialPath = filePath.split(projDir)[1].replace(c.resExt, c.cmiExt);
+        let cmiPath = path.join(projDir, c.compilerDirPartialPath, cmiPartialPath);
```
cristianoc, line 558: Should this use `replaceFileExtension`?

chenglou:
This doesn't work with JSX and there are several critical outcome printer problems to fix first, so I'll have to temporarily disable it. Sorry @fhammerschmidt. Thanks for the PR though. Let's see if we can re-enable this after the fixes

fhammerschmidt:
So you'd rather only ship it when it's perfect? Makes sense, I found it to be still rough around the edges even without JSX, e.g. when optional args are involved. 

I think it is still pretty nice to get a mostly complete `.resi` out of it and adapt the rest by hand, but I can wait.

chenglou:
Yeah we don't wanna ship things until they're usable. Too much of the tools are broken. It's bad for the community culture. And a note like "this doesn't work with JSX and maybe other stuff" will just get ignored.

This one's on us though. But to be fair it's caused by PPX, which the compiler has no knowledge of. Will have to think of a way to fix it.

chenglou:
Also, sorry for delaying merging this so much, given your effort to keep up with the latest changes...

chenglou:
@fhammerschmidt I'm thinking that maybe it'll be good if we can detect that the file has JSX. Not before interface generation (since that'd require parsing and making up new parser apis _just_ for a temp fix), but after, if we can use string grep to detect presence of JSX and other invalid things like `@module`. Then we can put an alert that this feature currently don't work with those ppxes

cristianoc:
For context, this:
```rescript
@react.component
let make = (~name) => React.string(name)
```

generates this:
```rescript
external makeProps: (~name: 'name, ~?key: string, unit) => {"name": 'name} =
  "" "#rescript-external"
let make: {"name": string} => React.element
```
which does not even compile.

Instead, it ought to ideally generate this:
```rescript
@react.component
let make: (~name: string) => React.element
```

cristianoc:
For additional context, some pretty significant simplifications to the ppx are being considered, so it's probably not worth investing much in workarounds for fixing at this stage. And just detecting JSX could be a decent short term thing.

cristianoc:
If you have access to the raw source text, even just looking for `@react.component` might be a good enough approx.

cristianoc:
The other issue, solved only in master for now, is that the pretty printer was not working in interface files for inner modules.
Since the fix is only in master, there's an implicit dependency on the next compiler release.

Issue #72, opened 2021-01-29
Functions that have both type annotation and doc are not shown in the autocomplete window


gaku-sei:
Say we have an `ArrayExtra` module, and define a function inside like the following:

```rescript
@ocaml.doc("Hello doc")
let usefulFunction: unit => int = () => 1
```

Typing `ArrayExtra.u` will show the following:

<img width="498" alt="Capture d’écran 2021-01-29 à 14 02 29" src="https://user-images.githubusercontent.com/3982371/106233791-a0c7b580-623a-11eb-9ab5-f2f9a8d1d3be.png">

If I remove the type annotation, or the doc, it works:

<img width="1081" alt="Capture d’écran 2021-01-29 à 14 04 12" src="https://user-images.githubusercontent.com/3982371/106233877-df5d7000-623a-11eb-8cac-094a53f87a44.png">

While I'm aware the syntax for doc will change soon and not use `@ocaml.doc` anymore, and while I'm also aware that type annotations are optional in many cases, I thought it could be interesting to understand why the function doesn't show up 😕 

_Edit_: My bad, it seems it works for some other functions. So when a function with type annotation + doc appears in the suggestions list, it'll always appear, but some functions don't appear for any reason, and the only to force them to appear is to drop the types for instance.

PR #94, opened 2021-03-15
Sync up with rescript-editor-support c57ed6d6c5a8f0ac93e0ae1878fd8d0a…

cristianoc:
…fb1a63d3

Sync up with https://github.com/rescript-lang/rescript-editor-support/commit/c57ed6d6c5a8f0ac93e0ae1878fd8d0afb1a63d3

#### New features
- Add support for autocomplete for pipe-first `foo->`: the type of `foo` is used to determine the module to take completions from.
- Add support for autocomplete for decorators such as `@module` and `@val`.
- Add support for autocomplete of labelled arguments `foo(~label... )`.

#### Fixes

- Fix issue for uncurried functions where the internal definition of `Js.Fn.arity` is shown on hover. (See https://github.com/rescript-lang/rescript-editor-support/issues/62).
- Fix type hint when hovering over labeled arguments of components (See https://github.com/rescript-lang/rescript-editor-support/issues/63).
- Fix issue where values declared with type annotation would not show up in autocomplete, and would show no doc comment on hover. (See https://github.com/rescript-lang/rescript-vscode/issues/72).
- Fix hover on definitions inside a react component module, or whenever multiple definitins for the same value exist in the module (See https://github.com/rescript-lang/rescript-editor-support/issues/67).
- Fix autocomplete issue where multiple open's were considered in the wrong priority order (See https://github.com/rescript-lang/rescript-editor-support/issues/72).

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -280,7 +280,7 @@ process.on("message", (msg: m.Message) => {
           hoverProvider: binaryExists,
           definitionProvider: binaryExists,
           completionProvider: binaryExists
-            ? { triggerCharacters: ["."] }
+            ? { triggerCharacters: [".", ">", "@", "~"] }
             : undefined,
         },
       };


chenglou:
Nice, nice

PR #97, opened 2021-03-26
Add markdown code block syntax highlight support

wangbinyq:


chenglou:
Hey, thanks! Though is this the idiomatic way of doing it? It seems typescript's grammar repo doesn't have such thing?

wangbinyq:
I rewrite the file from the [vscode-reasonml](https://github.com/reasonml-editor/vscode-reasonml) repo, just some minor changes.

fhammerschmidt:
@zth can we get semantic highlighting in markdown blocks as well?

zth:
@fhammerschmidt I'm not sure. Does TS do it?

aspeddro:
I was looking for how vscode apply syntax highlighting in markdown. It doesn't use semantic highlighting just TexMate grammar.

[Here](https://github.com/microsoft/vscode/blob/main/extensions/markdown-basics/package.json#L33-L86) are supported languages.

fhammerschmidt:
Having at least a bit of highlighting would already be nice.
Currently it is just all the same color.

image

zth:
I'm thinking we should merge this as is. I agree that some highlighting is better than none. Ok with you @cristianoc ?

Side note: There's more interesting things we can do eventually to better support md, like formatting code inside of ReScript md blocks.

PR #99, opened 2021-03-29
Sync rescript-editor-support

cristianoc:
Sync rescript-editor-support to commit https://github.com/rescript-lang/rescript-editor-support/commit/03ee0d97b250474028d4fb08eac81ddb21ccb082

[Update: don't show deprecation on hover].

Since release 1.0.5:

#### New features
- Add support for autocomplete for pipe-first `foo->`: the type of `foo` is used to determine the module to take completions from.
- Add support for autocomplete for decorators such as `@module` and `@val`.
- Add support for autocomplete of labelled arguments `foo(~label... )`.
- Add support for @deprecated attributes in autocomplete.
- Support for upcoming `rescript` npm package for the compiler. Looks for `rescript` in addition to `bs-platform` in `node_modules`.

#### Fixes

- Fix issue for uncurried functions where the internal definition of `Js.Fn.arity` is shown on hover. (See https://github.com/rescript-lang/rescript-editor-support/issues/62).
- Fix type hint when hovering over labeled arguments of components (See https://github.com/rescript-lang/rescript-editor-support/issues/63).
- Fix issue where values declared with type annotation would not show up in autocomplete, and would show no doc comment on hover. (See https://github.com/rescript-lang/rescript-vscode/issues/72).
- Fix hover on definitions inside a react component module, or whenever multiple definitins for the same value exist in the module (See https://github.com/rescript-lang/rescript-editor-support/issues/67).
- Fix autocomplete issue where multiple open's were considered in the wrong priority order (See https://github.com/rescript-lang/rescript-editor-support/issues/72).
- Autocomplete: add support for `open!` in addition to `open`.


cristianoc:
Updated: don't show deprecation on hover.
CC @chenglou

chenglou:
For posteriority, this is what it looks like with duplicated warnings:

image
Screen Shot 2021-03-29 at 7 58 02 AM

chenglou:
Thanks

Issue #136, opened 2021-04-25
Should have find-references command?


cristianoc:
If so, to be taken back from this removal:
https://github.com/rescript-lang/rescript-editor-support/commit/3d7d8e4dbf27c3fc6b8165e2738f2ce4af4c7684

TS uses it also when using jump-to-definition on a definition.

cristianoc:
Added.

PR #137, opened 2021-04-25
Add a command to find all the references to an item.

cristianoc:
…s of an element.


cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/136

PR #138, opened 2021-04-25
remove odoc parser

cristianoc:


analysis/src/State.ml
```diff
@@ -4,12 +4,8 @@ open TopTypes
 let isMl path =
   Filename.check_suffix path ".ml" || Filename.check_suffix path ".mli"
 
-let odocToMd text = MarkdownOfOCamldoc.convert text
-
-let compose fn1 fn2 arg = fn1 arg |> fn2
-
 let converter src =
```
chenglou, line 7: Can this be further simplified?

cristianoc: probably

PR #139, opened 2021-04-25
Hook up command to go to references.

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/136

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
 	"requires": true,
 	"packages": {
 		"": {
+			"name": "rescript-vscode",
 			"version": "1.0.8",
 			"hasInstallScript": true,
 			"license": "MIT",
@@ -72,6 +73,7 @@
 			"dependencies": {
 				"anymatch": "~3.1.1",
 				"braces": "~3.0.2",
+				"fsevents": "~2.1.2",
 				"glob-parent": "~5.1.0",
 				"is-binary-path": "~2.1.0",
 				"is-glob": "~4.0.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
 	"requires": true,
 	"packages": {
 		"": {
-			"name": "rescript-vscode",
 			"version": "1.0.8",
 			"hasInstallScript": true,
 			"license": "MIT",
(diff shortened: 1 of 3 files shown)


server/src/server.ts
```diff
@@ -375,19 +376,37 @@ function onMessage(msg: m.Message) {
         // error: code and message set in case an exception happens during the definition request.
       };
       send(definitionResponse);
+    } else if (msg.method === p.ReferencesRequest.method) {
+      // https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references
+      let result: Location | null = utils.runAnalysisAfterSanityCheck(
```
chenglou, line 381: The annotation here is wrong btw. It should be `Location[]`

package-lock.json
```diff
@@ -73,7 +72,6 @@
 			"dependencies": {
 				"anymatch": "~3.1.1",
 				"braces": "~3.0.2",
-				"fsevents": "~2.1.2",
```
chenglou, line 76: That's weird

server/src/server.ts
```diff
@@ -375,19 +376,37 @@ function onMessage(msg: m.Message) {
...
       let result:
         | CompletionItem[]
         | null = utils.runAnalysisAfterSanityCheck(msg, (filePath) => [
-          "complete",
-          filePath,
-          msg.params.position.line,
-          msg.params.position.character,
-          tmpname,
-        ]);
+        "complete",
```
chenglou, line 404: These should probably be put back

server/src/server.ts
chenglou: I should put a comment somewhere, but basically `utils.runAnalysisAfterSanityCheck` returns `any` since it decodes some JSON. We don't carry the `result`s around but I wanted to annotate them just in case and for doc.

cristianoc: I got to used to types actually meaning something, as opposed to be room decoration.

server/src/server.ts
cristianoc: This is just auto formatting. Should I revert the formatting?

server/src/server.ts
cristianoc: done

package-lock.json
cristianoc: no idea -- anything to do?

server/src/server.ts
chenglou: Ah no then. I use Prettier because @IwanKaramazow's editor was configured to use it. Nvm

package-lock.json
chenglou: Yeah chokidar definitely depends on fsevents, so let's just revert this lockfile and then I'll investigate later

PR #140, opened 2021-04-25
Hover different on ocaml files.

cristianoc:
The hover command gives identical results for hover on doc comments in .res and .ml files.
Hovever vscode shows markdown only for those from .res files.


cristianoc:
Actually I was testing it wrong -- with the last published extension.
On master it works as expected.

PR #146, opened 2021-04-26
Try a better prod binary name

chenglou:
This is so that in activity monitor and other places, they look idiomatic and debuggable


chenglou:
@ryyppy I've made the name change as requested. This is better for the user.
@cristianoc heads up: binary name change. Hopefully not too many of these. This also fixes the pre-release binary naming mistake I mentioned earlier.

PR #148, opened 2021-04-26
Fix CI that has been broken on Windows.

chenglou:


cristianoc:
Edited this by mistake by re-using the same branch name `ci`.

Issue #152, opened 2021-04-26
Autocompletion gives me @react.react.component


cknitt:
If I type `@react.` and then press enter on the autocomplete suggestion “react.component”, I end up with `@react.react.component`.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/152

PR #154, opened 2021-04-26
Add test cases for completion of react.component

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/152


cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/152

PR #159, opened 2021-04-27
Upgrade most of deps; fix TS errors; put it on CI

chenglou:


server/src/server.ts
```diff
@@ -343,15 +343,14 @@ function onMessage(msg: m.Message) {
         send(response);
       }
     } else if (msg.method === p.HoverRequest.method) {
-      let result: Hover | null = utils.runAnalysisAfterSanityCheck(
-        msg,
-        (filePath) => [
```
chenglou, line 348: @cristianoc here's another inversion of control. Though it's unclear that passing just the argument data is better since it can be manipulated and interpreted unpredictably by the function too. Anyway this fixes a type error. A bit more duplication though

PR #189, opened 2021-04-28
Try a different rescript version

chenglou:


chenglou:
9.1.1 doesn't hang on windows

PR #190, opened 2021-04-28
Try rescript 9.1.0

chenglou:


chenglou:
9.1.0 doesn't hang. 9.1.0-dev.1 has some known install issue later patched.

PR #191, opened 2021-04-28
Try bs-platform latest

chenglou:


chenglou:
No hanging anymore. Fixed by #199

Issue #178, opened 2021-04-27
Hover over module that implements a module type does not work


cknitt:
(Beta2)

Given

```rescript
module JsLogger = {
  let log = (msg: string) => Js.log(msg)
}
```

hover over the module `JsLogger` works fine:

<img width="313" alt="Bildschirmfoto 2021-04-27 um 18 53 16" src="https://user-images.githubusercontent.com/591384/116281262-d58bd000-a789-11eb-9034-36dfe0d254cc.png">

However, for

```rescript
module type Logger = {
  let log: string => unit
}

module JsLogger: Logger = {
  let log = (msg: string) => Js.log(msg)
}
```

hover neither works for the module `JsLogger` nor for the module type `Logger`.

PR #193, opened 2021-04-28
Towards better treatment of modules and module types.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/178


cristianoc:
Tests work as expected now.
There's still the pre-existing issue of resolving transitive module references.

PR #197, opened 2021-04-29
Test using normal filename concat without trying to normalize `./foo`.

cristianoc:


cristianoc:
The relevant code originates here: https://github.com/jaredly/reason-language-server/commit/e7671ff134713612757e6ef01a3e7de7922a4978

Unclear whether any of that is needed here.

PR #199, opened 2021-04-29
Fix another infinite loop on paths on Windows.

cristianoc:


analysis/src/ModuleResolution.ml
```diff
@@ -1,5 +1,5 @@
 let rec resolveNodeModulePath ~startPath name =
   let path = Filename.concat (Filename.concat startPath "node_modules") name in
   if Files.exists path then Some path
-  else if startPath = "/" then None
+  else if startPath = "/" || Filename.dirname startPath = startPath then None
```
chenglou, line 4: The first condition can be dropped right?

cristianoc: Yes it can be dropped.

chenglou:
This is probably it.

Issue #202, opened 2021-04-29
Tests: cut the path at the current root rather than file name


cristianoc:
To catch more path issues on CI.
See: https://github.com/rescript-lang/rescript-vscode/commit/255e57819c4a4db329772fb97ffd68a70b4a4955

PR #206, opened 2021-04-30
Keep file paths from current dir in tests.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/202


cristianoc:
Not helping much.

Issue #213, opened 2021-05-01
JSX autocomplete breaks after certain prop types


zth:
Autocomplete for JSX stops working for me after I've added a prop with a string literal. So, this:

```
<MyComponent someProp=someVar anotherProp="Text here"
```
Completion works fine after someProp, but breaks after anotherProp. 

I think it happened with other things than string literals too, but I'm not at my computer right now. I'll amend this issue with info about later if it still happens.

cristianoc:
True, only works with `{"Text here"}` at the moment. Can be relaxed.

cristianoc:
Works for me. Do you have the complete example?

zth:
@cristianoc let me try and put up a proper reproduction as soon as I'm by the computer again.

zth:
@cristianoc turns out it wasn't strings at all, I was just fooled by the complexity of the component I had when I found it. Instead, it seems to be variants with payloads and _without_ wrapping {}. So, this:
```
@react.component
open Component

let make = () => <> <Component style=First("Text")  /> </>
```

Wrapping with `{}` makes autocomplete work again (`<Component style={First("Text")} />`), and variants without payloads does not break regardless of if they're wrapped or not. Reproduction here:
https://github.com/zth/rescript-editor-tooling-playground/blob/master/src/OtherComp.res

cristianoc:
Thanks yes that's expected. Can be extended to support that too.

zth:
Out of curiosity, why is it expected?

cristianoc:
Because that case is not covered yet.

zth:
Ah, right.

cristianoc:
Done in https://github.com/rescript-lang/rescript-vscode/pull/214

zth:
Thanks, impressive turn-around @cristianoc!

PR #214, opened 2021-05-02
JSX Autocomplete: handle more cases of atomic expression  to the rhs of `=` without braces (variant, polymorphic variant, function call, etc)

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/213


cristianoc:
This handles all the cases as far as I know.

cristianoc:
One case was missing: nested JSX.

PR #215, opened 2021-05-03
Fix crash when projectRootPath is undefined

CarlOlson:
This prevents a crash that I was getting when using emacs.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -114,7 +114,7 @@ let compilerLogsWatcher = chokidar
       stabilityThreshold: 1,
     },
   })
-  .on("all", (_e, changedPath) => {
+  .on("all", (_e, _changedPath) => {
     sendUpdatedDiagnostics();
   });
 let stopWatchingCompilerLog = () => {
@@ -534,6 +534,8 @@ function onMessage(msg: m.Message) {
       msg.result != null &&
       // @ts-ignore
       msg.result.title != null &&
+      // @ts-ignore
+      msg.result.projectRootPath != null &&
       // @ts-ignore
       msg.result.title === c.startBuildAction
     ) {


cristianoc:
@CarlOlson any ides when `projectRootPath` is not set, and why?

chenglou:
Yeah, would need a repro for this

CarlOlson:
> @CarlOlson any ides when `projectRootPath` is not set, and why?

Not sure. When I load a project it asks if I want to start a build, it only gives me one choice, and crashes if I select "start build". 

This is emacs-specific, but the popup happens inside of helm. I suppose there could be an interaction between helm and lsp-mode.

> Yeah, would need a repro for this

What kind of repo? We should assume there could be malformed input and handle this null.

---

An alternative would be to remove this code. It is just annoying.

chenglou:
The reason is that we don't support this repo's usage for other editors; all invariants internally are implementation details. See [this forum thread](https://forum.rescript-lang.org/t/rescript-emacs-support-with-rescript-vscode/1056).

I'll cc @jjlee. We need to know whether this is an emacs-specific bug.

CarlOlson:
I urge you to embrace LSP and not duplicate effort. Everyone can work together and share ideas this way.

But I understand your opinion, closing.

ryyppy:
> I urge you to embrace LSP and not duplicate effort

If we had the choice, we'd completely eliminate LSP, because it really is a technical burden that will ruin all our work reducing complexity.

JetBrains (IntelliJ) is probably the only sane company who is strictly against the LSP approach, because it knows that vertical integration is the key for a good product.

We are not opting for mediocrity, but a polished product experience. Being held hostage by a badly spec'ed protocol that mostly benefits the VSCode landscape is not good. That's why we are treating the LSP server as an implementation detail until we figured out proper integrations for each plugin.

jjlee:
This does not affect VS Code.

Also, for what it's worth:
1. This is handled by the lsp-rescript package that I submitted to MELPA the other day: https://github.com/melpa/melpa/pull/7558
2. An emacs lsp-mode maintainer already told me in chat that they consider differences from VS code behaviour to be bugs (hope I'm not misrepresenting, that's how I remember it) -- and I expect later to fix this upstream in lsp-mode
3. @CarlOlson per this thread: https://forum.rescript-lang.org/t/rescript-emacs-support-with-rescript-vscode/1056/24 please don't report bugs from using rescript-vscode in emacs against the rescript-vscode project, you can use this project instead to report issues: https://github.com/jjlee/rescript-mode .  I assume you started using the LSP server just by hooking it yourself without using the packages I submitted to MELPA (hopefully in future that thread or this bug will show up in searches and this won't keep happening)?  Where would have been a good place to point this out where you would have seen it?

CarlOlson:
@ryyppy I agree alot with what you are saying, but I still value LSP being a standard.

> If we had the choice, we'd completely eliminate LSP, because it really is a technical burden that will ruin all our work reducing complexity.

> We are not opting for mediocrity, but a polished product experience

This would be something that is good to know. However, where would I learn that? \*rhetorical\* 
I'll move further discussion to the forums.  

---

@jjlee Thanks for the reply, I appreciate it.

> I assume you started using the LSP server just by hooking it yourself without using the packages I submitted to MELPA

I've adapted my ReasonML code to work better with Rescript, based on `reason-mode`. I'm sure your package is great, but I use ReScript daily and value my current configuration that I know works.

> Where would have been a good place to point this out where you would have seen it?

I think the official docs list your repo. I am hesitant to depend on code owned by a single user. Moving any "unsupported" plugins to a community organization would help.

jjlee:
> I've adapted my ReasonML code to work better with Rescript, based on `reason-mode`. I'm sure your package is great, but I use ReScript daily and value my current configuration that I know works.

That's cool.  The `lsp-rescript` project does not depend on my `rescript-mode`, it's just lsp-mode config, which does solve this problem.

It would be nice if you would continue any further conversation on rescript-vscode github discussions rather than here or the ReScript forum.  I understand that you're not using that project: I suggested that on the ReScript forum just to have a place to discuss emacs/ReScript issues -- since the ReScript project itself doesn't want to be distracted from their core focus, which I think we should respect.

PR #218, opened 2021-05-04
CI: build with ubuntu-18.04

cristianoc:


cristianoc:
Recap:
`avsm/setup-ocaml@v1.1.10` was upgraded to `avsm/setup-ocaml@v2` as Windows build started failing. However it seems reverting to `v1.1.10` currently works on all platforms.
One difference is `v2` expects `ocaml-compiler: 4.06.1` and `v1.1.10` expects `ocaml-version: 4.06.1`. Added both so it's easier to switch.

It seems `ubuntu-18.04` does not work on `v2` but does work on `v1.1.10`. At least, today it does!

PR #227, opened 2021-05-07
Remove unneeded ci config

chenglou:


cristianoc:
I kept it there as a reminder, should we need to upgrade again for some reason, but indeed it's redundant.

PR #229, opened 2021-05-08
Fix using path as uri for diagnostics

amiralies:
This is based on lsp spec.
Some clients won't work with file paths (including neovim's built in lsp client).


amiralies:
cc/ @cristianoc

amiralies:
cc/ @chenglou

PR #230, opened 2021-05-08
Add rename provider

amiralies:
This PR adds lsp rename provider based on references ruquest on analysis cli.
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_rename

also fixed a small type error in references provider


cristianoc:
This looks pretty interesting. Looking.

server/src/server.ts
```diff
@@ -383,11 +385,49 @@ function onMessage(msg: m.Message) {
         // error: code and message set in case an exception happens during the definition request.
       };
       send(definitionResponse);
+    } else if (msg.method === p.RenameRequest.method) {
+      // https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_rename
+      let params = msg.params as p.RenameParams;
```
cristianoc, line 390: There's a bit of code duplication with the rename command.
Maybe it makes sense to move out functionality that could lead to errors in future. E.g. definitely the way the `references` command is called, which parameters, etc. So that a future edit does not change only one place out of two, by mistake.

server/src/server.ts
```diff
@@ -383,11 +385,49 @@ function onMessage(msg: m.Message) {
...
+        result,
+      };
+
+      send(renameResponse);
     } else if (msg.method === p.ReferencesRequest.method) {
       // https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references
       let params = msg.params as p.ReferenceParams;
       let filePath = fileURLToPath(params.textDocument.uri);
-      let result: Location | null = utils.runAnalysisAfterSanityCheck(
+      let result: Location[] | null = utils.runAnalysisAfterSanityCheck(
```
cristianoc, line 430: Thanks!

cristianoc: Actually we had a few of these.
See this alternative way which I'm going to merge now:
https://github.com/rescript-lang/rescript-vscode/pull/232

cristianoc: Merged --- perhaps rebase on master instead.

amiralies:
> @amiralies I tried on `analysis/tests/src/References.res`: this renames `References.res` but not `Cross.res`.
> Unless `Cross.res` is already open in the editor, in which case it works.
> I think this has to do with vscode's internals. But we would like to have a robust rename, just as we have robust error messages. Any ideas how to do that?

I couldn't reproduce this, tested on vscode, nvim with coc and nvim with built in lsp

https://user-images.githubusercontent.com/13261088/117534123-5dcd6900-b005-11eb-820e-cb766f44a3dc.mp4

Note that i close files before renaming.

server/src/server.ts
amiralies: I'm not sure if the types on #232 are the correct types.
I've also took look at protocol types in ts and couldn't find correct types.
I think we should rely on manual types based on protocl docs. for example in the docs the response type for references request is `Location[] | null` but there's not type in ts spec for that.

cristianoc:
Would you try opening the editor from scratch. Open a single file, then do rename?

amiralies:
> Would you try opening the editor from scratch. Open a single file, then do rename?

Sure, It works on my desktop.

https://user-images.githubusercontent.com/13261088/117536502-8eff6680-b010-11eb-80ce-0e422db020a2.mp4

cristianoc:
Actually, it works. Sorry for the mix-up.
Thanks for doing this!

PR #231, opened 2021-05-08
Add local opam switch dir to gitignore

amiralies:
so we can create a local switch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ analysis/tests/.merlin
 *.cmx
 *.o
 *.install
+# opam local switch
+analysis/_opam


cristianoc:
There's no opam in the repo.

amiralies:
Oh I know, if you create a local opam switch for the project (my ocaml version provided by os package manager is 4.10, and I don't want to use my global 4.06 opam switch since it includes other deps, I want an isolated zero dep ocaml setup for this project so I've to create a local opam switch) this folder (`_opam`) appears.

amiralies:
Oh I know, if you create a local opam switch for the project (my ocaml version provided by os package manager is 4.10, and I don't want to use my global 4.06 opam switch since it includes other deps, I want an isolated zero dep ocaml setup for this project so I've to create a local opam switch) this folder (`_opam`) appears.

PR #247, opened 2021-05-21
Rename labels: fix issue with `~xx` eating up `x`.

cristianoc:


cristianoc:
@IwanKaramazow see this for reference.
I'll merge, and will retract later in case this should be done in the parser instead.

PR #254, opened 2021-05-26
Fix references to file modules

amiralies:


amiralies:
This pr adds support for global references to a module.
if you have these structure.
```rescript
// A.res
let a = 2
```
```rescript
// B.res
let x = A.a
```
```rescript
// C.res
let x = A.a
```

requesting refrence on `A` in either `B` or `C` results in references to all these three files.
note that reference to `A` itself is a special `(0,0), (0,0) `loc.

cristianoc:
See how that interacts with rename:

Screenshot 2021-05-26 at 04 10 59

cristianoc:
There's also this confusing thing that the first line
```rescript
let x = 12
```
is considered as a reference to `References.res`.
That would leave me puzzled as a user.

So basically 2 things to be solved in user experience.

amiralies:
> See how that interacts with rename:

Will fix that



> is considered as a reference to `References.res`.

Do you think that's bad experience? for me it's like referencing to let binding when i get references on the usage site
for example
```rescript
let x = 2

let b = x
```
if i get refrence on second x, it includes the first let binding which declares x.

cristianoc:
This module reference is different. We want a reference to a module, we get the first line of a file. The user will look at the first line, and might think: how did I get here? That first line has nothing to do with the original query, except it happens to be in the file that the query is about.
It's not a terrible experience, but not a great experience either. That said, I'm not sure it's possible to do better.

amiralies:
Yeah your argument is valid.
and it's happening becaus module decleration for file-modules is implicit, we don't have a exact location to refer to that so we refer to top of file.

amiralies:
File rename demo

https://user-images.githubusercontent.com/13261088/119719831-80bda100-be7e-11eb-9be5-fc59abca07c5.mp4

server/src/server.ts
```diff
@@ -294,16 +294,34 @@ function rename(msg: p.RequestMessage) {
...
+    let textEdits: { [uri: string]: TextEdit[] } = {};
+    let documentChanges: (p.RenameFile | p.TextDocumentEdit)[] = [];
+
     locations.forEach(({ uri, range }) => {
-      let textEdit: TextEdit = { range, newText: params.newName };
-      if (uri in changes) {
-        changes[uri].push(textEdit);
+      if (utils.isRangeTopOfFile(range)) {
+        let filePath = fileURLToPath(uri);
+        let newFilePath = `${path.dirname(filePath)}/${params.newName}${path.extname(filePath)}`;
```
cristianoc, line 303: what about Windows paths?

analysis/src/References.ml
```diff
@@ -468,15 +468,34 @@ let forLocalStamp ~full:{file; extra; package} stamp tip =
...
+      match Hashtbl.find_opt package.pathsForModule moduleName with
       | None -> []
-      | Some locs ->
-        locs
-        |> List.map (fun loc ->
-               (Uri2.fromPath loc.Location.loc_start.pos_fname, [loc]))
+      | Some paths -> (
+        let moduleSrcToRef src = (Uri2.fromPath src, [Utils.topLoc src]) in
+        match paths with
+        | Impl (_, None) -> []
```
cristianoc, line 492: How about move this extraction of source files to `SharedTypes` alongside `getSrc`.

cristianoc (commented):
This is great, thanks a lot.
I've added a few comments.

In addition, how about adapting the test files so they cover all cases, including implementation and interface files, as well as references from another file than the current one.

amiralies:
thank you i will address them.
I've a question, should we move rename stuff into analysis cli in order to test them separately or testing references is enough?

cristianoc:
I think eventually it will have to move, as e.g. the `isRangeTopOfFile` check really belongs there, where one can make it less indirect.

amiralies:
I will move if it's okay. 

another thing have to be fixed. looks like there's a problem when the project has `namespace: true` in bsconfig.

cristianoc:
Great go ahead with moving.

cristianoc:
@amiralies meanwhile I've done some refactoring relevant to a couple of lines of this diff:
https://github.com/rescript-lang/rescript-vscode/pull/256

Specifically, type `paths` in `SharedTypes.ml`.

I can merge and this gets rebased, or it can be done later.

amiralies:
Maybe you can merge your branch then I will update my branch/PR

cristianoc:
Merged

cristianoc:
@amiralies a couple of next steps that fix build and clean up after rebase:
https://github.com/rescript-lang/rescript-vscode/pull/261

Feel free to incorporate and continue. Or I'm also happy to continue if you prefer.

amiralies:
I will merge your branch on mine, and continue the work here if it's okay.

cristianoc:
Perfect

cristianoc:
Windows tests are failing, I think because the order in which files are processed is nondeterministic. And can become platform-dependent.

analysis/src/Commands.ml
```diff
@@ -176,6 +176,78 @@ let documentSymbol ~path =
...
+  let uri = Uri2.fromPath path in
+  let result =
+    match ProcessCmt.getFullFromCmt ~uri with
+    | None -> Protocol.null
+    | Some full -> (
+      let pos = Utils.protocolLineColToCmtLoc ~line ~col in
+      match References.locItemForPos ~full pos with
+      | None -> Protocol.null
+      | Some locItem ->
+        let allReferences = References.allReferencesForLocItem ~full locItem in
```
cristianoc, line 189: This list I think can be non-deterministic.
Some fixed order should be chosen (e.g. by sorting the results on Uri then loc).

amiralies: I will check what's going wrong on a windows setup.

amiralies:
Works to be done aside from issue on windows:
- Rename tests
- Use rename command in server

analysis/src/Commands.ml
amiralies: Seems like there was an issue checking for interface file on windows

cristianoc: Great thanks. I've also looked at removing some nondeterminism:
https://github.com/rescript-lang/rescript-vscode/pull/263

Which I might add later on. To make debugging across platforms easier.

amiralies:
I think it's ready for review !

analysis/src/Commands.ml
```diff
@@ -176,6 +176,78 @@ let documentSymbol ~path =
...
+          |> List.partition (fun (_, loc) -> Utils.isTopLoc loc)
+        in
+        let fileRenames =
+          referencesToToplevelModules
+          |> List.map (fun (uri, _) ->
+                 let path = Uri2.toPath uri in
+                 let dir = Filename.dirname path in
+                 let ext = Filename.extension path in
+                 let sep = Filename.dir_sep in
+                 let newPath = dir ^ sep ^ newName ^ ext in
```
cristianoc, line 205: Better use `Filename.concat` instead of string concat with `sep`.

analysis/src/Commands.ml
```diff
@@ -176,6 +176,78 @@ let documentSymbol ~path =
...
+                 let newUri = Uri2.fromPath newPath in
+                 Protocol.
+                   {
+                     kind = `rename;
```
cristianoc, line 209: Nit: I think it's easier to expose a constructor function in the protocol that only takes 2 parameters: `oldUri` and `newUri` so the type of kind does not leak here.

cristianoc: Actually, I would just remove `kind` and use in the printing function only.

analysis/src/Protocol.ml
```diff
@@ -18,6 +18,20 @@ type location = {uri : string; range : range}
 
 type documentSymbolItem = {name : string; kind : int; location : location}
 
+type renameFile = {kind : [`rename]; oldUri : string; newUri : string}
```
cristianoc, line 21: Suggestion: remove `kind`.

analysis/src/Protocol.ml
```diff
@@ -65,3 +79,35 @@ let stringifyDocumentSymbolItem i =
...
     (Json.escape i.name) i.kind
     (stringifyLocation i.location)
+
+let stringifyRenameFile rf =
+  Printf.sprintf {|{
+  "kind": "%s",
+  "oldUri": "%s",
+  "newUri": "%s"
+}|}
+    (match rf.kind with `rename -> "rename")
```
cristianoc, line 89: this can go once field `kind` is removed, but the printing of "rename" stays.

analysis/src/Utils.ml
```diff
@@ -7,6 +7,12 @@ let topLoc fname =
     loc_ghost = false;
   }
 
+let isTopLoc (loc : Warnings.loc) =
```
cristianoc, line 10: It might be possible to clean this up further, but looks OK for now.

cristianoc:
Indeed, it looks great!
Left some minor comment as review.

amiralies:
Thanks, I was going to bed. I see you already fixed them

PR #260, opened 2021-05-29
Fix formatting textedit range

### LINKED ISSUE #259: Unexpected formatter behavior when there are redundant lines
try to format this
```rescript
let a =










  2
```

file changes to:
```rescript
let a = 2

  2
```

amiralies:
Closes #259

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -398,9 +398,13 @@ function format(msg: p.RequestMessage): Array<m.Message> {
         extension === c.resiExt
       );
       if (formattedResult.kind === "success") {
+        let max = code.length;
         let result: p.TextEdit[] = [
           {
-            range: utils.getSourceRange(code),
+            range: {
+              start: { line: 0, character: 0 },
+              end: { line: max, character: max },
+            },
             newText: formattedResult.result,
           },
         ];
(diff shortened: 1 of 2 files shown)


amiralies:
Trivial fix would be using file content's length (before formatting)

cristianoc:
This also seems to fix:
```ts
              end: { line: max+1, character: max },
```

amiralies:
yeah we can use file length. but note that we should use current text length instead of formatted text length.
(think of  TextEdit as a range you're selected in editor and gonna replace that with newText)

cristianoc:
A fixed large number would also work I guess, thought it looks less clean.

cristianoc:
As long as it's simple and it works, no preferences.

cristianoc:
Actually let me double check: are we taking about formatting the whole file?
Or is the idea to allow formatting a smaller part in isolation.

amiralies:
We just support formatting the whole file.
But lsp doesn't have a method of replacing the whole file text. it accepts a list of edits.
what we need is return a single `TextEdit` in a list with a `range` covering the whole file (before formatting) and `newText` which is the formatted result.

amiralies:
I've updated to use current file's length as max range. I guess it's okay for now. we can change it later if we faced performance issues or other problems,

cristianoc:
Great

PR #263, opened 2021-06-05
Set

cristianoc:


cristianoc:
Relevant parts moved to https://github.com/rescript-lang/rescript-vscode/pull/264

Issue #176, opened 2021-04-27
Outline stale


chenglou:
`documentSymbol` event is sent to us upon file refresh and keystroke, but not on save. So our outline view is always one step behind when we change a binding or module. Not sure what to do

amiralies:
This is known issue for vscode, I'm not sure if we can do better our side aside from compiling on edit which is not ideal.
see this https://github.com/microsoft/vscode/issues/108722.
I think people would use this feature mostly when they are reading the code so I think it's fine to ship this as it is.

zth:
This has been fixed recently by @cristianoc and I think we can close this?

PR #272, opened 2021-06-10
Enable document symbol provider

amiralies:


amiralies:
See my comment https://github.com/rescript-lang/rescript-vscode/issues/176#issuecomment-850803668

PR #278, opened 2021-06-22
Tests for jump to definition for interfaces and implementations.

cristianoc:
```rescript
// Foo.res
let x = 10
//  ^ when clicked, jump to Foo.resi
```

```rescript
// Foo.resi
let x : int
//  ^ when clicked, jump to Foo.res
```

```rescript
// User.res
Foo.x
//  ^ when clicked, jump to Foo.res
```


cristianoc:
@amiralies any thoughts on this?

amiralies:
Looks good to me. I think ocaml-lsp works like this too.
Maybe in future we can implement the `go to decleration` request as a direct way of going to interface from usage site.

cristianoc:
Go to declaration can be achieved via go to type definition.

Issue #107, opened 2021-04-20
Feature Request: A Button or Command to open corresponding .bs.js file in split pane (and track changes to current .res file)


huntwj:
The documentation for ReScript is constantly telling me to keep the generated JS open in another pane. It would be great if this extension maintained that pane for me so as I swap between .res files, the corresponding .bs.js file is always on the right.

zth:
Closing this as a variant of this feature was implemented in https://github.com/rescript-lang/rescript-vscode/pull/294

PR #294, opened 2021-08-10
Command to open compiled file

fhammerschmidt:
This implements #107 

Press CMD+Shift+P and select `ReScript: Open the compiled JS file for this implementation file.`
It should open the corresponding `.bs.js`, `.js`, `.mjs` in the correct directory (whether in-source or not).

Unfortunately, this requires to parse `bsconfig.json`. I did not know how to account for an array with more than one element of package-specs, so for now it just takes the first one, but I guess it is better to use an object anyway, as [suggested by the docs](https://rescript-lang.org/docs/manual/latest/build-configuration#package-specs).

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -45,3 +45,7 @@ export let resExt = ".res";
 export let resiExt = ".resi";
 export let cmiExt = ".cmi";
 export let startBuildAction = "Start Build";
+
+// bsconfig defaults according configuration schema (https://rescript-lang.org/docs/manual/latest/build-configuration-schema)
+export let bsconfigModuleDefault = "commonjs";
+export let bsconfigSuffixDefault = ".js";
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -187,37 +187,42 @@ export const toCamelCase = (text: string): string => {
 };
 
 const readBsConfig = (projDir: p.DocumentUri) => {
-  let bsconfigFile = fs.readFileSync(
-    path.join(projDir, c.bsconfigPartialPath),
-    { encoding: "utf-8" }
-  );
+  try {
+    let bsconfigFile = fs.readFileSync(
+      path.join(projDir, c.bsconfigPartialPath),
+      { encoding: "utf-8" }
+    );
 
-  return JSON.parse(bsconfigFile);
+    let result = JSON.parse(bsconfigFile);
+    return result;
+  } catch (e) {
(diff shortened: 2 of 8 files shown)


fhammerschmidt:
Yes! Just did that and also factored out all the commands into separate files.
Bonus: You can open the compiled file both with the CMD+SHIFT+P menu and the following icon in the navigation menu: 
image

Let me know if you think there is a better icon for this action: https://code.visualstudio.com/api/references/icons-in-labels#icon-listing

server/src/utils.ts
```diff
@@ -186,16 +186,20 @@ export const toCamelCase = (text: string): string => {
     .replace(/(\s|-)+/g, "");
 };
 
+const readBsConfig = (projDir: p.DocumentUri) => {
+  let bsconfigFile = fs.readFileSync(
+    path.join(projDir, c.bsconfigPartialPath),
+    { encoding: "utf-8" }
+  );
+
+  return JSON.parse(bsconfigFile);
```
cristianoc, line 195: How about catching the parse error inside this function. So there's no need to wrap the callers in a try block. Assuming that was the only reason to use a try block in the caller, which I think is true at least for the previous uses of bsconfig parse.

fhammerschmidt:
IMO yes, I used it all day while working on a big monorepo and it always worked well so far.

server/src/constants.ts
```diff
@@ -45,3 +45,7 @@ export let resExt = ".res";
 export let resiExt = ".resi";
 export let cmiExt = ".cmi";
 export let startBuildAction = "Start Build";
+
+// bsconfig defaults according configuration schema (https://rescript-lang.org/docs/manual/latest/build-configuration-schema)
+export let bsconfigModuleDefault = "commonjs";
```
cristianoc, line 50: Tested on the sample project in `analysis/tests`: it fails to find `lib/js/...` as it's looking into `lib/commonjs/...`.
The project's bsconfig does not speicify any of the relevant data, so relies on all the defaults.

cristianoc (changes requested):
See comments.

fhammerschmidt:
I tried to be more explicit with the modulename -> foldername resolving. 
Also added another case when you directly set the "package-specs" field to a string.
Tested it with the example project now as well (and I tested all possible module format strings).

cristianoc (approved):
Fantastic. Merging.

PR #325, opened 2021-12-02
Fix escaped backtick highlighting in string literal (#324)

### LINKED ISSUE #324: broken syntax highlighting when escaping a backtick in string literal
example: 
```res
let regex = `\``

let foo = 10
```
It does compile correctly as seen [here](https://rescript-lang.org/try?code=DYUwLgBATiDmIA8IF4IAMA6a0CgekgDMB7YlCARgAYcg).

If it's not the right place to report it, let me know!

tsnobip:
should fix #324:
```res
let regex = `\``

let foo = 10
```

Should I first create a PR in [rescript-lang/rescript-sublime](https://github.com/rescript-lang/rescript-sublime)?

### DIFF
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -127,7 +127,7 @@
         {
           "name": "string.template",
           "begin": "([a-z_][0-9a-zA-Z_]*)?(`)",
-          "end": "[^\\]`",
+          "end": "[^\\\\]`",
           "beginCaptures": {
             "1": {
               "name": "variables.annotation"
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -127,7 +127,7 @@
         {
           "name": "string.template",
           "begin": "([a-z_][0-9a-zA-Z_]*)?(`)",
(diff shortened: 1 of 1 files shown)


grammars/rescript.tmLanguage.json
```diff
@@ -127,7 +127,7 @@
         {
           "name": "string.template",
           "begin": "([a-z_][0-9a-zA-Z_]*)?(`)",
-          "end": "`",
+          "end": "[^\\]`",
```
cristianoc, line 130: Should this be ```"[^\\\\]`"``` to json encode the escaped backslash.

tsnobip: why so many backslash? `\\` is already an escaped backslash, isn't it?

tsnobip: or is it because we need to add another layer of escaping because it's a json? Then yes!

tsnobip: yeah I guess this makes sense:
```suggestion
          "end": "[^\\\\]`",
```

cristianoc:
Thanks!

PR #334, opened 2021-12-30
Dead code analysis editor mode via reanalyze

zth:
This implements a "Dead code analysis mode". When activated, the extension will run reanalyze continuously as the ReScript compiler runs, and report back any errors found directly into the editor. It'll also set up code actions leveraging the suggestions reanalyze gives for how to suppress the issues reported.

See the video below for more details.

https://user-images.githubusercontent.com/1457626/147784733-4eabe71f-7b20-441b-a9b7-bedda07b4d4e.mp4

Thank you @cristianoc for guidance both around reanalyze and details around how the extension does things.

Happy to hear any feedback or comments, and whether this is something you're interested in merging.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -115,6 +115,14 @@ let deleteProjectDiagnostics = (projectRootPath: string) => {
     projectsFiles.delete(projectRootPath);
   }
 };
+let sendCompilationFinishedMessage = () => {
+  let notification: m.NotificationMessage = {
+    jsonrpc: c.jsonrpcVersion,
+    method: "rescript/compilationFinished",
+  };
+  
+  send(notification);
+};
 
 let compilerLogsWatcher = chokidar
   .watch([], {
@@ -124,6 +132,7 @@ let compilerLogsWatcher = chokidar
   })
   .on("all", (_e, changedPath) => {
     sendUpdatedDiagnostics();
+    sendCompilationFinishedMessage();
   });
 let stopWatchingCompilerLog = () => {
   // TODO: cleanup of compilerLogs?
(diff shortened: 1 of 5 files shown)


server/src/server.ts
```diff
@@ -115,6 +115,14 @@ let deleteProjectDiagnostics = (projectRootPath: string) => {
     projectsFiles.delete(projectRootPath);
   }
 };
+let sendCompilationFinishedMessage = () => {
+  let notification: m.NotificationMessage = {
+    jsonrpc: c.jsonrpcVersion,
+    method: "rescript/compilationFinished",
+  };
+  
+  send(notification);
```
cristianoc, line 124: Is there a special reason for sending a custom notification rather than just calling ` customCommands.deadCodeAnalysisWithReanalyze` here?

zth: Yes, the reason for the notification itself is because my idea is to keep this feature contained inside of the extension code rather than the server. That has a few reasons:
* I didn't want to introduce state into the server (keeping track of whether the dead code analysis mode is activated or not)
* The code that controls the status bar item lives inside of the extension code itself, and not the server. I'm unsure whether this could live inside of the server at all and communicate via the LS protocol, but even if it could, I'm not sure whether I personally think it should live in the server or not. It's a pretty VSCode specific thing, at least at this point.
* It felt cleaner to separate the dead code diagnostics from the regular ReScript compiler diagnostics. That means it's easy to clear them independently of other diagnostics (which is done for example when stopping the dead code analysis mode, even if not all errors has been fixed).
* This way, any client using the language server can react to compilation finishing.

I'm very much open to trying to move all/most of the functionality into the server itself though, if we think that's a better place for it to live.

cristianoc: Makes perfect sense, thanks for the explanation.

cristianoc (approved):

Issue #345, opened 2022-01-20
Version 1.2.0 breaks syntax highlighting after backtick


cknitt:
1.1.3:
<img width="178" alt="Bildschirmfoto 2022-01-20 um 13 27 43" src="https://user-images.githubusercontent.com/591384/150339031-92128a02-c253-4b04-9476-f7e4562c4b85.png">

1.2.0:
<img width="187" alt="Bildschirmfoto 2022-01-20 um 13 27 15" src="https://user-images.githubusercontent.com/591384/150339059-66288974-6788-4a97-bb79-fd640961821c.png">

cristianoc:
Coming from here:
https://github.com/rescript-lang/rescript-vscode/pull/325

Looks like and analogue of this question has been asked before with no satisfactory suggestions:
https://stackoverflow.com/questions/59303371/is-there-any-way-to-skip-the-escaped-quote-inside-the-double-quoted-textmate-rul

Unless someone has an alternative suggestion, the way forward is to revert https://github.com/rescript-lang/rescript-vscode/pull/325
and live with the fact that escaped backticks break highlighting.

cristianoc:
Turns out one of the stack overflow suggestions, to use negative-look-behind, seems to work (though the response states it does not):
https://github.com/rescript-lang/rescript-vscode/pull/346

Tried on these patterns:
```rescript
let a = "abc"

Js.log(` \` ccc ${a} ddd  `)

Js.log(`a = ${a}`)

Js.log(``)

let b = 34
```

@cknitt would you be able to test that, and check that no other issues arise?

cknitt:
Yes, I will test against one of our codebases.

PR #346, opened 2022-01-20
Fix issue with highlighting of templates.

cristianoc:
Fix issue with highlighting of templates introduced in https://github.com/rescript-lang/rescript-vscode/pull/325

and reported in

https://github.com/rescript-lang/rescript-vscode/issues/345


cristianoc:
Reference: https://www.regular-expressions.info/lookaround.html

PR #357, opened 2022-01-29
grammar: remove to and downto as keywords

noverby:
They are not reserved according to: https://rescript-lang.org/docs/manual/latest/reserved-keywords


amiralies:
I think we also need to change for's grammar to highlight to/downto in this PR.

zth:
Just dropping a note that this will be at least partially handled by https://github.com/rescript-lang/rescript-vscode/pull/367. The reason they've remained keywords in the grammar is because they help make `for` loops look OK even if they're not technically reserved. The linked PR will address that.

PR #358, opened 2022-02-01
Normalize relative paths by avoiding prefixes of the form `./something`.

cristianoc:


mellson:
@cristianoc I just tested this locally on my machine and the suggested fix does indeed work! Thanks 👍🏻

PR #364, opened 2022-02-23
Fix creating interface files for namespaced packages

fhammerschmidt:
I found out that the compiler appends `-MyNamespace` to `cmi` file names if the `"namespace"` attribute is set in `bsconfig.json`. This case is now handled correctly by the `create-interface` command.

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -188,20 +188,31 @@ export const toCamelCase = (text: string): string => {
 
 export const getNamespaceNameFromBsConfig = (
   projDir: p.DocumentUri
-): string | null => {
-  let bsconfigFile = fs.readFileSync(
-    path.join(projDir, c.bsconfigPartialPath),
-    { encoding: "utf-8" }
-  );
+): execResult => {
+  try {
+    let bsconfigFile = fs.readFileSync(
+      path.join(projDir, c.bsconfigPartialPath),
+      { encoding: "utf-8" }
+    );
 
-  let bsconfig = JSON.parse(bsconfigFile);
+    let bsconfig = JSON.parse(bsconfigFile);
+    let result = "";
 
-  if (bsconfig.namespace === true) {
-    return toCamelCase(bsconfig.name);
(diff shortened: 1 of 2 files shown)


server/src/utils.ts
```diff
@@ -180,6 +180,33 @@ export let replaceFileExtension = (filePath: string, ext: string): string => {
...
+    { encoding: "utf-8" }
+  );
+
+  let bsconfig = JSON.parse(bsconfigFile);
+
+  if (bsconfig.namespace) {
+    if (typeof bsconfig.namespace === "string") {
+      return toCamelCase(bsconfig.namespace);
+    } else {
+      return toCamelCase(bsconfig.name);
```
cristianoc, line 203: Nit: I think this code path is triggered with `"namespace": false`

cristianoc (changes requested):
Thanks for the fix!
One nit left in the code.

server/src/utils.ts
```diff
@@ -180,6 +180,31 @@ export let replaceFileExtension = (filePath: string, ext: string): string => {
...
+
+export const getNamespaceNameFromBsConfig = (
+  projDir: p.DocumentUri
+): string | null => {
+  let bsconfigFile = fs.readFileSync(
+    path.join(projDir, c.bsconfigPartialPath),
+    { encoding: "utf-8" }
+  );
+
+  let bsconfig = JSON.parse(bsconfigFile);
```
cristianoc, line 197: I don't have a particular suggestion here, just noticing that the compiler accepts comments in `bsconfig.json`, which will cause a parse error here.
Perhaps just avoid crashing the extension would be OK.

cristianoc (approved):

server/src/utils.ts
fhammerschmidt: I made it a bit more user-friendly and show a separate message when the JSON cannot be parsed.
I tested it out on a bsconfig with comment and it shows the error message as expected.

Issue #362, opened 2022-02-22
Dead code analysis - all warnings are marked as dead code


zth:
Most warnings the dead code analysis produces are indeed about dead code. However, some are about redundant optional arguments, etc. Currently, all warnings in the editor are highlighted as dead code, even if it isn't. We should differentiate properly between that when highlighting the dead code in the editor.

PR #366, opened 2022-03-08
Dce touchups

zth:
- Expose the "Stop dead code analysis" command to the user so it can be run via the command palette. We already have the start command exposed, so I think it makes sense to expose the stop command as well, in addition to the stop button in the status bar.
- Do not mark "redundant optional argument" warnings as dead/unused code, since it's not really. Closes https://github.com/rescript-lang/rescript-vscode/issues/362


cristianoc:
Thanks!

PR #377, opened 2022-04-01
Add GH action to package VSIX file for testing

fhammerschmidt:
I thought this would be useful for testing, since it makes it easier for everybody to just download and install the VSIX directly after any master or PR build.

This adds about half a minute of build time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,11 +120,15 @@ jobs:
     - name: Cleanup tar file
       run: rm binary.tar
       working-directory: ./server/analysis_binaries
+
+    - name: Store short commit SHA for filename
+      id: vars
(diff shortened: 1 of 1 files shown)


zth:
This is awesome! I'm going to test the vsix produced by the action soon. @cristianoc looking good to you too?

cristianoc:
Looks great!

cristianoc:
@fhammerschmidt just noticed that it's not easy to un-install a `.vsix` extension when the version number, e.g. `1.2.1` is the same as a published extension.
The only way I've found to revert the downloaded extension is to first completely delete from disk `1.2.1` and then install again.
While switching to an older version and back to `1.2.1` gives me back the manually installed one, not the one from the marketplace.

PR #379, opened 2022-04-03
dedicated parsing for error messages related to interface files

zth:
This detects a few diagnostics related to interface files, and does the following:

### One diagnostic for the implementation, one for the interface file
Previously the diagnostic would point to the implementation only. This changes it to produce one diagnostic for the implementation file, and one for the interface file. This will hopefully make it a bit easier to find where the actual issue is located.

image

### Clean up error messages
The error message the compiler produces is quite dense. This produces slightly cleaned up, separate error messages for the implementation and the interface respectively.

Before (one message only):
```
The implementation /Users/zth/git/rescript-intro/src/Interface.res
       does not match the interface src/interface-RescriptIntro.cmi:
       ...
       In module Test:
       Values do not match:
         let bar: (
  ~some: 'a,
  ~thing: 'b,
  ~another: 'c,
  ~andSome: 'd,
  ~anotherOne: 'e,
  ~andAgain: 'f,
  ~andAgain1: 'g,
  ~andAgain2: 'h,
  ~andAgain3: 'i,
) => int
       is not included in
         let bar: (
  ~some: int,
  ~thing: int,
  ~another: string,
  ~andSome: bool,
  ~anotherOne: bool,
  ~andAgain: bool,
  ~andAgain1: bool,
  ~andAgain2: bool,
) => int
       File "/Users/zth/git/rescript-intro/src/Interface.resi", line 4, characters 3-187:
         Expected declaration
       File "/Users/zth/git/rescript-intro/src/Interface.res", line 6, characters 7-10:
         Actual declaration
```

...and in this PR - error message for the implementation:
```
This implementation does not match the type signature defined in the interface file ("Interface.resi"). The interface defines the following type signature:

let bar: (
  ~some: int,
  ~thing: int,
  ~another: string,
  ~andSome: bool,
  ~anotherOne: bool,
  ~andAgain: bool,
  ~andAgain1: bool,
  ~andAgain2: bool,
) => int

...but this implementation's type signature is:

let bar: (
  ~some: 'a,
  ~thing: 'b,
  ~another: 'c,
  ~andSome: 'd,
  ~anotherOne: 'e,
  ~andAgain: 'f,
  ~andAgain1: 'g,
  ~andAgain2: 'h,
  ~andAgain3: 'i,
) => int
```

And for the interface file:
```
This type signature does not match the implementation. The implementation in "Interface.res" currently has the following type signature:

let bar: (
  ~some: 'a,
  ~thing: 'b,
  ~another: 'c,
  ~andSome: 'd,
  ~anotherOne: 'e,
  ~andAgain: 'f,
  ~andAgain1: 'g,
  ~andAgain2: 'h,
  ~andAgain3: 'i,
) => int
```

The original message is still printed by the compiler in the terminal.

### More
This is also a preparation for experimenting with code actions related to interface file diagnostics.

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -474,6 +474,213 @@ let parseFileAndRange = (fileAndRange: string) => {
   };
 };
 
+// Extracts file info from: File "/Users/whatever/project/src/Interface.resi",
+// line 2, characters 1-23
+//
+// Note that "characters" above might actually span multiple lines, but the
+// compiler just reports the start line and character, and we need to calculate
+// the final line/character ourselves via offset if we need it.
+let parseFileInfo = (fileInfoLine: string) => {
+  let regex = /File "(.+)", line (\d+), characters ([\d-]+):/;
+  let trimmedFileInfoLine = fileInfoLine.trim();
+  let match = trimmedFileInfoLine.match(regex);
+  if (match === null) {
+    return null;
+  }
+  let [_, file, line, characters] = match;
+  if (file == null || line == null || characters == null) {
+    return null;
+  }
+  let [startChar, endChar] = characters.split("-");
+  if (startChar == null || endChar == null) {
+    return null;
+  }
+  let lineAdjusted = parseInt(line) - 1;
+  return {
+    file: pathToURI(file),
+    range: {
+      start: {
+        line: lineAdjusted,
+        character: parseInt(startChar) - 1,
+      },
+      end: {
+        line: lineAdjusted,
+        character: parseInt(endChar) - 1,
+      },
+    },
+  };
+};
+
+let handleInterfaceFileDiagnostics = (
+  diagnosticMessage: string[],
+  result: filesDiagnostics,
+  parsedDiagnostic: parsedDiagnostic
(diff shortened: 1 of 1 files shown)


amiralies:
Although I think it's a great change and delivers better UX to editor users, the conflict between compiler error msg and editor error message can be confusing. I prefer this to be done somewhere in the compiler.

zth:
> Although I think it's a great change and delivers better UX to editor users, the conflict between compiler error msg and editor error message can be confusing. I prefer this to be done somewhere in the 

> Although I think it's a great change and delivers better UX to editor users, the conflict between compiler error msg and editor error message can be confusing. I prefer this to be done somewhere in the compiler.

Are you up for taking a stab at implementing it in the compiler?

amiralies:
I think we can use super erros implemenation to deliver such behavior iirc. i should take a look.

zth:
> I think we can use super erros implemenation to deliver such behavior iirc. i should take a look.

That would be great if you had a look at that! In fact, there's quite a lot of error messages that needs cleaning up. So if you figure out a good workflow it'd be nice to also look at a few more of them.

Issue #380, opened 2022-04-04
Keep regex-based highlighting where semantic highlighting is not available


fhammerschmidt:
I don't understand why the grammar needed to be cleaned up. 

1. VSCode has a lot of places where code is displayed without semantic highlighting, like in the working tree view or in hover snippets.
2. Many themes just don't enable support for it [(default is disabled)](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-coloring-in-color-themes).
3. Sometimes you get a flash, which is pretty obvious with lots of module names which get painted from one color to the other. While we cannot prevent that, it would be less noticeable that way.
4. No bsconfig means no highlighting now. While formatting still works.

Obviously, it's not correct in many cases, but let's please find some middle ground here.
It would already help a ton just to put the module parts back in.

zth:
Yes, we agree, this is a valid point. Let's tackle this in two ways:

1. Restore a subset of what was previously in the grammar, to clean up the "worst offenders". Ensure that anything colored by the grammar directly will get the same color as semantic highlighting would give it (so we avoid flashes where we can).
2. Implemented the `range` method for semantic highlighting, and see whether that helps coloring in hovers etc. Some background: we've currently only implemented the method for semantically highlighting a full file. But, there's an additional method, `range`, that allows for coloring independent chunks of code. We'll implement that and see whether that helps with some of the lack of coloring too.

@fhammerschmidt would you be interested in taking a stab at point 1? We can go back and forth, testing together, etc.

fhammerschmidt:
Yeah, will try to submit a (draft) PR tonight.

I think the most important parts got lost in these two commits, so in the first iteration I will just restore the module/JSX parts:

- https://github.com/rescript-lang/rescript-vscode/commit/0e56b5e9f061ef06921e5fbde64590fecf77ec9b
- https://github.com/rescript-lang/rescript-vscode/commit/dd9c37566522df7ca137f82798c489518309c312

zth:
Closing this as all relevant changes has landed in master. Thanks @fhammerschmidt !

PR #381, opened 2022-04-05
Restore some modules and JSX grammar

fhammerschmidt:
As discussed in #380, this PR restores some grammar that got cleaned up in https://github.com/rescript-lang/rescript-vscode/commit/0e56b5e9f061ef06921e5fbde64590fecf77ec9b and https://github.com/rescript-lang/rescript-vscode/commit/dd9c37566522df7ca137f82798c489518309c312.

Here's how it looks with this PR (semantic highlighting on the right for comparison):

image
image

### DIFF
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -350,8 +350,133 @@
       ]
     },
     "jsx": {
-      "match": "<>|</>|</|/>",
-      "name": "punctuation.definition.tag"
+      "patterns": [
+        {
+          "match": "<>|</>|</|/>",
+          "name": "punctuation.definition.tag"
+        },
+        {
+          "match": "</([A-Z_][0-9a-zA-Z_]*)",
+          "captures": {
+            "0": {
+              "name": "punctuation.definition.tag"
+            },
+            "1": {
+              "name": "entity.name.namespace"
+            }
+          }
+        },
+        {
+          "match": "</([a-z_][0-9a-zA-Z_]*)",
+          "captures": {
+            "0": {
+              "name": "punctuation.definition.tag"
+            },
+            "1": {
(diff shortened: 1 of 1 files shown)


cristianoc:
@fhammerschmidt thank you for looking into this.

See https://github.com/rescript-lang/rescript-vscode/pull/367/commits/55b91665e2a9d91f631f48f129074578527dfa3c#diff-4398e3b2ab3e7e7e70a93409fa893abf68608d45a5ad1ac7faaf888dac77005dR768
where previously modules were emitted as namespaces, but were then changed to class.

That was because some themes, such as Dracula, do not support namespace. For all the themes that do support namespaces that I have seen, they show namespaces the same as they do class, hence the change. But happy to get more input on what other themes do.

zth:
@cristianoc so ideally, `namespace` is to be replaced with a `class` tag in this PR, right?

fhammerschmidt:
Yup, Dracula for instance requires `class` to highlight the modules.

Before: 
image

After:
image

cristianoc (approved):
Thanks a lot for doing this.

cristianoc:
Anything else to put back into the grammar, or ready to merge?

fhammerschmidt:
The only thing that is different compared to the old style is that non-polymorphic variants are the same color as other values. Not sure if it's feasible to do that in TextMate now, otherwise I think it's fine.

cristianoc:
> The only thing that is different compared to the old style is that non-polymorphic variants are the same color as other values. Not sure if it's feasible to do that in TextMate now, otherwise I think it's fine.

Those can be restored too. With semantic highlighting they are currently emitted as `enumMember`.

cristianoc:
Looks like it's ready to go.

PR #383, opened 2022-04-06
Range based semantic highlighting

zth:
As noted in https://github.com/rescript-lang/rescript-vscode/issues/380


zth:
For posterity - we decided to not implement range based semantic highlighting. 

After testing, we came to the conclusion that range based highlighting seems to be used as a performance optimization when opening a file. 

But, in our solution the bottleneck isn't the performance of figuring out the semantic tokens, but rather the throttling VSCode does itself when asking for tokens. So, implementing this would currently bring no benefit.

Closing for now, we can revisit in the future.

PR #385, opened 2022-04-07
relax requirement of files being in a ReScript project for formatting

zth:
This lifts the current requirement for files to be located in an actual ReScript project to be formatted. It falls back to formatting using the vendored parser pretty printer in the extension itself if the file isn't located in a ReScript project.

### DIFF
--- a/analysis/src/Cli.ml
+++ b/analysis/src/Cli.ml
@@ -56,6 +56,10 @@ Options:
 
     ./rescript-editor-analysis.exe createInterface src/MyFile.res lib/bs/src/MyFile.cmi
 
+  format: print to stdout the formatted version of the provided file
+
+    ./rescript-editor-analysis.exe format src/MyFile.res
+
   test: run tests specified by special comments in file src/MyFile.res
 
     ./rescript-editor-analysis.exe test src/src/MyFile.res
@@ -86,6 +90,7 @@ let main () =
   | [_; "createInterface"; path; cmiFile] ->
     Printf.printf "\"%s\""
       (Json.escape (CreateInterface.command ~path ~cmiFile))
+  | [_; "format"; path] -> Commands.format ~path
   | [_; "test"; path] -> Commands.test ~path
   | args when List.mem "-h" args || List.mem "--help" args -> prerr_endline help
   | _ ->
--- a/analysis/src/Cli.ml
+++ b/analysis/src/Cli.ml
@@ -90,7 +90,8 @@ let main () =
   | [_; "createInterface"; path; cmiFile] ->
     Printf.printf "\"%s\""
       (Json.escape (CreateInterface.command ~path ~cmiFile))
-  | [_; "format"; path] -> Commands.format ~path
+  | [_; "format"; path] ->
+    Printf.printf "\"%s\"" (Json.escape (Commands.format ~path))
   | [_; "test"; path] -> Commands.test ~path
   | args when List.mem "-h" args || List.mem "--help" args -> prerr_endline help
   | _ ->
--- a/analysis/src/Commands.ml
+++ b/analysis/src/Commands.ml
@@ -250,6 +250,19 @@ let rename ~path ~line ~col ~newName =
   in
   print_endline result
 
+let format ~path =
+  if Filename.check_suffix path ".res" then
+    let {Res_driver.parsetree = structure; comments} =
+      Res_driver.parsingEngine.parseImplementation ~forPrinter:true
+        ~filename:path
+    in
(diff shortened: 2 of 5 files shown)


analysis/src/Commands.ml
```diff
@@ -250,6 +250,19 @@ let rename ~path ~line ~col ~newName =
   in
   print_endline result
 
+let format ~path =
+  if Filename.check_suffix path ".res" then
+    let {Res_driver.parsetree = structure; comments} =
+      Res_driver.parsingEngine.parseImplementation ~forPrinter:true
+        ~filename:path
+    in
+    Res_driver.printEngine.printImplementation 80 path comments structure
```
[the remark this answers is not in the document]
cristianoc: !Res_cli.ResClflags.width

analysis/src/Commands.ml
```diff
@@ -250,6 +250,19 @@ let rename ~path ~line ~col ~newName =
...
+  else if Filename.check_suffix path ".resi" then
+    let {Res_driver.parsetree = structure; comments} =
+      Res_driver.parsingEngine.parseInterface ~forPrinter:true ~filename:path
+    in
+    Res_driver.printEngine.printInterface 80 path comments structure
```
cristianoc, line 264: !Res_cli.ResClflags.width

analysis/src/Commands.ml
```diff
@@ -250,6 +250,19 @@ let rename ~path ~line ~col ~newName =
...
 
+let format ~path =
+  if Filename.check_suffix path ".res" then
+    let {Res_driver.parsetree = structure; comments} =
+      Res_driver.parsingEngine.parseImplementation ~forPrinter:true
+        ~filename:path
+    in
+    Res_driver.printEngine.printImplementation 80 path comments structure
+  else if Filename.check_suffix path ".resi" then
+    let {Res_driver.parsetree = structure; comments} =
```
cristianoc, line 261: nit: `Res_driver.parsetree = signature`

analysis/src/Commands.ml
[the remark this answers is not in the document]
fhammerschmidt: It's 100: https://github.com/rescript-lang/syntax/blob/2d4fef51d07312957f907e6d1e19a40618058518/src/res_multi_printer.ml#L3 

But in the playground it seems to be 80? It's weird.

server/src/utils.ts
```diff
@@ -128,7 +144,8 @@ export let formatUsingValidBscNativePath = (
 export let runAnalysisAfterSanityCheck = (
   filePath: p.DocumentUri,
   args: Array<any>,
-  projectRequired=false
+  projectRequired = false,
+  parseJson = true
```
cristianoc, line 148: The `createInterface` command, for example, does the json escape in `Cli.ml`, so we don't need to special case the command here.

analysis/src/Cli.ml
```diff
@@ -86,6 +90,7 @@ let main () =
   | [_; "createInterface"; path; cmiFile] ->
     Printf.printf "\"%s\""
       (Json.escape (CreateInterface.command ~path ~cmiFile))
```
cristianoc, line 92: see: `Json.escape`.

cristianoc (changes requested):

server/src/utils.ts
```diff
@@ -104,16 +104,32 @@ export let formatUsingValidBscNativePath = (
...
+        "never",
+        "-format",
+        formatTempFileFullPath,
+      ]);
+      return {
+        kind: "success",
+        result: result.toString(),
+      };
+    } else {
+      let result = runAnalysisAfterSanityCheck(
```
cristianoc, line 121: This is kind of a strange code path for a function called `formatUsingValidBscNativePath`.
One possibility would be to make the `formatUsingValidBscNativePath` function thinner, and avoid calling it at all unless there's a valid `bscNativePath`.

For example, instead of passing `isInterface` as a parameter, and creating the temp file internally, one could create the temp file externally, and pass it as a parameter.
Then move logic that calls the parser to the caller's site.

zth: What do you think about simply renaming it `formatCode`? That feels more in line with what it does now.

cristianoc: Go for it!

cristianoc (approved):

cristianoc (changes requested):
Actually, it currently formats even if there are syntax errors.
Needs to check for errors coming form the parser, and do no formatting in that case.

cristianoc (approved):
Looking good!

cristianoc:
Looking good. OK to merge?

PR #389, opened 2022-04-13
[poc] detect and transform JSX prop errors to something more readable

zth:
_This is a PoC, and should probably be implemented for real in the compiler instead of here._

This PR detects and modifies errors that come from trying to set a prop on a JSX DOM element that does not exist. Example:

```rescript
// It's "onChange", not "onchange"
<div onchange={_ => {....
```

When doing so currently the error produced by the compiler prints the full `domProps` record (which has almost 500 props), and in general doesn't really help the user figuring out what went wrong (and what it can do to fix it). The current error looks something like this:

```
...
 ~?typeof: string,
  ~?vocab: string,
  ~?dangerouslySetInnerHTML: {"__html": string},
  ~?suppressContentEditableWarning: bool,
) => ReactDOM.Props.domProps
This argument cannot be applied with label ~onchange
```

And after this fix, that same error will look like this:
```
This JSX element does not take the prop `onchange`. 
Did you mean `onChange`?

`onChange: ReactEvent.Form.t => unit=?`
```

Here's a GIF showing what it looks like in the editor:

rescript-intro-1649834066802

A few extra things to note:
- This only applies for _DOM JSX elements_, not for custom components like `<MyComponent someProp />`. Those errors are a lot better than the DOM JSX errors (although they could probably use some cleaning up eventually too).
- The full, underlying error is still displayed in the terminal (or wherever the ReScript compiler is run).
- The error message itself will be upstreamed to the compiler if users end up liking this representation.
- The loc's for the error does not point to the label itself because of reasons, but in the future if it does, we can also produce a quick fix code action here to replace the prop name with the hint provided.
- A technical note - I vendored a small and self contained fuzzy match implementation I found, so we can continue having 0 dependencies.

### DIFF
--- a/server/src/processDiagnosticMessage.ts
+++ b/server/src/processDiagnosticMessage.ts
@@ -1,0 +1,77 @@
+import { filter } from "./vendor/fuzzy";
+
+export let processDiagnosticMessage = (diagnosticMessage: string[]) => {
+  // TODO: This should be upstreamed to the compiler at some point soon.
+  // Heuristic for detecting whether this is a JSX call.
+  if (
+    diagnosticMessage[0]?.trim() ===
+      "The function applied to this argument has type" &&
+    // We assume that we can detect JSX calls by checking if they have an
+    // optional key label, and that it returns `ReactDOM.Props.domProps`.
+    diagnosticMessage.some((line) => line.trim().includes("~?key: string,")) &&
+    diagnosticMessage.some((line) =>
+      line.trim().endsWith(") => ReactDOM.Props.domProps")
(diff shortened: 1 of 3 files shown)


cristianoc:
This is sweet. Should try again with the new jsx based on records with optional fields, see if we get this for free.

zth:
Closing this as we're not moving forward with this inside the extension itself. We'll see if the new JSX PPX with optional record fields improves the situation for free. If not, or if we want to enhance things further, we'll do it directly in the compiler, and not here.

PR #404, opened 2022-05-04
add command to switch implementation and interface

mattdamon108:
This PR is adding the new command `switch-impl-intf` which enables users to switch the implementation `*.res` to interface `*.resi` back and forth. Actually, we can find this same feature in the OCaml vscode extension too.

I hope this feature helps users to open the implementation file once after navigating to the interface file with the go to definition command.

I'm not sure whether the contribution is open to not members. If it is open, I'm going to add a button in the editor title menu.

https://user-images.githubusercontent.com/36963424/166794591-eb110a6f-1e2d-46d6-a625-886e236912c2.mov


client/src/commands/switch_impl_intf.ts
```diff
@@ -0,0 +1,50 @@
+import * as fs from "fs";
+import { LanguageClient, RequestType } from "vscode-languageclient/node";
+import { window } from "vscode";
+
+interface CreateInterfaceRequestParams {
+  uri: string;
+}
+
+let createInterfaceRequest = new RequestType<
```
fhammerschmidt, line 9: Maybe this one can be reused/imported from the `create_interface.ts` file?

fhammerschmidt:
Looks good, only had a small nitpick. Which icon will you use for the button?

client/src/commands/switch_impl_intf.ts
mattdamon108: Good point. I've fixed it. https://github.com/rescript-lang/rescript-vscode/pull/404/commits/40366f106adc2951ae74a79731ec8c356d712269

mattdamon108:
> Looks good, only had a small nitpick. Which icon will you use for the button?

@fhammerschmidt I don't have a particular skill to design an icon 🫣 I found the free svg repo in which we might pick one. https://www.svgrepo.com/svg/392618/arrow-large-update-direction-navigation I would be looking as below.

image

If the ReScript core team has a resource to create an icon for it, I love to use it.

fhammerschmidt:
I think the icon is fine, but I guess you would need to add an attribution to the README (CC license)?

zth:
I like this! 

One thing I thought of though - maybe we should prompt the user before creating the interface file, if it doesn't already exist? Let them decide whether to create (and then open) the interface file. I think it might be confusing otherwise, as in "oh did I already have an interface file for this file?".

mattdamon108:
@zth Very good point! I’ll add the step to ask user for creating the interface.

mattdamon108:
@fhammerschmidt I've added the button in the editor title menu. https://github.com/rescript-lang/rescript-vscode/pull/404/commits/14bcf09aa8150384a4b6d16f7a8007993daf4769
@zth I've added the popup(`window.showInformationMessage`) to ask before creating the interface. https://github.com/rescript-lang/rescript-vscode/pull/404/commits/4fc40e372acb8340157ced35bb308afb8e808bf2

client/src/commands/switch_impl_intf.ts
```diff
@@ -0,0 +1,49 @@
...
+    return window.showInformationMessage("Language server not running");
+  }
+
+  const editor = window.activeTextEditor;
+
+  if (!editor) {
+    return window.showInformationMessage("No active editor");
+  }
+
+  const isIntf = editor.document.uri.path.endsWith(".resi");
```
zth, line 17: Should probably check that's it's actually a `.res` file as well here, if it's not `.resi`. And alert the user that this command can only run on `.res` or `.resi` files.

mattdamon108: It makes sense to me! I’ll add some codes to check if it is *.res either!

mattdamon108: https://github.com/rescript-lang/rescript-vscode/pull/404/commits/51b017c401a7355f45823448543c8f4dfea7a2c6

client/src/commands/switch_impl_intf.ts
```diff
@@ -40,7 +40,7 @@ export const switchImplIntf = async (client: LanguageClient) => {
       ...["No", "Yes"]
     );
 
-    if (selection === "No") return;
+    if (selection !== "Yes") return;
```
mattdamon108, line 43: This allows the creation of an interface file only when the user input "Yes".

zth (approved):
This is great, thank you for your contribution @mattdamon108 ! 😄

mattdamon108:
@zth Great! 😃

PR #415, opened 2022-05-12
Complete switch constructor

cristianoc:


analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,9 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
 posCursor:[355:23] posNoWhite:[355:22] Found expr:[355:12->355:23]
 []
 
+Complete tests/src/Completion.res 362:8
+posCursor:[362:8] posNoWhite:[362:7] Found expr:[360:8->365:3]
+posCursor:[362:8] posNoWhite:[362:7] Found expr:[361:2->365:3]
+XXX Pexp_match with 1 cases not handled
```
cristianoc, line 1389: This shows that parser recovery leads to a single case being considered in the switch.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,15 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+[]
+
+Complete tests/src/Completion.res 367:30
+posCursor:[367:30] posNoWhite:[367:29] Found expr:[367:11->367:30]
+posCursor:[367:30] posNoWhite:[367:29] Found expr:[367:16->367:30]
+XXX Pexp_match with 1 cases not handled
```
cristianoc, line 1395: This shows that parser recovery also gives 1 case when no more cases others follow.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,21 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+Complete tests/src/Completion.res 367:30
+posCursor:[367:30] posNoWhite:[367:29] Found expr:[367:11->376:3]
+posCursor:[367:30] posNoWhite:[367:29] Found expr:[367:16->376:3]
+XXX Pexp_match with 1 cases not handled
+[]
+
+Complete tests/src/Completion.res 372:8
+posCursor:[372:8] posNoWhite:[372:7] Found expr:[370:8->376:3]
+posCursor:[372:8] posNoWhite:[372:7] Found expr:[371:2->376:3]
+XXX Pexp_match with 2 cases not handled
```
cristianoc, line 1401: This example confirms that the first 2 cases are squashed into one.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1387,29 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
 posCursor:[355:23] posNoWhite:[355:22] Found expr:[355:12->355:23]
 []
 
+Complete tests/src/Completion.res 362:8
+posCursor:[362:8] posNoWhite:[362:7] Found expr:[360:8->365:3]
+posCursor:[362:8] posNoWhite:[362:7] Found expr:[361:2->365:3]
+XXX Pexp_match with 1 cases not handled
+XXX first case pattern:[362:7->364:5] expression:[364:9->364:10]
+posCursor:[362:8] posNoWhite:[362:7] Found pattern:[362:7->364:5]
+posCursor:[362:8] posNoWhite:[362:7] Found pattern:[362:7->362:8]
```
cristianoc, line 1396: From the location, this could be the pattern `T`.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1387,32 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+XXX Ppat_construct T:[362:7->362:8]
```
cristianoc, line 1397: Indeed here's the `T` pattern.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,19 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+[]
+
+Complete tests/src/Completion.res 373:21
+posCursor:[373:21] posNoWhite:[373:20] Found expr:[371:8->376:3]
+posCursor:[373:21] posNoWhite:[373:20] Found expr:[372:2->376:3]
+posCursor:[373:21] posNoWhite:[373:20] Found pattern:[373:7->375:5]
+posCursor:[373:21] posNoWhite:[373:20] Found pattern:[373:7->373:21]
+XXX Ppat_construct AndThatOther.T:[373:7->373:21]
```
cristianoc, line 1399: Looks like we can catch more complex paths too.

analysis/src/CompletionFrontEnd.ml
```diff
@@ -666,6 +667,24 @@ let completionWithParser ~debug ~path ~posCursor ~currentFile ~text =
...
+          (Pos.toString posCursor) (Pos.toString posNoWhite)
+          (Loc.toString pat.ppat_loc);
+      (match pat.ppat_desc with
+      | Ppat_construct (lid, _) ->
+        let lidPath = flattenLidCheckDot lid in
+        if debug then
+          Printf.printf "XXX Ppat_construct %s:%s\n"
+            (lidPath |> String.concat ".")
+            (Loc.toString lid.loc);
+        setResult (Cpath (CPId (lidPath, Value)))
```
cristianoc, line 684: Triggering completion as a `Value`. There's no need for specific completions for `Constructor` because of the upper case. Lower-case ids are normal values, and upper-case values are constructors.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,51 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+XXX Ppat_construct T:[362:7->362:8]
+Completable: Cpath Value[T]
+[{
+    "label": "That",
+    "kind": 4,
+    "tags": [],
+    "detail": "That\n\ntype v =  This | That",
+    "documentation": null
+  }, {
+    "label": "This",
```
cristianoc, line 1400: `This` and `That` are found as expected.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1381,3 +1383,51 @@ posCursor:[355:23] posNoWhite:[355:22] Found expr:[0:-1->355:23]
...
+    "kind": 4,
+    "tags": [],
+    "detail": "This\n\ntype v =  This | That",
+    "documentation": null
+  }, {
+    "label": "TableclothMap",
```
cristianoc, line 1406: This is a module starting with `T` which happens to be in scope.
This is accurate, as the user might be in the process of writing `ThatSpecificModule.SomeConstructor`.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1030,6 +1030,7 @@ Completable: Cpath Value[SomeLocal]
   }]
 
 Complete tests/src/Completion.res 271:20
+posCursor:[271:20] posNoWhite:[271:19] Found pattern:[271:7->274:3]
```
cristianoc, line 1033: This, and other similar lines indicate that this affects a bit existing test cases, but does not lead to a different outcome.
It just means that the relevant type, found below, happens to be inside a pattern containing the cursor.

analysis/tests/src/expected/Completion.res.txt
TheSpyder: I disagree with this, except in the rare case where the type of the switch variable is not known (because it might be used to disambiguate, as you suggested). That should only happen once, or in many cases never.

My understanding is that once the type is known the compiler no longer needs disambiguation to find constructors. At that point seeing all scope modules in the autocomplete list is just noise.

https://rescript-lang.org/try?code=LYewJgrgNgpgBAeQC4AsYCc4F44G8BQccSAngA7wBu2iAdvAD5wAqA7iHE8yujDIXFhI4AMxqUAhlAjwsAPgFEAzqwCWSAMYo4k6fAJEiAeiNxaHMKqUTgAI1UBzCBKSqQtRZzqy5cAIyeXOzYvgBMgSw8fCFwAMyeAL74SfhCouJSMiECKupaOpn6AiZwSqq0DrBwltZ2js6u7gJMyGjoAHQI9DEBREEc8nDhfZG8PnECSUA

I'm about to start looking at how to remove modules from the autocomplete list in this scenario, but now I realise it was intentional... can we discuss whether removing them is a good idea?

cristianoc: I think you're interested in type contexts, here: https://github.com/rescript-lang/rescript-vscode/issues/494
Not this PR.

TheSpyder: I am, but then I found this PR and it looked like completion in switch statements already included all the valid constructors.

cristianoc: This is an ongoing process. Not everything is at the level it could be.
In the process, some techniques have been used to aid autocompletion where the alternative would be to get nothing.

Issue #328, opened 2021-12-12
Please allow to restart the server


SofianeDjellouli:
When the server crashes, I have to reload the window to restart it. Can it be possible to restart it from the command palet, like the TS server for instance?

jmagaram:
Please see...

https://forum.rescript-lang.org/t/vs-code-extension-stops-working-anyone-else/2397

The extension seems SO VERY close to being really great. But it fails regularly - maybe once a day - for unknown reasons on my little project. It is really hard to get it going again. Some magic of build with watcher, clean, "kill" process, restart VS Code, etc. will eventually get it going again - don't necessarily need any changes to my code - but I can't figure out a reliable set of steps that will quickly get me up and running. The best option is to make the extension never fail. Second best would be some built-in diagnostics or something in the output window so I could tell you why it is failing and you could fix it. But until that day I'd really like an easy way to get it working again. A command palette command for "Restart ReScript Extension" or something similar would be useful. It would do the clean, fix stale files, whatever. Overall I just want to see some more reliability.

Lomand:
Those random restarts are due to
```
>> Fatal error: Longident.flat
Fatal error: exception Misc.Fatal_error
```
This issue should go away once https://github.com/rescript-lang/syntax/pull/449 is merged.

srikanthkyatham:
Hi @zth 

I would be happy help in this issue. Let me know how and where to proceed.

zth:
@srikanthkyatham excellent! Some quick background:

The language server is created here: 
https://github.com/rescript-lang/rescript-vscode/blob/master/client/src/extension.ts#L102-L107
...and then started here: 
https://github.com/rescript-lang/rescript-vscode/blob/master/client/src/extension.ts#L217-L218

Notice that the result of `start` is pushed to `subscriptions` - that's important so VSCode can clean up appropriately as the extension is deactivated/activated.

Essentially, you'll need to:
1. Add and expose a new command that restarts the client. There are many examples of commands in the repo you can look at.
2. Ensure that the old client is properly shut down and cleaned up when restarting. A tip is to look for any dangling processes (`ps aux` or similar) that would indicate that the server is still running.

Feel free to open a WIP PR early if you'd like!

cristianoc:
There's also some global state that presumably needs to be cleared. E.g. `stupidFileContentCache` and `projectsFiles` and `codeActionsFromDiagnostics`.

zth:
[...]

PR #423, opened 2022-05-14
feat: add restart language  server command

a-c-sreedhar-reddy:
Implements #328

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -194,7 +194,7 @@ export function activate(context: ExtensionContext) {
     customCommands.switchImplIntf(client);
   });
 
-  commands.registerCommand("rescript-vscode.restartLanguageServer", () => {
+  commands.registerCommand("rescript-vscode.restart_language_server", () => {
     client.stop().then(() => {
       client = new LanguageClient(
         "ReScriptLSP",
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -201,30 +201,34 @@ export function activate(context: ExtensionContext) {
   commands.registerCommand("rescript-vscode.restart_language_server", () => {
     client.stop().then(() => {
       client = createLanguageClient();
-      client.start();
+      attachDeadCodeAnalysis(client);
+      context.subscriptions.push(client.start());
     });
   });
 
-  // This sets up a listener that, if we're in dead code analysis mode, triggers
-  // dead code analysis as the LS server reports that ReScript compilation has
-  // finished. This is needed because dead code analysis must wait until
-  // compilation has finished, and the most reliable source for that is the LS
-  // server, that already keeps track of when the compiler finishes in order to
-  // other provide fresh diagnostics.
-  client.onReady().then(() => {
-    context.subscriptions.push(
-      client.onNotification("rescript/compilationFinished", () => {
-        if (inDeadCodeAnalysisState.active === true) {
-          customCommands.deadCodeAnalysisWithReanalyze(
-            inDeadCodeAnalysisState.activatedFromDirectory,
-            diagnosticsCollection,
-            diagnosticsResultCodeActions
-          );
-        }
-      })
-    );
-  });
+  function attachDeadCodeAnalysis(client: LanguageClient) {
+    // This sets up a listener that, if we're in dead code analysis mode, triggers
+    // dead code analysis as the LS server reports that ReScript compilation has
+    // finished. This is needed because dead code analysis must wait until
+    // compilation has finished, and the most reliable source for that is the LS
+    // server, that already keeps track of when the compiler finishes in order to
+    // other provide fresh diagnostics.
+    client.onReady().then(() => {
+      context.subscriptions.push(
+        client.onNotification("rescript/compilationFinished", () => {
+          if (inDeadCodeAnalysisState.active === true) {
+            customCommands.deadCodeAnalysisWithReanalyze(
+              inDeadCodeAnalysisState.activatedFromDirectory,
+              diagnosticsCollection,
+              diagnosticsResultCodeActions
+            );
+          }
+        })
+      );
+    });
+  }
 
+  attachDeadCodeAnalysis(client);
   // Start the client. This will also launch the server
   context.subscriptions.push(client.start());
 }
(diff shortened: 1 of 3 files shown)


a-c-sreedhar-reddy:
cc @zth @cristianoc

cristianoc:
Thank you @a-c-sreedhar-reddy. Loving the tiny size of the change!

@zth would you like to also take a look?

client/src/extension.ts
```diff
@@ -194,6 +194,18 @@ export function activate(context: ExtensionContext) {
...
   });
 
+  commands.registerCommand("rescript-vscode.restartLanguageServer", () => {
+    client.stop().then(() => {
+      client = new LanguageClient(
+        "ReScriptLSP",
+        "ReScript Language Server",
+        serverOptions,
+        clientOptions
+      );
```
zth, line 204: What do you think about making this a function `createLanguageClient()` or similar, and re-use at the initial creation of the language client too? Just so they don't accidentally diverge.

zth (commented):
Looking great! Just had one mini question.

package.json
```diff
@@ -62,6 +62,10 @@
 				"command": "rescript-vscode.stop_dead_code_analysis",
 				"title": "ReScript: Stop dead code analysis."
 			},
+			{
+				"command": "rescript-vscode.restartLanguageServer",
```
fhammerschmidt, line 66: the other commands use `snake_case`, so please let's stay consistent here and use `restart_language_server`.

a-c-sreedhar-reddy: Thanks for the review @fhammerschmidt  . I have updated it to snake_case.

client/src/extension.ts
a-c-sreedhar-reddy: @zth 
can the `createLanguageClient` function accept `serverOptions,clientOptions` parameters and return the client?

I have also noticed that there is a notification handler attached to the client `rescript/compilationFinished`. 

Would it be a good idea to do this in the `createLanguageClient` function by accepting the notification handler as a callback?
or
Is it better to do this after `createLanguageClient`?

fhammerschmidt: Not sure about @zth's intention, but I would just put everything into a closure and reuse it.

```js
  let createLanguageClient = () => {
    return new LanguageClient(
      "ReScriptLSP",
      "ReScript Language Server",
      serverOptions,
      clientOptions
    );
  }
```

and then 

```js
client = createLanguageClient();
```

at least as long as you do not expect that some of the options need to be changed.

a-c-sreedhar-reddy: Hi @zth I have created the `createLanguageClient` and used it.

zth: @a-c-sreedhar-reddy sorry, missed this! The last piece missing, I think, is that you'll need to re-instantiate the listener for the "compilation completed" event. You can test whether this is needed by using the restart command, and then after the command has restarted the client:
1. Start Dead code analysis mode
2. Make some change that fixes (or introduces) a new error that affects the dead code analysis (add a single unused let binding for instance)
3. Make sure that error is picked up and reported in the editor

The dead code analysis mode relies on the compilation completed event, so we need to make sure that's still subscribed to, even after restarts.

a-c-sreedhar-reddy: @zth The dead code analysis was not getting triggered. as we stopped the old client and created a new one.

Made the changes to run deadcode analysis on `compilationFinished` event.

zth (approved):
Great, thank you for your contribution! 🎉

client/src/extension.ts
```diff
@@ -194,26 +198,37 @@ export function activate(context: ExtensionContext) {
...
-    );
+  commands.registerCommand("rescript-vscode.restart_language_server", () => {
+    client.stop().then(() => {
+      client = createLanguageClient();
+      attachDeadCodeAnalysis(client);
+      context.subscriptions.push(client.start());
+    });
   });
 
+  function attachDeadCodeAnalysis(client: LanguageClient) {
```
cristianoc, line 209: How about moving this function up so it's not in the middle of the logic.

client/src/extension.ts
```diff
@@ -194,26 +198,37 @@ export function activate(context: ExtensionContext) {
...
-            diagnosticsCollection,
-            diagnosticsResultCodeActions
-          );
-        }
-      })
-    );
+  commands.registerCommand("rescript-vscode.restart_language_server", () => {
+    client.stop().then(() => {
+      client = createLanguageClient();
+      attachDeadCodeAnalysis(client);
```
cristianoc, line 204: This is done both here and below -- how about moving inside `createLanguageClient` so there's no chance to forget it.

client/src/extension.ts
```diff
@@ -194,26 +198,37 @@ export function activate(context: ExtensionContext) {
...
+      context.subscriptions.push(client.start());
```
cristianoc, line 205: This is also done both here and below, should this also be moved inside `createLanguageClient`? Or is there the question of doing some more things before `client.start()`.

client/src/extension.ts
cristianoc: Also unsure why this:
```ts
commands.registerCommand("rescript-vscode.open_compiled", () => {
    customCommands.openCompiled(client);
  });
```
does not need to be done again after a restart.

And if it does not, how do we know it will be always the case. E.g. why it's not going to require the same treatment as `attachDeadCodeAnalysis` in future if it changes, and one might overlook to make the change here.

In general, is it possible to put all the logic in one place, and execute exactly the same logic at the beginning and after a restart?

cristianoc (approved):
Some comments and a bunch of questions.

client/src/extension.ts
a-c-sreedhar-reddy: the above command works 
because the handler uses `client` which will always be the latest value because of closure.

client/src/extension.ts
a-c-sreedhar-reddy: I think we can start the client in the `createLanguageClient` function.

client/src/extension.ts
a-c-sreedhar-reddy: @cristianoc moved the function to the top.

client/src/extension.ts
a-c-sreedhar-reddy: @cristianoc @zth I have made the change to attach deadcodeanalysis in the createLanguageServer function.

a-c-sreedhar-reddy:
Thanks @zth @cristianoc @fhammerschmidt

PR #428, opened 2022-05-22
Use a different rescript-editor-analysis.exe when on ARM

ellbur:


cristianoc:
One comment is that this change affects M1 macs, and would need to not change the path in that case. As the installation does not have a `darwin-arm64` folder.

Even for linux arm, there's no automatic installation that creates `linux-arm64` so I'm not sure where that is created.
If you create it yourself manually, why not put the executable simply in `linux`?

server/src/constants.ts
```diff
@@ -1,5 +1,8 @@
 import * as path from "path";
 
+let archString = process.arch == "arm64" ? "-arm64" : "";
+let platformArchString = process.platform + archString;
```
cristianoc, line 4: Breaks on Mac m1.

cristianoc (changes requested):

server/src/constants.ts
cristianoc: Should be gated to only apply on linux.

ellbur:
> Even for linux arm, there's no automatic installation that creates `linux-arm64` so I'm not sure where that is created. If you create it yourself manually, why not put the executable simply in `linux`?

Yes, I am creating it myself manually. The issue is that [vim-rescript](https://github.com/rescript-lang/vim-rescript) includes executables in its package, so it needs a way to include a separate executable for ARM.

I will fix it so it only applies to Linux, and test whether code reformatting still works.

ellbur:
I tested that code formatting does still work (using `:RescriptFormat` in vim-rescript).

cristianoc:
https://github.com/rescript-lang/rescript-vscode/pull/467

Issue #23, opened 2020-11-24
Prevent the "Start a build for this project to get the freshest data?" notification from spamming


banacorn:
It pops up every time when a file is opened.
I don't even know what this notification means.

There should be an option for disabling this notification (in fact, every notification)
This is just annoying.

banacorn:
It seems like the notification would appear only when the `.bsb.lock` is missing.

https://github.com/rescript-lang/rescript-vscode/blob/9bbacb44196726549f36df2b3a205b0a701ff50f/server/src/server.ts#L144-L147

The notification becomes spammy if `.bsb.lock` is too short-lived. 
Sometimes my `.bsb.lock` would keep getting deleted, but I don't know who did it.

alexfedoseev:
IIRC `.bsb.lock` gets created/delete by compiler. As long as some `bsb` process is watching, `.bsb.lock` exists.

> Sometimes my .bsb.lock would keep getting deleted, but I don't know who did it.

It might happen when you rebuild project often.

chenglou:
It should only pop up for the first file of a project you open? Is that not the case for you? Got a video?

banacorn:
Unfortunately I can't seem to reproduce it.

It happened after I converted everything from ReasonML to ReScript.
The problem only occurred in the first few hours, but I'm not sure what I did to stop it from spamming.

@arafatamin, @thezjy, @mickeyvip, @BlueHotDog, @darenju, @kgtkr, @FilipKittnar.
Hey sorry for tagging you, could you share your experiences and observations about this problem?

darenju:
Now that you mention it, I think I had the same problem than you: converting from ReasonML to ReScript.

I have not tried to make a new project using ReScript directly (partly because of the incomplete syntax highlighting), so I cannot say if the problem goes away after some time, or just after the first file was open.

FilipKittnar:
> It should only pop up for the first file of a project you open? Is that not the case for you? Got a video?

@chenglou , this happens very inconsistently but this morning it actually happened so I took a video. It can be seen how the .bsb.lock keeps appearing and disappearing when switching between .res files. But as I said, this happens inconsistently. Yesterday it was fine.
https://u.pcloud.link/publink/show?code=XZ9oY4XZA5FnABEd1qf7DraQFzJsbYaIlday

banacorn:
I'm suspecting that this happens when there's only **ONE** `.res` file opened at a time.

banacorn:
[...]

PR #432, opened 2022-05-28
sharing config between client and server

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/23

This introduces configuration in the extension, periodically synced to the language server for use. We'll probably need to revisit this if we get more intricate configuration needs later on, but for now this is simple enough and will do for the cases we currently have.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -11,6 +11,7 @@ import {
   DidChangeTextDocumentNotification,
   DidCloseTextDocumentNotification,
   DidChangeConfigurationNotification,
+  InitializeParams,
 } from "vscode-languageserver-protocol";
 import * as utils from "./utils";
 import * as codeActions from "./codeActions";
@@ -280,8 +281,6 @@ if (process.argv.includes("--stdio")) {
   send = (msg: m.Message) => process.send!(msg);
   process.on("message", onMessage);
 }
-
-askForAllCurrentConfiguration();
 
 function hover(msg: p.RequestMessage) {
   let params = msg.params as p.HoverParams;
@@ -860,8 +859,14 @@ function onMessage(msg: m.Message) {
         askForAllCurrentConfiguration();
       }, 10_000);
 
-      // Pull config right away as we've initied.
-      askForAllCurrentConfiguration();
+      // Save initial configuration, if present
+      let initParams = msg.params as InitializeParams;
+      let initialConfiguration = initParams.initializationOptions
+        ?.extensionConfiguration as extensionConfiguration | undefined;
+
+      if (initialConfiguration != null) {
+        extensionConfiguration = initialConfiguration;
+      }
 
       send(response);
     } else if (msg.method === "initialized") {
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -28,7 +28,7 @@ interface extensionConfiguration {
 let extensionConfiguration: extensionConfiguration = {
   askToStartBuild: true,
 };
-let pullConfigurationInterval: NodeJS.Timeout | null = null;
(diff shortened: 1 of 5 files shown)


server/src/server.ts
```diff
@@ -20,6 +22,14 @@ import { fileURLToPath } from "url";
 import { ChildProcess } from "child_process";
 import { WorkspaceEdit } from "vscode-languageserver";
 
+interface extensionConfiguration {
```
cristianoc, line 25: curious: how does one decide whether this should be an interface or a type?

server/src/server.ts
```diff
@@ -763,6 +793,9 @@ function onMessage(msg: m.Message) {
     } else if (msg.method === DidCloseTextDocumentNotification.method) {
       let params = msg.params as p.DidCloseTextDocumentParams;
       closedFile(params.textDocument.uri);
+    } else if (msg.method === DidChangeConfigurationNotification.type.method) {
```
cristianoc, line 796: should this be `.type.method` or `.method` like in the other cases?

server/src/server.ts
```diff
@@ -820,6 +853,21 @@ function onMessage(msg: m.Message) {
         result: result,
       };
       initialized = true;
+
+      // Periodically pull configuration from the client.
+      pullConfigurationInterval = setInterval(() => {
```
cristianoc, line 858: perhaps `pullConfigurationAtInterval` or `pullConfigurationPeriodically` ?

server/src/server.ts
```diff
@@ -820,6 +853,21 @@ function onMessage(msg: m.Message) {
...
+        askForAllCurrentConfiguration();
+      }, 10_000);
```
cristianoc, line 860: this number is what I'd expect to be called `pullConfigurationInterval`

cristianoc (approved):
How about adding the config to the `example-project` as an example.
Also, should the README be updated.

server/src/server.ts
zth: It's pretty arbitrary, but they at least partly have different use cases. My understanding is interfaces have better perf and are recommended in general for object shapes.

server/src/server.ts
zth: According to the types, this notification does not have `.method`, only `.type.method`.

server/src/server.ts
zth: Sure!

zth:
Updated the readme. Can't set up the configuration in the example project though, because VSCode doesn't save settings in the repo by default.

cristianoc:
Thanks! Merging.

cristianoc:
Nit: does not look like changing the setting and restarting the server picks them up.
Reloading the window does.
At least in debug mode. Haven't tried if it's the same when installing the .vsix file

zth:
You mean when restarting via the restart command recently added?

cristianoc:
> You mean when restarting via the restart command recently added?

yes

cristianoc:
Or I guess the more general question is: why polling. If one needs to restart anyway.
Maybe having to restart is not bad, and one can make everything simpler and remove the polling logic.

zth:
Hmm, I don't need to restart anything, polling works fine for me as I tested. How are you testing it?

cristianoc:
What I've tried is: change the setting, then ask the server to restart.

cristianoc:
Don't know what else to test without restarting the entire editor.

zth:
Ahh, I think I know the issue. I'll fix it.

zth:
Done here: https://github.com/rescript-lang/rescript-vscode/pull/438

PR #438, opened 2022-05-30
ensure updated config is synced on use of restart command

zth:
Initial read configuration was never refreshed as the restart command runs, which means you'd always get potentially stale initial config after a restart.

Also clean up a few init things that I suspect has never been cleaned up after the initial adaption of the LS client example.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -69,7 +69,44 @@ let client: LanguageClient;
 // });
 
 export function activate(context: ExtensionContext) {
-  function attachCodeAnalysis(client: LanguageClient) {
+  function createLanguageClient() {
+    // The server is implemented in node
+    let serverModule = context.asAbsolutePath(
+      path.join("server", "out", "server.js")
+    );
+    // The debug options for the server
+    // --inspect=6009: runs the server in Node's Inspector mode so VS Code can attach to the server for debugging
+    let debugOptions = { execArgv: ["--nolazy", "--inspect=6009"] };
+
+    // If the extension is launched in debug mode then the debug server options are used
+    // Otherwise the run options are used
+    let serverOptions: ServerOptions = {
+      run: { module: serverModule, transport: TransportKind.ipc },
+      debug: {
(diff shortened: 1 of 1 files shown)


client/src/extension.ts
```diff
@@ -122,21 +122,20 @@ export function activate(context: ExtensionContext) {
     },
   };
 
-  // Options to control the language client
-  let clientOptions: LanguageClientOptions = {
```
cristianoc, line 126: Is there a way to make sure `let clientOptions` is run instead? Why are we not doing exactly the same steps as the first start, on a restart?
Are there any steps that should *not* be repeated?

cristianoc: E.g. at the coarsest level why not deactivate the client and activate it again.

zth: I don't quite follow what you mean. `let clientOptions` is run...?

cristianoc: > I don't quite follow what you mean. `let clientOptions` is run...?

The "function containing it" is run.

cristianoc: Or else, we need to ask the same question of any other `let foo` in that file, which upon future changes, might not do what is's expected to do after a restart.

cristianoc: Also a curiosity: who calls `deactivate` and when?

zth: @cristianoc done.

cristianoc (approved):
Looks great.

The only thing that comes to mind, is how to make it completely explicit that the code that follows `createLanguageClient` is the global state. (which is preserved on restart).

cristianoc:
> Looks great.
> 
> The only thing that comes to mind, is how to make it completely explicit that the code that follows `createLanguageClient` is the global state. (which is preserved on restart).

In fact, there's one other small bit of global state:
```
let client: LanguageClient;
```

which is followed by a bunch of commented out code. Then the `createLanguageClient` function. Then the rest of the global state.

PR #440, opened 2022-05-31
arity0 hovering issues

cristianoc:


cristianoc:
See https://github.com/rescript-lang/syntax/pull/512

PR #444, opened 2022-05-31
Test latest compiler on master.

cristianoc:
Testing the following:
- Take the compiler from master.
- Use the syntax submodule at the time of the latest compiler release.
- Run editor integration tests.


cristianoc:
CI is expected to fail, the test was on my machine.

PR #453, opened 2022-06-11
Feature: Support Inlay Hint

aspeddro:
I used `DocumentSymbol.ml` to try to understand how the parser works. There's a lot of room for improvement.

I'll list some that I need help with:

- Update type hints when compilation is finished
- Find a better way to get the return type. I split the string at `=>` and got the last item in the list.
- A better way to find the starting position of the variable.
- Get type hints for a range, currently gets the entire file.
- Type Hint for tuples destructuring

  ```res
  let tuple = ("ReScript", "lol")
  
  let (lang, _) = tuple
  ```

  Output:
  ```text
  let (lang: string, _) = tuple
  ```

  - Ignore type hint for `_`?

- Provide a setting to disable inlayHint
- Add type hint for pipe
- Type hint support in parameters

cc @zth


cristianoc:
This is probably a good feature to make configurable with an opt-in. @zth what do you think?
Side comment, prob better to move the code cleanup/rename part to a separate PR later on, so it's not mixed up with the new functionality.

cristianoc:
Unless it's clearly separated in the commits already.

aspeddro:
Inlay Hint requires an updated version of `vscode-languageserver-node`. It is part of the latest 3.17 specification.

https://github.com/microsoft/vscode-languageserver-node#3170-protocol-800-json-rpc-800-client-and-800-server

zth:
So, adding to @aspeddro and @cristianoc , I think we should do this:
- Change this PR so it only involves implementing inlay hints. Renaming and other refactoring (and even formatting of things that wasn't previously formatted, but has no other changes than just formatting) should go in a separate PR so this stays as on point as possible.
- Ensure there's an option for toggling inlay hints. I can fix that ✋ 
- Ensure the inlay hints are refreshed at compile finish. I can fix that too ✋ 
- Figure out the missing cases, like you mentioned @aspeddro. We'll need the expertise of @cristianoc here probably. But we'll get to that after we've done the other things.

I'll go ahead and fix the points I mention above asap. @aspeddro can you remove formatting, refactoring etc from the PR meanwhile?

Great work so far btw, this is really exciting and will be a great feature!

PS. I wonder if we could just reuse the full completion logic for finding the identifiers to give inlay hints? I have a feeling that will map quite well. It even works on unsaved sources, and we can easily filter out anything it doesn't find a type for yet. Thoughts @cristianoc ?

aspeddro:
> Change this PR so it only involves implementing inlay hints. Renaming and other refactoring (and even formatting of things that wasn't previously formatted, but has no other changes than just formatting) should go in a separate PR so this stays as on point as possible.

Done

zth:
@aspeddro quick update - just like you said/did, we'll need to upgrade the VSCode client version for this to work out. I will do that in a separate PR that we can then rebase this on once merged. That'll make the diff a bit cleaner. You'll hear back from me in a little while.

aspeddro:
Updates, I made some advances in the analysis bin:

- Type hints in parameters. 
- Tuple support

image

#### TODO:

- Show only type of return for function
- Add type hint for pipe call

aspeddro:
About the return type, is it better to implement in `syntax` repo or in analysis?

It is similar to [res_outcome_printer.ml](https://github.com/rescript-lang/syntax/blob/master/src/res_outcome_printer.ml#L162).

zth:
Curious, where would you show only the return type?

aspeddro:
> Curious, where would you show only the return type?

After `)`

```res
let add = (a: int, b: int): int => a + b
```

zth:
> > Curious, where would you show only the return type?
> 
> After `)`
> 
> ```rescript
> let add = (a: int, b: int): int => a + b
> ```
> 
> If the function has type annotation, then no type hint will be added.

Ah of course!

Cc @cristianoc for the other questions, I don't have much input there unfortunately.

server/src/server.ts
```diff
@@ -23,9 +23,13 @@ import { WorkspaceEdit } from "vscode-languageserver";
 
 interface extensionConfiguration {
   askToStartBuild: boolean;
+  autoRunCodeAnalysis: boolean;
+  inlayHints: boolean;
 }
 let extensionConfiguration: extensionConfiguration = {
   askToStartBuild: true,
+  autoRunCodeAnalysis: false,
```
cristianoc, line 31: What does this change imply?
This config is pre-existing and seemed to work. How is this related to the existing way of reading this configuration? Do we have 2 ways now? And when do they fire? And which one wins? And is one of them redundant?

server/src/server.ts
```diff
@@ -849,6 +877,15 @@ function onMessage(msg: p.Message) {
       };
       send(response);
     } else if (msg.method === "initialize") {
+      // Save initial configuration, if present
+      let initParams = msg.params as InitializeParams;
+      let initialConfiguration = initParams.initializationOptions
+        ?.extensionConfiguration as extensionConfiguration | undefined;
+
+      if (initialConfiguration != null) {
+        extensionConfiguration = initialConfiguration;
```
cristianoc, line 886: What happens without this change? What behaviour is affected?
Asking as this is not just about the new configuration option just introduced.

cristianoc (commented):
Commented on aspects where it's not completely obvious that behaviour of the existing features is not affected.

Asking lots of questions as innocent looking changes have broken the extension several times in the past.

zth:
I'd like to keep this moving if we're aiming to get this merged anytime soon (which I'd personally like to). @aspeddro what's your take on the status of this PR, are there work left to be done, or can we start a more thorough review? Do you have time to continue working on this in the near future?

cristianoc:
I guess next step is to rebase on master, so that some code movement disappears and one can review only what's new.

aspeddro:
> Do you have time to continue working on this in the near future?

yup

> About the return type, is it better to implement in `syntax` repo or in analysis?
> 
> It is similar to [res_outcome_printer.ml](https://github.com/rescript-lang/syntax/blob/master/src/res_outcome_printer.ml#L162).

There is still work, for example getting the return type ☝️

zth:
> yup

Great! Would you mind rebasing on master?

> About the return type, is it better to implement in syntax repo or in analysis?
> It is similar to [res_outcome_printer.ml](https://github.com/rescript-lang/syntax/blob/master/src/res_outcome_printer.ml#L162).

@cristianoc do you have any input on best practices for getting the return type of a function, and where to put that logic?

cristianoc:
> > yup
> 
> Great! Would you mind rebasing on master?
> 
> > About the return type, is it better to implement in syntax repo or in analysis?
> > It is similar to [res_outcome_printer.ml](https://github.com/rescript-lang/syntax/blob/master/src/res_outcome_printer.ml#L162).
> 
> @cristianoc do you have any input on best practices for getting the return type of a function, and where to put that logic?

If you have the full type, the return is just from following the arrow type. It really depends on what "return type" means. As the concept per se does not exist in a curried language.

zth:
@aspeddro suggestion - let's ignore the return type of functions for now. There's a feature request for handling the case of showing what type a function has via a code lens instead (https://forum.rescript-lang.org/t/feature-request-is-it-possible-to-add-codelens-decoration-on-the-vscode-extension/3565/3). I personally think that's better, since that'll show above the let binding itself, where there's more room. 

What's the status of parameter names and pipe calls?

aspeddro:
> What's the status of parameter names and pipe calls?

I made little progress on this feature. I will explore these features in the near future.

In this PR I will only add the basic support. Ready for review.

aspeddro:
I don't know why the tests are failing.

analysis/src/Hint.ml
```diff
@@ -0,0 +1,118 @@
...
+      | Const_char _ -> "char"
+      | Const_string _ -> "string"
+      | Const_float _ -> "float"
+      | Const_int32 _ -> "int32"
+      | Const_int64 _ -> "int64"
+      | Const_nativeint _ -> "int")
+  | Typed (_, t, locKind) ->
+    let fromType typ =
+      typ |> Shared.typeToString
+      |> Str.global_replace (Str.regexp "[\r\n\t]") ""
```
zth, line 23: Why are we doing this...? Is it to "pluck" off what you want from `typeToString`?

zth (commented):
A few minor questions from me!

analysis/src/Hint.ml
aspeddro: `typeToString` returns with newlines. 

Example:

```res
let ppp = (Ext.make, Ext.makeProps)
```

`Ext.make` typeToSring return

```
React.componentLike<
  {"align": option<string>},
  React.element,
>
```

image

image
**Forgot to remove empty spaces.**

analysis/src/Protocol.ml
```diff
@@ -1,6 +1,7 @@
-type position = {line: int; character: int}
-type range = {start: position; end_: position}
-type markupContent = {kind: string; value: string}
+type position = {line : int; character : int}
+type range = {start : position; end_ : position}
+type markupContent = {kind : string; value : string}
+type inlayHint = {position : position; label : string; kind : int; paddingLeft: bool; paddingRight: bool}
```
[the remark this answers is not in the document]
aspeddro: Fixed https://github.com/rescript-lang/rescript-vscode/pull/453/commits/674e8be77db6505c33fc32e412d54b2f02a15cd6

analysis/src/Hint.ml
aspeddro: Trim whitespace fixed by https://github.com/rescript-lang/rescript-vscode/pull/453/commits/a4353efd190c4a2ef48273611b38816bfc6206fd

zth: Ahh right, and the newline character appears in the output if we don't remove them I guess?

I guess the cleanest option would be to have a mode which can print without line breaks. But that might not be feasible/might be to over engineer it. What do you think @cristianoc ?

zth: I tested a bit. I think the whitespace is better to keep. I wonder if we should do something about the potential length of the hint, as in hide it if it's too long. What do other extensions with the same capabilities do @aspeddro , do you know that?

Great work in general, this is a really neat feature that'll be great to have for us!

aspeddro: rust analyzer has a config `rust-analyzer.inlayHints.maxLength`. Default is 25; https://github.com/rust-lang/rust-analyzer/pull/8177

Great suggestion I will add.

aspeddro: Done b8d1ece20839119463eb69e349865514e080ae82

aspeddro: Keep whitespace fixed https://github.com/rescript-lang/rescript-vscode/pull/453/commits/bd4cfdd19d329e6179f8365114a9f7985af0a8e3

analysis/src/Commands.ml
```diff
@@ -28,6 +28,12 @@ let completion ~debug ~path ~pos ~currentFile =
     |> List.map Protocol.stringifyCompletionItem
     |> Protocol.array)
 
+let inlayhint ~path ~maxLength ~debug =
+  let result = match Hint.inlay ~path ~maxLength ~debug with
+  | [] -> Protocol.null
```
cristianoc, line 33: nit: why separate these 2 cases?

client/src/extension.ts
```diff
@@ -235,6 +235,17 @@ export function activate(context: ExtensionContext) {
   // Start the client. This will also launch the server
   client.start();
 
+  // Restart the language client automatically certain configuration changes.
```
cristianoc, line 238: check the text

client/src/extension.ts
```diff
@@ -235,6 +235,17 @@ export function activate(context: ExtensionContext) {
   // Start the client. This will also launch the server
   client.start();
 
+  // Restart the language client automatically certain configuration changes.
+  // These are typically settings that affect the capabilities of the language
+  // client, and because of that requires a full restart.
+  context.subscriptions.push(
+    workspace.onDidChangeConfiguration(({ affectsConfiguration }) => {
+      if (affectsConfiguration("rescript.settings.inlayHints")) {
```
cristianoc, line 243: Is there a special reason to restart only when this specific setting changes?
I imagine as a user I would be quite confused "I think last time I could just change setting, this time I need to restart, strange...".

cristianoc (commented):
The code looks fine.
Just a few comments/suggestions.

cristianoc:
The latest test results were not checked in. Done that. Could you check that this is still the expected result?

cristianoc:
The changelog and readme need to be updated.

cristianoc:
Should this do anything with interface files?

cristianoc (changes requested):
This was left behind w.r.t. master.
Needs rebasing.

cristianoc:
Here's a large file: `analysis/examples/larger-project/src/res_core.res`
Could you check how it behaves there? How does it feel?

cristianoc:
> Here's a large file: `analysis/examples/larger-project/src/res_core.res`
> Could you check how it behaves there? How does it feel?

Tried that, and turning on inlay hints make the extension unusable.
Something needs to be done about that. Putting it behind config is not sufficient.

analysis/src/Hint.ml
zth: Nice!

client/src/extension.ts
zth: It's my text, I'll fix it

client/src/extension.ts
zth: It's only restarted automatically when required. Some configs can be changed on the fly, but other configs require restarting the server because they affect how the lsp is instantiated. If we don't have this then the user would need to be promoted to restart the server manually to have their config take effect.

zth:
> > Here's a large file: `analysis/examples/larger-project/src/res_core.res`
> > Could you check how it behaves there? How does it feel?
> 
> Tried that, and turning on inlay hints make the extension unusable. Something needs to be done about that. Putting it behind config is not sufficient.

Does the extension lock up, or what happens? Can we trace whether it's due to the amount of data being sent from the analysis, or if the analysis is slow, or what it might be?

Ideas: find a reasonable max amount of hints to send and simply cut off if more are found? Are there other modes, like range based inline hints rather than full document hints, that we could support and that might help alleviate?

zth:
There seems to be a range parameter sent with the inlay hint request that we're currently not taking into account: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintParams

Let's try and implement that and see if it helps. I would assume VSCode leverage that to reduce the amount of hints asked for att once in large files. 

I wonder how to handle the automatic refreshing on compile though since I don't think we have that information there. Let's think about that for a bit.

aspeddro:
> Should this do anything with interface files?

Type hint requests are now for `.res` files only.

aspeddro:
> There seems to be a range parameter sent with the inlay hint request that we're currently not taking into account: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintParams
> 
> Let's try and implement that and see if it helps. I would assume VSCode leverage that to reduce the amount of hints asked for att once in large files.
> 
> ~I wonder how to handle the automatic refreshing on compile though since I don't think we have that information there. Let's think about that for a bit.~ scratch that, we don't need that information for a refresh, we're just telling the client things might need refreshing, and it'll issue the requests for hints that it needs to refresh

Added support for range. `analysis/examples/large-project/src/res_core.res` is no longer locking the extension

zth:
Great work @aspeddro , and thank you for the quick turn around! I'll have another look asap.

aspeddro:
> The latest test results were not checked in. Done that. Could you check that this is still the expected result?

Testes Fixed.

client/src/extension.ts
cristianoc: The question is why not the other changes too.
How do I know other configs don't need restarting the server?
And if they don't need restating the server today, how do I know they don't need to restart the server tomorrow.
And if tomorrow they will need a restart, what are the chances the person doing the change will think of that, realise this piece of code exists, and come here to update it?

Overall, why special case thins config, is the question.

cristianoc:
Have you tried with `res_core.res`? What was your experience?

aspeddro:
> Have you tried with `res_core.res`? What was your experience?

Yup. When scrolling sharply to any part of the file, eg top (line 1) to middle (line 3k), it takes about 3 seconds to show the type hints. But if you scroll smoothly the result is responsive.

cristianoc:
@zth how's your experience on that file?

cristianoc:
> > Have you tried with `res_core.res`? What was your experience?
> 
> Yup. When scrolling sharply to any part of the file, eg top (line 1) to middle (line 3k), it takes about 3 seconds to show the type hints. But if you scroll smoothly the result is responsive.

Has something changed?  I have the same experience now, instead of just "loading..." on any over on anything.

cristianoc (approved):

cristianoc:
I'd like to know what happens when one scrolls slowly.
I suspect the 3 seconds delay is still there, but you don't see it as the work happens on a bunch of not-yet-visible code.
So there's a lot of processing happening all the time.

analysis/src/Hint.ml
```diff
@@ -32,15 +32,21 @@ let locItemToTypeHint ~full:{file; package} locItem =
         | `Field -> fromType t))
   | _ -> None
 
-let inlay ~path ~maxLength ~debug =
-  let maxlen = try Some(int_of_string maxLength) with Failure _ -> None in
+let inlay ~path ~pos ~maxLength ~debug =
```
cristianoc, line 35: Do we still need `maxLength` now that ranges are considered?
I though length is implicit in the loc that vscode sends?

cristianoc:
Did some tests, and turning on this feature, on slow scroll, on my m1 Mac, pushes code renderer slightly above WindowServer, in terms of cpu utilisation. While it was slightly below beforehand.
So there's an observable difference, but not a huge one.

cristianoc:
In terms of memory, vscode stays under every one of the 4 instances of `ocamllsp` I have on my machine at the moment (no idea why multiple instances).

analysis/src/Hint.ml
aspeddro: yup. `maxLength` filters hints within the range (`pos`).

cristianoc: And why is it needed? I though vscode would decide what range to ask for.

cristianoc: Unless we measure a perf improvement it should not be there.

aspeddro: It's for these cases: https://github.com/rescript-lang/rescript-vscode/pull/453#discussion_r925964108



Some type hints are long and we want to filter. I added it to the binary to not filter on `server.ts` side.

aspeddro: It is more convenient to filter in bin analysis

zth:
I tried inlay hints in TS on a huge file, and it's noticably faster than ours. Maybe we should investigate how many calls are made, and if there's a bottleneck somewhere for us. This might of course also simply be the result of the TS server keeping much more cached in memory. We don't know.

But I'm not sure if we should do that in this PR, or in a follow up (and also investigate pipe calls etc). We've marked this as an experimental feature, and it's behind a setting that's off by default. More inclined to merge now and continue in separate PRs so this isn't left hanging for too long. I think it's in a good state already, especially given it's experimental.

analysis/src/Hint.ml
cristianoc: Oh got it. I thought was also some thing similar to range.
Let me check the documentation if it's clear what it is.

cristianoc: Checked: the config description that the user sees is a bit vague.
Can you clarify the description?

cristianoc:
Yup let's merge. I only asked for one small clarification for the config. Then we can merge away.
I thought it's good to go in details now that it's fresh in everybody's mind. Easier than having to come back later.

cristianoc:
@aspeddro thank you so much for doing this.

analysis/src/Hint.ml
aspeddro: Done https://github.com/rescript-lang/rescript-vscode/pull/453/commits/4df48ae9a1df6ab5f132ae7c67d813cbbaa190a8

PR #456, opened 2022-06-15
null check reanalyze output

zth:
So we don't accidentally crash the extension if we don't get proper json from reanalyze. It happens locally for me, which is another issue I'll investigate. But, this will at least ensure we don't crash.

### DIFF
--- a/client/src/commands/code_analysis.ts
+++ b/client/src/commands/code_analysis.ts
@@ -19,22 +19,22 @@ export type DiagnosticsResultCodeActionsMap = Map<
   { range: Range; codeAction: CodeAction }[]
 >;
 
+export type DiagnosticsResultFormat = Array<{
+  name: string;
+  kind: string;
+  file: string;
+  range: [number, number, number, number];
+  message: string;
+  annotate?: {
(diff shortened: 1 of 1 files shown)


client/src/commands/code_analysis.ts
```diff
@@ -202,13 +202,22 @@ export const runCodeAnalysisWithReanalyze = (
...
     diagnosticsResultCodeActions.clear();
 
+    let json: DiagnosticsResultFormat | null = null;
+
     try {
-      var json = JSON.parse(data);
+      json = JSON.parse(data);
     } catch (e) {
       window.showErrorMessage(
         `Something went wrong parsing the json output of reanalyze: '${e}'`
```
cristianoc, line 211: Can you add a bit more. How about encouraging reporting the issue.

cristianoc (approved):

client/src/commands/code_analysis.ts
zth: I'll add this in a separate PR when I've figured out what makes sense to tell the user here.

Issue #462, opened 2022-06-18
On click-hover on a module from a dependency, "Start a build" pops up.


cristianoc:
Eg. when click-hover on `RescriptRelay` in:
```rescript
ReactDOMExperimental.renderConcurrentRootAtElementWithId(
  <RescriptRelay.Context.Provider environment=RelayEnv.environment>
    <App />
  </RescriptRelay.Context.Provider>,
  "app",
)
```

cristianoc:
In general, "Start a build" should never appear when the file is in `node_modules` I guess.

CC @zth

zth:
That prompt is potentially triggered anytime the client opens a file. I'm thinking maybe we shouldn't prompt more than once per project root anyway? We could store that state next to the diagnostics-per-project state, and ensure we only ever show that prompt once.

PR #464, opened 2022-06-20
only prompt to start build once per project

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/462


server/src/server.ts
```diff
@@ -42,6 +42,7 @@ let projectsFiles: Map<
     openFiles: Set<string>;
     filesWithDiagnostics: Set<string>;
     bsbWatcherByEditor: null | ChildProcess;
+    hasPromptedToStartBuild: boolean;
```
cristianoc, line 45: Not about this PR per se.
Can we be super-explicit that this whole area is the server's global state?

PR #473, opened 2022-06-29
Fix interface generation if @inline is present

### LINKED ISSUE #425: Interface generation does not work if @inline is present
E.g. if I have a file `Example.res`:

```rescript
@inline
let gray = "#f0f0f0"
```

the create-interface command yields me this broken `Example.resi` file:

```rescript
@inline
let gray = "#f0f0f0"
```

expected:

```rescript
@inline
let gray: string
```

An-Tu:
Hello! I'm familiar with ReScript but it's my first try with OCaml. So if I did something wrong please feel free to correct me.

fix #425

### DIFF
--- a/analysis/src/CreateInterface.ml
+++ b/analysis/src/CreateInterface.ml
@@ -74,9 +74,7 @@ end = struct
     !res |> List.rev
 
   let contains attributeForSearch t =
-    t
-    |> List.find_opt (fun {name} -> name = attributeForSearch)
-    |> Option.is_some
+    t |> List.exists (fun {name} -> name = attributeForSearch)
 
   let toString t =
     if List.length t = 0 then ""
--- a/analysis/src/CreateInterface.ml
+++ b/analysis/src/CreateInterface.ml
@@ -162,15 +162,39 @@ let printSignature ~extractor ~signature =
           Buffer.add_string buf "\n"
       in
       processSignature ~indent rest
-    | Sig_value (_id, {val_kind = Val_prim prim; val_loc}) :: items
+    | Sig_value (id, ({val_kind = Val_prim prim; val_loc} as vd)) :: items
       when prim.prim_native_name <> "" && prim.prim_native_name.[0] = '\132' ->
-      (* Rescript primitive name, e.g. @val external ...
-         Copy the external declaration verbatim from the implementation file *)
+      (* Rescript primitive name, e.g. @val external ... *)
       let lines =
         let posStart, posEnd = Loc.range val_loc in
         extractor |> SourceFileExtractor.extract ~posStart ~posEnd
       in
-      Buffer.add_string buf ((lines |> String.concat "\n") ^ "\n");
+      let firstLine =
+        match lines with
+        | [] -> ""
+        | firstLine :: _ -> firstLine
+      in
+      let inlineAttr = "@inline" in
+      let inlineAttrLen = String.length inlineAttr in
+      let attrStartIdx = String.index firstLine '@' in
+
+      if
+        try String.sub firstLine attrStartIdx inlineAttrLen = inlineAttr
+        with _ -> false
+      then
+        (* Generate type signature for @inline declaration *)
+        let intend = String.make attrStartIdx ' ' in
+        let divider = if List.length lines > 1 then "\n" else " " in
+
+        let sigStr =
+          sigItemToString
+            (Printtyp.tree_of_value_description id {vd with val_kind = Val_reg})
+        in
+        Buffer.add_string buf (intend ^ inlineAttr ^ divider ^ sigStr ^ "\n")
+      else
+        (* Copy the external declaration verbatim from the implementation file *)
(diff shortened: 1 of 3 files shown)


zth:
Awesome! 🎉 I'll let @cristianoc comment on the OCaml itself, but the output generated by the test looks great.

An-Tu:
I realized that my new function `SourceFileExtractor.extractLine` isn't needed. So I removed it.

analysis/src/CreateInterface.ml
```diff
@@ -162,15 +162,39 @@ let printSignature ~extractor ~signature =
...
+        | [] -> ""
+        | firstLine :: _ -> firstLine
+      in
+      let inlineAttr = "@inline" in
+      let inlineAttrLen = String.length inlineAttr in
+      let attrStartIdx = String.index firstLine '@' in
+
+      if
+        try String.sub firstLine attrStartIdx inlineAttrLen = inlineAttr
+        with _ -> false
```
cristianoc, line 183: One would catch the specific exception `| Invalid_argument _ ->` .

But here one can also check the size beforehand.

cristianoc (commented):
Can you add 2 examples to the test set:
```rescript
@genType
@inline
let f2 = 10

@genType @inline
let f3 = 10
```

see what happens

cristianoc:
> Hello! I'm familiar with ReScript but it's my first try with OCaml. So if I did something wrong please feel free to correct me.
> 
> fix #425

That's an impressive first try!

An-Tu:
> Can you add 2 examples to the test set:
> 
> ```rescript
> @genType
> @inline
> let f2 = 10
> 
> @genType @inline
> let f3 = 10
> ```
> 
> see what happens

I missed that there can be more than one attribute. Fixed...I hope :)

analysis/src/CreateInterface.ml
```diff
@@ -31,6 +31,58 @@ module SourceFileExtractor = struct
...
+           if trimmedLine |> String.length > 0 && String.get trimmedLine 0 = '@'
+           then (
+             let attrDecIdxSeparator = ref (-1) in
+             let i = ref 1 in
+             let lineLength = String.length line in
+
+             while !attrDecIdxSeparator = -1 do
+               if !i < lineLength then
+                 let idx = !i in
+                 let currentChar = line.[idx] in
```
cristianoc, line 59: How do we know that `line[1]` exists?

cristianoc: I think it can be made more obvious.

An-Tu: Oh, thanks. I think I get it. I'll do it.

An-Tu:
@cristianoc One more try :)

analysis/src/CreateInterface.ml
```diff
@@ -31,6 +31,82 @@ module SourceFileExtractor = struct
...
+
+  let getLineIdx ((lineIdx, _), _) = lineIdx
+
+  let getAttrOffset ((_, attrOffset), _) = attrOffset
+
+  let getAttrName (_, attribute) = attribute
+
+  let makeAttr lineIdx attrOffsetStart attrOffsetEnd line =
+    ( (lineIdx, attrOffsetStart),
+      String.sub line attrOffsetStart (attrOffsetEnd - attrOffsetStart) )
```
cristianoc, line 55: Nit: to know that this is safe (not out of bound) you need to look at the callers of this function.
But if this function were to be moved into `make`, which is the only user, then the evidence that this is safe would be right there.

analysis/src/CreateInterface.ml
```diff
@@ -31,6 +31,82 @@ module SourceFileExtractor = struct
...
+           done;
+
+           match !state with
+           | Collect attrOffset ->
+             res :=
+               makeAttr lineIdx attrOffset (String.length line) line :: !res
+           | _ -> ());
+    !res |> List.rev
+
+  let rec contain attributeForSearch t =
```
cristianoc, line 79: This is OK.
It could also be a one-line `List.find`.

analysis/src/CreateInterface.ml
```diff
@@ -162,15 +238,29 @@ let printSignature ~extractor ~signature =
...
-         Copy the external declaration verbatim from the implementation file *)
+      (* Rescript primitive name, e.g. @val external ... *)
       let lines =
         let posStart, posEnd = Loc.range val_loc in
         extractor |> SourceFileExtractor.extract ~posStart ~posEnd
       in
-      Buffer.add_string buf ((lines |> String.concat "\n") ^ "\n");
+      let attributes = AttributesUtils.make lines in
+
+      if AttributesUtils.contain "@inline" attributes then
```
cristianoc, line 250: This is nice and clean.

cristianoc (approved):
This is looking great!
Let some style comments/thoughts. Take / ignore any one.

An-Tu:
@cristianoc Thanks a lot for your review! I fixed your comments.

PR #475, opened 2022-07-05
Test compiler 10.0 alpha

cristianoc:


analysis/tests/src/expected/Completion.res.txt
```diff
@@ -813,12 +813,6 @@ Completable: Cpath Module[For]
     "tags": [],
     "detail": "module",
     "documentation": null
-  }, {
```
cristianoc, line 816: This is the only change.
Expected as the module has been removed.

Issue #474, opened 2022-07-05
No type hint for react.component and ReactDOMStyle in the pre-release vscode editor plugin.


Mng12345:
### Environment
rescript version: 9.1.4
editor plugin version: pre-release after 1.30
system: windows10 x64

![mnggiflab-download (23)](https://user-images.githubusercontent.com/21278900/177282694-e71fc1de-2d95-4bab-b6b9-808ce6ac7485.gif)
![mnggiflab-download (24)](https://user-images.githubusercontent.com/21278900/177283510-7fc54490-3c98-465d-a5c3-27bf9d03417f.gif)
![mnggiflab-download (25)](https://user-images.githubusercontent.com/21278900/177283787-9f14eaf2-7acb-4e81-aaf5-a2fef57ae856.gif)

Mng12345:
The type hint for `make` is wrong.
![mnggiflab-download (27)](https://user-images.githubusercontent.com/21278900/177285326-16220f48-0938-40c8-be2e-8d356f16ad87.gif)
I think the logic of type hint may be not correct.
![mnggiflab-download (29)](https://user-images.githubusercontent.com/21278900/177286574-075f1583-b93e-403a-bcd0-c1b6a23dfad6.gif)

zth:
We've most likely found the cause for this. Will ping you again when there's a fix you can test. Thanks for reporting!

zth:
@Mng12345 this should be fixed if you download the latest `vsix` from CI.

PR #476, opened 2022-07-05
upgrade vscode-languageclient

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/474

This updates `vscode-languageclient` to the latest version, fixing a bug introduced when upgrading from v7 to v8 of that package. The bug was that the event for content in the editor being updated would arrive _after_ the request for formatting or completion. This caused bugs in both formatting and completion where the respective commands would operate on stale content if the commands ran fast enough, before the event signaling that the editor content had changed had time to dispatch.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


cristianoc (approved):
Amazing find.

PR #478, opened 2022-07-05
The path option was added to the rescript.settings

### LINKED ISSUE #412: Feature request: use `rescript` binary from ancestor `node_modules`
Thanks for creating this tool. I am really enjoying ReScript so far.

I am programming ReScript on my tiny chromebook, and replicating the large `rescript` node package in each package is taking up a lot of space. 

To address this, I installed `rescript` in `~/node_modules`. Building and running works fine! The binaries are detected and the output `.js` files go in the right place.

Only problem is, the language server does not find `node_modules/.bin/rescript`. My suggestion is to change [`findNodeBuildOfProjectRoot`](https://github.com/rescript-lang/rescript-vscode/blob/ce210510d3450b2a80a465947f95f7742d8ad7a3/server/src/utils.ts#L74) to look up the directory stack:

```typescript
export let findNodeBuildOfProjectRoot = (
  projectRootPath: p.DocumentUri
): null | { buildPath: p.DocumentUri; isReScript: boolean } => {
  while (true) {
    let rescriptNodePath = path.join(projectRootPath, c.rescriptNodePartialPath);
    let bsbNodePath = path.join(projectRootPath, c.bsbNodePartialPath);

    if (fs.existsSync(rescriptNodePath)) {
      return { buildPath: rescriptNodePath, isReScript: true };
    } else if (fs.existsSync(bsbNodePath)) {
      return { buildPath: bsbNodePath, isReScript: false };
    } else {
      let parent = path.dirname(projectRootPath);
      if (fs.existsSync(parent) && parent != projectRootPath) {
        projectRootPath = parent;
      }
      else {
        return null;
      }
    }
  }
};
```

I tried making this change and it seems to work fine (with [vim-rescript](https://github.com/rescript-lang/vim-rescript)).

An-Tu:
This PR fixes #412 .

I added the handling of the relative path and also the path with contains the home directory symbol. But I'm not sure that it is ok.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Automatically start ReScript's code analysis."
+				},
+				"rescript.settings.binaryPath": {
+					"type": ["string", "null"],
+					"default": null,
+					"description": "Path to the directory where ReScript binaries are. You can use it if you haven't or don't want to use the installed ReScript from node_modules in your project."
 				}
 			}
 		},
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -262,7 +262,20 @@ let openedFile = (fileUri: string, fileContent: string) => {
         // handle in the isResponseMessage check in the message handling way
         // below
       } else {
-        // we should send something to say that we can't find bsb.exe. But right now we'll silently not do anything
+        let binaryPath =
+          extensionConfiguration.binaryPath === null
+            ? path.join(projectRootPath, "node_modules", ".bin")
+            : extensionConfiguration.binaryPath;
+
+        let request: p.NotificationMessage = {
+          jsonrpc: c.jsonrpcVersion,
+          method: "window/showMessage",
+          params: {
+            type: p.MessageType.Error,
+            message: `Can't find ReScript binary on path ${binaryPath}`,
+          },
+        };
+        send(request);
       }
     }
 
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -24,9 +24,11 @@ import { filesDiagnostics } from "./utils";
 
 interface extensionConfiguration {
   askToStartBuild: boolean;
+  binaryPath: string | null;
 }
 let extensionConfiguration: extensionConfiguration = {
   askToStartBuild: true,
+  binaryPath: null,
 };
 let pullConfigurationPeriodically: NodeJS.Timeout | null = null;
 
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -75,6 +75,11 @@ let findBscBinary = (filePath: p.DocumentUri) =>
     ? utils.findBscBinaryFromProjectRoot(filePath)
     : utils.findBscBinaryFromConfig(extensionConfiguration.binaryPath);
 
+let getConjecturalDirOfBuildBinary = (projectRootPath: p.DocumentUri) =>
+  extensionConfiguration.binaryPath === null
+    ? path.join(projectRootPath, c.nodeModulesBinDir)
+    : extensionConfiguration.binaryPath;
+
 interface CreateInterfaceRequestParams {
   uri: string;
 }
@@ -267,17 +272,14 @@ let openedFile = (fileUri: string, fileContent: string) => {
         // handle in the isResponseMessage check in the message handling way
         // below
       } else {
-        let binaryPath =
-          extensionConfiguration.binaryPath === null
-            ? path.join(projectRootPath, "node_modules", ".bin")
-            : extensionConfiguration.binaryPath;
-
         let request: p.NotificationMessage = {
           jsonrpc: c.jsonrpcVersion,
           method: "window/showMessage",
           params: {
             type: p.MessageType.Error,
-            message: `Can't find ReScript binary on path ${binaryPath}`,
+            message: `Can't find ReScript binary in the directory ${getConjecturalDirOfBuildBinary(
+              projectRootPath
+            )}`,
           },
         };
         send(request);
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -65,6 +65,11 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
+let findBinary = (projectRootPath: p.DocumentUri) =>
+  extensionConfiguration.binaryPath === null
+    ? utils.findNodeBuildOfProjectRoot(projectRootPath)
+    : utils.findBinaryFromConfig(extensionConfiguration.binaryPath);
+
 interface CreateInterfaceRequestParams {
   uri: string;
 }
@@ -235,7 +240,7 @@ let openedFile = (fileUri: string, fileContent: string) => {
       // TODO: sometime stale .bsb.lock dangling. bsb -w knows .bsb.lock is
       // stale. Use that logic
       // TODO: close watcher when lang-server shuts down
-      if (utils.findNodeBuildOfProjectRoot(projectRootPath) != null) {
+      if (findBinary(projectRootPath) != null) {
         let payload: clientSentBuildAction = {
           title: c.startBuildAction,
           projectRootPath: projectRootPath,
@@ -1059,7 +1064,7 @@ function onMessage(msg: p.Message) {
       // TODO: close watcher when lang-server shuts down. However, by Node's
       // default, these subprocesses are automatically killed when this
       // language-server process exits
-      let found = utils.findNodeBuildOfProjectRoot(projectRootPath);
+      let found = findBinary(projectRootPath);
       if (found != null) {
         let bsbProcess = utils.runBuildWatcherUsingValidBuildPath(
           found.buildPath,
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -596,7 +596,11 @@ function format(msg: p.RequestMessage): Array<p.Message> {
   } else {
     // code will always be defined here, even though technically it can be undefined
     let code = getOpenedFileContent(params.textDocument.uri);
-    let formattedResult = utils.formatCode(filePath, code);
+    let formattedResult = utils.formatCode(
+      extensionConfiguration.binaryPath,
+      filePath,
+      code
+    );
     if (formattedResult.kind === "success") {
       let max = code.length;
       let result: p.TextEdit[] = [
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -70,6 +70,11 @@ let findBuildBinary = (projectRootPath: p.DocumentUri) =>
     ? utils.findBuildBinaryFromProjectRoot(projectRootPath)
     : utils.findBuildBinaryFromConfig(extensionConfiguration.binaryPath);
 
+let findBscBinary = (filePath: p.DocumentUri) =>
+  extensionConfiguration.binaryPath === null
+    ? utils.findBscBinaryFromProjectRoot(filePath)
+    : utils.findBscBinaryFromConfig(extensionConfiguration.binaryPath);
+
 interface CreateInterfaceRequestParams {
   uri: string;
 }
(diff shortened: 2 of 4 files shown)


An-Tu:
It seems that I missed. I thought that the relative paths handling is necessary but I double-checked and it looks like not. I removed the extra code.

server/src/server.ts
```diff
@@ -905,6 +908,13 @@ function onMessage(msg: p.Message) {
 
       if (initialConfiguration != null) {
         extensionConfiguration = initialConfiguration;
+        if (extensionConfiguration.path !== "" &&
+          extensionConfiguration.path[0] === "~"
```
cristianoc, line 912: How about removing this? Not sure why one needs to special case the home directory.

server/src/utils.ts
```diff
@@ -72,17 +72,30 @@ export let findBscNativeOfFile = (
...
 export let findNodeBuildOfProjectRoot = (
-  projectRootPath: p.DocumentUri
+  projectRootPath: p.DocumentUri,
+  pathToBinFromConfig: p.DocumentUri
 ): null | { buildPath: p.DocumentUri; isReScript: boolean } => {
-  let rescriptNodePath = path.join(projectRootPath, c.rescriptNodePartialPath);
-  let bsbNodePath = path.join(projectRootPath, c.bsbNodePartialPath);
+  if (pathToBinFromConfig !== "") {
+    if (fs.existsSync(pathToBinFromConfig)) {
+      if (pathToBinFromConfig.endsWith(c.rescriptBinName)) {
```
cristianoc, line 80: I'm confused here: it seems the config either sets the path to `bsb` or to `rescript`. Should it not give the path to the directory that contains those binaries instead?

Also, what about formatting? That uses the `bsc` command which does not seem to be considered here.

server/src/utils.ts
```diff
@@ -72,17 +72,30 @@ export let findBscNativeOfFile = (
...
+        return { buildPath: pathToBinFromConfig, isReScript: false };
+      }
+    }
+    return null;
+  } else {
+    let rescriptNodePath = path.join(projectRootPath, c.rescriptNodePartialPath);
+    let bsbNodePath = path.join(projectRootPath, c.bsbNodePartialPath);
 
-  if (fs.existsSync(rescriptNodePath)) {
```
cristianoc, line 80: How about putting formatting changes in a separate commit so one gets a clean commit history.

cristianoc (commented):
Thanks for working on this.
Was this PR tested?

amiralies:
I suggest to use a more descriptive name for `path`. `path` is very generic something like `pathToRescript` `rescriptPath` is better IMO.

server/src/server.ts
An-Tu: I tested it. If an absolute or relative path were used it is ok, but if the path starts with `~` then it's not working.

server/src/utils.ts
An-Tu: I looked at this issue #412 , and there was an example of using the path to the binary. But yes, maybe we should use the directory path.

server/src/utils.ts
An-Tu: Ops. I'll do it.

An-Tu:
> I suggest to use a more descriptive name for `path`. `path` is very generic something like `pathToRescript` `rescriptPath` is better IMO.

It will look like `"rescript.settings.path": "/some/path"` in settings. It doesn't seem very generic. Or do you think that there will be more settings in the future that can contain other paths?

An-Tu:
> Was this PR tested?

Yes, I tested manually.

cristianoc:
Perhaps binaryPsth ?

amiralies:
The thing is `path` itself doesn't show what it actually means. 
is this path to rescript binary? path to a custom analysis binary or even project root path?
some other extensions i've used also use more descriptive or scoped names.
for example iirc rust-analyzer uses `rust-analyzer.server.path` (path to language server) and flow uses  `pathToFlow`.

server/src/utils.ts
An-Tu: fixed

An-Tu:
I changed `path` to `binaryPath`

server/src/server.ts
An-Tu: I double-checked here. If remove this then paths that start with the home directory symbol are not working. Or do you suggest not handling this option at all? Or did I miss something?

package.json
```diff
@@ -131,6 +131,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Automatically start ReScript's code analysis."
+				},
+				"rescript.settings.binaryPath": {
+					"type": "string",
+					"default": "",
+					"description": "Path to the ReScript binary. You can use it if you haven't or don't want to use the installed ReScript binary from node_modules in your project."
```
cristianoc, line 138: Can you update the description: it's the path to the directory where binaries are.

server/src/server.ts
```diff
@@ -1013,7 +1052,10 @@ function onMessage(msg: p.Message) {
       // TODO: close watcher when lang-server shuts down. However, by Node's
       // default, these subprocesses are automatically killed when this
       // language-server process exits
-      let found = utils.findNodeBuildOfProjectRoot(projectRootPath);
+      let found = utils.findNodeBuildOfProjectRoot(
+        projectRootPath,
+        extensionConfiguration.binaryPath
```
cristianoc, line 1057: There are all these extra parameters added. Are they necessary?
Can `findNodeBuildOfProjectRoot` use `extensionConfiguration.binaryPath` directly without passing it as an argument?

server/src/server.ts
```diff
@@ -23,9 +24,11 @@ import { WorkspaceEdit } from "vscode-languageserver";
 
 interface extensionConfiguration {
   askToStartBuild: boolean;
+  binaryPath: string;
 }
 let extensionConfiguration: extensionConfiguration = {
   askToStartBuild: true,
+  binaryPath: "",
```
cristianoc, line 31: This is OK but perhaps a little fragile. How about using null here instead of "".

package.json
An-Tu: done

server/src/server.ts
An-Tu: done

server/src/server.ts
An-Tu: The function `findNodeBuildOfProjectRoot` is defined in the `server/src/utils.ts` and I am not sure how can it uses the `extensionConfiguration.binaryPath` without passing as an argument because the configuration is defined here in the `server/src/server.ts` :(

An-Tu: I can add a new function something like `findBinaryFromConfig` and use it. It will allow to not use of an extra argument in the `findNodeBuildOfProjectRoot`.

I will try

An-Tu: done

cristianoc (commented):
I was wondering whether this can possibly affect the existing way of determining how the path is computed.
I could not answer the question as there are a dozen changes happening at the same time.

If there's a way to change less, then it's easier to review.

If there isn't then this needs to be broken up into several small refactoring PRs where each one is obviously correct.

An-Tu:
@cristianoc I have divided all changes into commits which should be easy to check. But if it is not suitable please tell me and I will prepare the separate PRs.

cristianoc:
> @cristianoc I have divided all changes into commits which should be easy to check. But if it is not suitable please tell me and I will prepare the separate PRs.

This format is perfect thanks. Let me take a look.

cristianoc (approved):

server/src/server.ts
```diff
@@ -596,7 +596,11 @@ function format(msg: p.RequestMessage): Array<p.Message> {
   } else {
     // code will always be defined here, even though technically it can be undefined
     let code = getOpenedFileContent(params.textDocument.uri);
-    let formattedResult = utils.formatCode(filePath, code);
+    let formattedResult = utils.formatCode(
+      extensionConfiguration.binaryPath,
```
cristianoc, line 618: I don't understand this logic. The fact that there exists a configuration, which is optional, should not show up here.
The logic should be: find the binary path and use it. Either here or directly in the utils command.

server/src/utils.ts
```diff
@@ -94,21 +107,30 @@ type execResult =
...
+  code: string
+): execResult => {
   let extension = path.extname(filePath);
   let formatTempFileFullPath = createFileInTempDir(extension);
   fs.writeFileSync(formatTempFileFullPath, code, {
     encoding: "utf-8",
   });
   try {
+    // Try to find the bsc bin from the binaryPath setting from the configuration.
+    let bscPath = findBscBinFromConfig(pathToBinFromConfig);
```
cristianoc, line 136: This logic is a bit strange. Seems biased towards the configuration. While the configuration is the exception not the rule.

The logic I would expect is:
- find the normal binary path
- if the config has a setting use that instead

server/src/utils.ts
```diff
@@ -94,21 +107,30 @@ type execResult =
...
     // to recursively search for bsc.exe upward
-    let bscNativePath = findBscNativeOfFile(filePath);
+    bscPath = bscPath == null ? findBscNativeOfFile(filePath) : bscPath;
 
-    // Default to using the project formatter. If not, use the one we ship with
-    // the analysis binary in the extension itself.
-    if (bscNativePath != null) {
-      let result = childProcess.execFileSync(bscNativePath, [
+    // Default to using the formatter from the binaryPath setting from the configuration
+    // or the project formatter.
```
cristianoc, line 143: This comment is also inverted.
Default to use the binary path that ships with the analysis binary.
But if there is one in the config, override.

This is only about how the code reads, the semantics is the same.

server/src/server.ts
```diff
@@ -65,6 +65,11 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
+let findBinary = (projectRootPath: p.DocumentUri) =>
```
cristianoc, line 68: Which binary?
`findBuildBinary`.

cristianoc: All these functions, including `findNodeBuildOfProjectRoot` need to be renamed to contain the name `Build`.
So `findBuildBinaryFromProjectRoot` and `findBuildBinaryFromConfig`.

server/src/utils.ts
```diff
@@ -84,20 +84,33 @@ let findBscBinFromConfig = (
...
-  projectRootPath: p.DocumentUri
-): null | { buildPath: p.DocumentUri; isReScript: boolean } => {
-  let rescriptNodePath = path.join(projectRootPath, c.rescriptNodePartialPath);
-  let bsbNodePath = path.join(projectRootPath, c.bsbNodePartialPath);
-
-  if (fs.existsSync(rescriptNodePath)) {
-    return { buildPath: rescriptNodePath, isReScript: true };
-  } else if (fs.existsSync(bsbNodePath)) {
-    return { buildPath: bsbNodePath, isReScript: false };
+let findBinaryBase = ({
```
cristianoc, line 87: Add a comment to explain what this does.

server/src/server.ts
```diff
@@ -262,7 +262,20 @@ let openedFile = (fileUri: string, fileContent: string) => {
         // handle in the isResponseMessage check in the message handling way
         // below
       } else {
-        // we should send something to say that we can't find bsb.exe. But right now we'll silently not do anything
+        let binaryPath =
```
cristianoc, line 265: Isn't there already a function that does this: finding the binary directory?

cristianoc:
I made a few reviews, going commit after commit.
The semantics seems fine. All the comments are about naming and clarity.

cristianoc:
Overall I think this should be the main logic:
- There's a concept of binary path, that's the path where binaries are found.
- There's a concept of finding specific binaries.
- There's a list of binaries, probably a variant.
- There are functions to obtain specific binaries.

The config would be mentioned only once in one place: find the binary path.

An-Tu:
Thanks a lot for your review! In the near future, I will try to polish this PR.

An-Tu:
@cristianoc I fixed the PR according to your reviews. Also, I have a few thoughts above. Could you please look at them?

PR #479, opened 2022-07-07
Add `await` as keyword

IwanKaramazow:


zth (approved):
👍

cristianoc:
We can add this already now.

cristianoc:
And extend semantic highlighting.

zth:
Maybe we could make `async` a regular textmate highlight? Wouldn't it always be `async (`?

cristianoc:
> Maybe we could make `async` a regular textmate highlight? Wouldn't it always be `async (`?

That's the parser's job. Better not make assumptions, which might change, on what may or may not follow `async`. Unless the grammar is best effort and semantic highlighting has the final say.

IwanKaramazow:
> Maybe we could make `async` a regular textmate highlight? Wouldn't it always be `async (`?

It's unfortunately trickier, could be a normal function call:
```rescript
@val
external setTimeout: (unit => unit, int) => unit = "setTimeout"

let async = () => {
  setTimeout(() => {
    Js.log("async")
  }, 5000)
}

let f = async (          )
```

Arrow functions (both async and normal) introduce a lot of complexity for little gain.

Issue #480, opened 2022-07-07
No autocompletion in string interpolation


dkirchhof:
Unfortunatelly i don't get autocompletion in string interpolations. It worked in v 1.2.x but since 1.3.x it doesn't work anymore.

Now i have installed the version from this action: https://github.com/rescript-lang/rescript-vscode/actions/runs/2628522856  

While typing, no autocompletion:
![image](https://user-images.githubusercontent.com/2778643/177820336-6a50b964-87b8-4e74-9879-07fada9b45b2.png)

While hovering, there is a type hint:
![image](https://user-images.githubusercontent.com/2778643/177820386-bfa20957-f5c6-4792-91c1-b841d3f5d5a2.png)

cristianoc:
Thank you.
Could you cut it down to the shortest possible self-contained example? That will speed up fixing it.

dkirchhof:
Sure :)
Following code breaks:

```rescript
module hello = {
  world = "hello world"
}

let foo = "foo bar"

let message = `hello ${foo}`
```

While writing this code, I realized, that it works, if i remove the module:
```rescript
let foo = "foo bar"

let message = `hello ${foo}`
```

cristianoc:
> ```rescript
> module hello = {
>   world = "hello world"
> }
> 
> let foo = "foo bar"
> 
> let message = `hello ${foo}`
> ```

Would you give an example in correct syntax and explain what it is that does not work?

Btw have you tried asking for completion explicitly? For me it's ctrl+space.

cristianoc:
Just in case, autocomplete is not going to work for code with syntax errors. It can sometimes, but not always.

dkirchhof:
Sorry, was to quick. There are several errors :facepalm: 

With correct code, the autocompletion works.

```rescript
module Hello = {
    let world = "hello world"
}

let foo = "foo bar"

let message = `hello ${foo}`
```

I will try to find an example where it breaks (with correct syntax :D).

dkirchhof:
Well another try:

no autocompletion (the red variable):
```rescript
let red = "#ff0000"

let header = `
    color: ${red};
    background-color: ${red};
`
```

autocompletion (without the second line which also includes a variable):
```rescript
let red = "#ff0000"

let header = `
    color: ${red};
`
```

cristianoc:
Thanks I can see it too.

cristianoc:
[...]

PR #483, opened 2022-07-08
Fix autocompletion in template expressions.

cristianoc:
The issue turns out to be the location information given by the parser.
This was fixed in the syntax repo and synched here.

See https://github.com/rescript-lang/rescript-vscode/issues/480


analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1518,3 +1518,30 @@ Completable: Cpath array->ma
...
   }]
 
+Complete src/Completion.res 397:14
+posCursor:[397:14] posNoWhite:[397:13] Found expr:[396:14->397:20]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[396:14->397:16], ...[397:16->397:20])
+posCursor:[397:14] posNoWhite:[397:13] Found expr:[396:14->397:16]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[396:14->397:13], ...[397:13->397:16])
+posCursor:[397:14] posNoWhite:[397:13] Found expr:[397:13->397:16]
+Pexp_ident red:[397:13->397:16]
+Completable: Cpath Value[red]
```
cristianoc, line 1528: This works.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1518,3 +1518,30 @@ Completable: Cpath array->ma
...
+
+Complete src/Completion.res 402:22
+posCursor:[402:22] posNoWhite:[402:20] Found expr:[400:14->402:31]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[401:16->402:27], ...[402:27->402:31])
+posCursor:[402:22] posNoWhite:[402:20] Found expr:[401:16->402:27]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[401:16->402:27], ...[402:24->402:27])
+posCursor:[402:22] posNoWhite:[402:20] Found expr:[401:16->402:27]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[400:14->401:16], ...[401:16->402:24])
+posCursor:[402:22] posNoWhite:[402:20] Found expr:[401:16->402:24]
+Completable: Cnone
```
cristianoc, line 1545: This does not -- it never finds that it's possible to complete here.
Need to see what code the template parser generates. It's a big string append, and also need to check that the locations are correct or this will never be found.

cristianoc: While before there was an ident inside ghosts, it does not seem the case here.
Either there isn't one based on location, or there is an is not found.

cristianoc:
The location is now more accurate.
One remaining issue is that the location of `red` in ``` `color: ${red}; ` ``` only covers `ed`.

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1518,3 +1518,27 @@ Completable: Cpath array->ma
     "documentation": null
   }]
 
+Complete src/Completion.res 397:14
+posCursor:[397:14] posNoWhite:[397:13] Found expr:[396:14->397:20]
+Pexp_apply ...__ghost__[0:-1->0:-1] (...[396:14->397:16], ...[397:16->397:19])
+posCursor:[397:14] posNoWhite:[397:13] Found expr:[396:14->397:16]
```
cristianoc, line 1524: This `[396:14->397:16]` should cover the location of `red` but only covers `ed`.

cristianoc:
This now works as expected. Need to move the changes to the syntax repo.

Issue #477, opened 2022-07-05
Errors get cleaned from problems tab without being resolved


fhammerschmidt:
In a project which is not a monorepo (no `namespace: true`, just one bsconfig)

1. `rescript build -with-deps -w` 
2. An error appears both on the cli of the compiler and the problems tab (in my case a simple "module X cannot be found")
3. Close the VSCode tab of the File where the error appeared.
4. Open it again
5. The error vanishes from the problems tab.

Tested with current master (4f8d3f9)

cristianoc:
Are the errors still in the compiler log?
The expectation is they are. But just to double check.

fhammerschmidt:
Yes, they are.

cristianoc:
It could be that on file close the diagnostics entry is removed and on open it is not restored.
If that's what's going on, this is mixing up storing data with displaying it. Does not seem necessary. The data should always reflect the current status of the compiler log. Separately, one can mark what files are open, if there is a reason to keep that info.

cristianoc:
This:
```ml
let closedFile = (fileUri: string) => {
  let filePath = fileURLToPath(fileUri);

  stupidFileContentCache.delete(filePath);
```

cristianoc:
> This:
> 
> ```ocaml
> let closedFile = (fileUri: string) => {
>   let filePath = fileURLToPath(fileUri);
> 
>   stupidFileContentCache.delete(filePath);
> ```

Not that, but definitely there's `deleteProjectDiagnostics` which also deletes the `projectFiles` entry.

@zth what do you think the fix should be?

zth:
Will have a look at this later today. I think this needs to be fixed before releasing `1.4.0`.

zth:
The root cause here seems to be the continuous syntax errors diagnostics. I think they accidentally reset diagnostics because they do their own updates. Will experiment with how we can either unify those, or ensure they use separate "namespaces" (if that's possible). Reporting back here a bit later.

zth:
@fhammerschmidt can you verify that https://github.com/rescript-lang/rescript-vscode/pull/484 fixes for you as well? I can't reproduce anymore anyway.

PR #484, opened 2022-07-08
fix problem tab accidentally being cleared

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/477

Problems tab was accidentally being cleared because the logic handling syntax error diagnostics did not account for pre-existing compiler errors.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -82,7 +82,9 @@ let openCompiledFileRequest = new v.RequestType<
   void
 >("rescript-vscode.open_compiled");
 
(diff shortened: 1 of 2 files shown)


server/src/server.ts
```diff
@@ -41,6 +42,8 @@ let projectsFiles: Map<
   {
     openFiles: Set<string>;
     filesWithDiagnostics: Set<string>;
+    filesDiagnostics: filesDiagnostics;
```
zth, line 45: These might be possible to unify at some point, but I want to make minimal changes in this PR.

PR #491, opened 2022-07-13
remove empty line from diagnostic message

aspeddro:
Before:

image

After:

image

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -67,13 +67,13 @@ export let findBscBinary = (
 
 type execResult =
   | {
-      kind: "success";
-      result: string;
-    }
+    kind: "success";
+    result: string;
+  }
   | {
-      kind: "error";
-      error: string;
-    };
+    kind: "error";
+    error: string;
+  };
 
 export let formatCode = (
   bscPath: p.DocumentUri | null,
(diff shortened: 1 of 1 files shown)


cristianoc:
Thank you. Wondering whether there's ever the case where one would want to respect the presence of an empty line in an error message.

For this one, would you also look at the corresponding error emitted by the compiler (in the compiler repo)? As I imagine this is equally undesirable on the command line.

aspeddro:
Some cases with empty line from [`super_errors`](https://github.com/rescript-lang/rescript-compiler/tree/master/jscomp/build_tests/super_errors) test.

```
We've found a bug for you!
  /home/pedro/Desktop/learning-rescript/src/intro/Error.res:7:24

  5 │ // let a: a = (Some(Some(5)): b)
  6 │ let a = list{1, 2, 3}
  7 │ List.map(n => n +. 2., a)

  This has type: list<int>
  Somewhere wanted: list<float>
  
  The incompatible parts:
    int vs float
  
  You can convert int to float with Belt.Int.toFloat.
  If this is a literal, try a number with a trailing dot (e.g. 20.).
```

```
We've found a bug for you!
  /home/pedro/Desktop/learning-rescript/src/intro/Error.res:5:13-28

  3 │ type a = option<aa>
  4 │ type b = option<bb>
  5 │ let a: a = (Some(Some(5)): b)
  6 │ let a = list{1, 2, 3}
  7 │ // List.map(n => n +. 2., a)

  This has type: b (defined as option<bb>)
  Somewhere wanted: a (defined as option<aa>)
  
  The incompatible parts:
    bb (defined as option<int>) vs aa (defined as option<string>)
    
    Further expanded:
      int vs string
  
  You can convert int to string with Belt.Int.toString.
```

cristianoc:
This needs to be handled on the super-errors side.
Not sure one can paper over all the things needed to show error nicely in the editor.
I would avoid complicating editor logic, unless it's as a temporary thing to be removed later.

cristianoc:
Specifically: compiler log should emit a structured format in json. The format should include file path and line info, name, severity, text to be shown.

Precedent: the analysis integration already does that, and one might even reuse exactly the same mechanism.
The editor should pick that up and decide how to print things.

PR #493, opened 2022-07-14
Explore type based autocomplete

zth:
Work in progress. Exploring type based autocomplete.


analysis/tests/src/expected/TypeContextCompletion.res.txt
```diff
@@ -0,0 +1,12 @@
+Complete src/TypeContextCompletion.res 10:44
+posCursor:[10:44] posNoWhite:[10:43] Found expr:[10:11->18:1]
+Pexp_apply ...[10:11->10:30] (~someVariant10:32->10:43=...[13:0->17:3])
+Completable: CnamedArg(Value[someVariantToString], "", [someVariant])
```
cristianoc, line 4: So this is where it gets a bit hairy. Not sure why it decides it's a named arg completion given that it's in the expression assigned to the label. It could be because that expression is exactly where parser recovery happened and some heuristic is needed because of the uncertainty that parser recovery brings.
Something orthogonal, but need to figure it out in order to understand this example.

Anyhow the `CnamedArg` completion finds the function name, and the labels already visited: `[someVariant]` so they are not repeated. This is not going to find anything as it's the only label.

analysis/tests/src/expected/TypeContextCompletion.res.txt
```diff
@@ -0,0 +1,12 @@
...
+Found type for function (~someVariant: someVariant) => string
```
cristianoc, line 5: Need to check where this is coming from: presumably named arg completion finding the type of the function.

analysis/tests/src/expected/TypeContextCompletion.res.txt
```diff
@@ -0,0 +1,12 @@
...
+[]
```
cristianoc, line 6: As predicted, no result.

cristianoc (commented):
First, let's walk through what's happening with this example.

analysis/tests/src/expected/TypeContextCompletion.res.txt
cristianoc: OK looking at the code, I think we're hitting this case:
```ml
    | [] ->
      if posAfterFunExpr <= posBeforeCursor && posBeforeCursor < endPos then
        Some (CnamedArg (contextPath, "", allNames))
```

So we went through the entire application and did not find anything interesting. So we'll just default to named arg completion.
This can anyway be overridden by visiting sub-expressions later. The innermost AST expression where something interesting is found wins. Because the innermost is the most precise hit of the cursor.

To see why this case is needed one can just turn it off and see what tests fail.
Looks like this is one such example: ```let x = Lib.foo(~```

cristianoc:
So let's avoid getting carried away by parser recovery which needs to be taken into account but is orthogonal.

What's going on is we do find a label, a function, and an expression being assigned to the label which hits the cursor. Only there's no code to mark that event.
What's needed is code to recognise that event, and emit a new completion primitive for expressions in a type context. E.g. `CtypedContext`.

I think the next step is to
- Recognise when it is the case that we hit an expression assigned to a label and the label does not hit the cursor but the expression does.
- When that happens, print "found typed context" in debug mode. That print should be visible in the test output.
- As a refinement, also print the name of the function and the name of the label.

Back to you @zth

cristianoc:
A print just like others in the file e.g. this one `        if debug then
          Printf.printf "Attribute id:%s:%s label:%s\n" id.txt
            (Loc.toString id.loc) label;
`

zth:
@cristianoc I think I got it: https://github.com/rescript-lang/rescript-vscode/pull/493/commits/5ed8bb42b6d7c85ad056c6c4b1ad0675a8937d0b

I added a regular labelled argument completion in `TypedContextCompletion.res` to help distinguish between the various types.

zth:
Continued my exploration here: https://github.com/rescript-lang/rescript-vscode/pull/497

analysis/src/CompletionFrontEnd.ml
```diff
@@ -149,11 +150,19 @@ let findNamedArgCompletable ~(args : arg list) ~endPos ~posBeforeCursor
...
       if
         labelled.posStart <= posBeforeCursor
         && posBeforeCursor < labelled.posEnd
+        && charBeforeCursor <> Some '='
+        (* We need to account for empty assignments being parsed as regular labelled arguments without an assignment *)
       then Some (Completable.CnamedArg (contextPath, labelled.name, allNames))
-      else if exp.pexp_loc |> Loc.hasPos ~pos:posBeforeCursor then None
+      else if
+        exp.pexp_loc |> Loc.hasPos ~pos:posBeforeCursor
+        || charBeforeCursor = Some '='
```
cristianoc, line 162: Let's go slowly and stick to the script.
This is what you do when there is no other choice.
I would like to see how far we can go before trying to do this.

Also, we don't know if it's the character before the cursor.

cristianoc:
I'd like to go slowly and make a tiny step at a time.

Let's see how far we can push it before checking for `=`.
It will mean trying several examples.
And if it does not work for all of them, something will be learned from that.

zth:
@cristianoc I've removed the `'='` special treatment, and added a few more examples. Looks like a typed context is picked up for everything but an empty assignment, which is likely due to parser recovery omitting the assignment if it's empty. 

Having type based completion on an empty assignment is critical I'd say. What do you think is the next step? And are there more examples you'd like to see added?

cristianoc:
Need to dig into the example that does not work. With the usual checking in and commenting on the debug output.

zth:
@cristianoc Ok, think I made some more progress. I stole a trick from the JSX prop completion that helps me identify the empty assignment. 

This is not in the commit, but I also verified that it's indeed a `%rescript.exprhole` that's being inserted in the expression by the parser error recovery. So the parser does recover the error properly and nicely, I was wrong.

analysis/src/SharedTypes.ml
```diff
@@ -433,6 +433,7 @@ module Completable = struct
     | Cpath of contextPath
```
cristianoc, line 433: Would you start adding comments to Cpath and contextPath?
There's the question of whether there should be an entirely different case of type t, or whether typed context info should be integrated into the existing ones.

cristianoc:
This is great progress!
I think we can move on to the next phase and produce the desired result for all the existing examples. Without worrying too much about how things are represented or code duplication for now. (We'll do that after tackling at least one more kind of type context).

analysis/src/SharedTypes.ml
zth: Would you mind helping me out with adding comments to those that exist already? I have a hard time parsing what they are/mean.

zth:
> This is great progress! I think we can move on to the next phase and produce the desired result for all the existing examples. Without worrying too much about how things are represented or code duplication for now. (We'll do that after tackling at least one more kind of type context).

Awesome! So, essentially, the next step is to produce the list of constructors for completion, disregarding exactly how, what code is shared, how it's represented exactly etc. Did I get that right?

cristianoc:
Right

analysis/src/SharedTypes.ml
```diff
@@ -425,6 +425,12 @@ module Completable = struct
     | CPObj of contextPath * string
     | CPPipe of contextPath * string
 
+  type typedContext = NamedArg of string
+
+  let typedContextToString typedContext =
+    match typedContext with
+    | NamedArg argName -> "NamedArg(" ^ argName ^ ")"
```
zth, line 432: This was a very quick solution just to complete the actual items. I understand this is probably not the right way to do it.

analysis/tests/src/expected/TypeContextCompletion.res.txt
```diff
@@ -11,36 +11,184 @@ Found type for function (~someVariant: someVariant) => string
...
+    "tags": [],
+    "detail": "Five(int)\n\n",
+    "documentation": null
+  }, {
+    "label": "Six(_)",
+    "kind": 4,
+    "tags": [],
+    "detail": "Six(option<string>)\n\n",
+    "documentation": null
+  }]
```
zth, line 56: Barring details like the payloads, filtering according to what the user has already typed, whether we should use snippets etc, this at least produces the output we're after - all of the variant constructors.

zth:
@cristianoc very crude version implemented that produces the correct output. I've tried to familiarize myself with the code a bit. Feel like I'm slowly getting there.

analysis/src/CompletionBackEnd.ml
```diff
@@ -992,6 +992,18 @@ let rec extractRecordType ~env ~package (t : Types.type_expr) =
     | _ -> None)
   | _ -> None
 
+let rec extractVariantType ~env ~package (t : Types.type_expr) =
```
cristianoc, line 995: There's probably something like that already. Or maybe not.
Anyhow no need to worry about this for now.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1686,3 +1698,62 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
...
+      (* TODO: Should probably share this with the branch handling CnamedArg... *)
+      let labels =
+        match
+          cp
+          |> getCompletionsForContextPath ~package ~opens ~rawOpens ~allFiles
+               ~pos ~env ~exact:true ~scope
+          |> completionsGetTypeEnv
+        with
+        | Some (typ, _env) ->
+          let rec getLabels ~env (t : Types.type_expr) =
```
cristianoc, line 1713: This must exist already. Doesn't matter for now.

analysis/src/SharedTypes.ml
cristianoc: I think this is just debug printing. Unless the comment refers to the type definition, which is perfectly fine for now.

analysis/src/SharedTypes.ml
```diff
@@ -425,6 +425,12 @@ module Completable = struct
     | CPObj of contextPath * string
     | CPPipe of contextPath * string
 
+  type typedContext = NamedArg of string
```
cristianoc, line 428: You can add a comment: the argument name of a function type.

analysis/tests/src/expected/TypeContextCompletion.res.txt
cristianoc: Filtering is the next step. Nothing else need to be done.
Snippets etc we don't have to worry about. Perhaps later.

analysis/tests/src/expected/TypeContextCompletion.res.txt
```diff
@@ -0,0 +1,194 @@
...
+    "label": "Four",
+    "kind": 4,
+    "tags": [],
+    "detail": "Four\n\n",
+    "documentation": null
+  }, {
+    "label": "Five(_)",
+    "kind": 4,
+    "tags": [],
+    "detail": "Five(int)\n\n",
```
cristianoc, line 92: There's the question of what one would like to show here.
O I get it. Now I understand what "snippets" means. They're the templates for a variant with payloads?
If so then sure why not tackle them next.

cristianoc (commented):
Nice!
Very good progress.

analysis/tests/src/expected/TypeContextCompletion.res.txt
zth: Yeah, essentially we could return this: `Five($1, $2)` and the editor would put the cursor on `$1` first, and then a tab would put the cursor on `$2`. This can be quite convenient when autocompleting things like this. For example, later when we do sub expression type aware completion (that's not a proper sentence...), the experience could be quite nice. Something like:
1. Complete `someFn(~somethingOptional=`, gives `Some($1)`
2. Select that completion, trigger autocomplete again. Get a list of all variant constructors if it's a variant, or all identifiers in scope matching the inner type of `somethingOptional`, etc

We do need to support both with and without snippets, because I'm unsure if all editors support them. I'll have a look at that later.

cristianoc: Don't know if "sub expression type aware completion" is a proper sentence but is sure sounds cool.
What kind of format does vocode support for these kinds of templates? Does it expect things with dollar signs.
It would be great if there were a way to convey the type to, but don't know what format is possible and one would not like super-long names. But if the names were hoverable, then that would be pretty sweet.

Not sure if you want to go down the rabbit hole now, or move on with other aspects and come back to this later.

analysis/src/SharedTypes.ml
```diff
@@ -425,6 +425,12 @@ module Completable = struct
...
+
+  let typedContextToString typedContext =
+    match typedContext with
+    | NamedArg argName -> "NamedArg(" ^ argName ^ ")"
+
   type t =
     | Cdecorator of string  (** e.g. @module *)
     | CnamedArg of contextPath * string * string list
```
cristianoc, line 436: `~prefix` with `[this, that]` seen.
So it would complete a label whose name starts with `prefix` and excludes labels `this` and `that` which have been used already and should not be suggested.

analysis/src/SharedTypes.ml
cristianoc: `Cpath` is for complex expressions

cristianoc: `CPString` is for string constants, used in pipe expressions `34 -> ...`
`CPArray` is for arrays used in pipe expressions `[] -> ....`
 `CPApply of contextPath * Asttypes.arg_label list` is application with a list of labels
 
`CPId of string list * completionContext` is for a path `A.` or `A.B.C.` and the rhs is the rest to be completed

 `CPField of contextPath * string` is the `.z` of record access `expr.z` 
  `CPObj of contextPath * string` is the `["hello"]` of object access `expr["hello"]`
  `CPPipe of contextPath * string` is `expr->foo` where `foo` is the function whose first argument is `expr`.

analysis/src/SharedTypes.ml
zth: Yes, it was referring to the type def.

analysis/src/CompletionBackEnd.ml
zth: I refactored this into a general one for extracting a type, so the code is slightly more ready for completing more things than just variants.

analysis/src/CompletionBackEnd.ml
zth: Yeah, it exists in another branch. I think it can be done in a better way, so can refactor that later.

analysis/src/SharedTypes.ml
zth: I made it a record instead and added comments.

analysis/tests/src/expected/TypeContextCompletion.res.txt
zth: We can come back later, there's other things snippet syntax would be nice to leverage for too. But yeah, too much of a side track now.

analysis/src/SharedTypes.ml
zth: I've done that now, _but_ I haven't done anything about `this`/`that` seen, because I don't think it makes sense in the context of completing a labelled argument? There's nothing else to have been seen already, right?

zth:
@cristianoc did a little bit of refactoring, and got filtering on prefix working. What do you reckon is the next step?

zth:
@cristianoc I couldn't help myself, so I did a bit of exploration on completing polyvariants in addition to regular variants in https://github.com/rescript-lang/rescript-vscode/pull/493/commits/ffa06f417329040f43fc05b916d918103c9ed94b

We can just ignore that for now if it interferes with the process we're doing. Or it's easy to revert as well.

analysis/src/CompletionBackEnd.ml
```diff
@@ -711,6 +711,16 @@ let detail name (kind : Completion.kind) =
   | FileModule _ -> "file module"
   | Field ({typ}, s) -> name ^ ": " ^ (typ |> Shared.typeToString) ^ "\n\n" ^ s
   | Constructor (c, s) -> showConstructor c ^ "\n\n" ^ s
+  | PolyvariantConstructor {name; payload} -> (
+    "#" ^ name
+    ^
+    (* TODO: I get the feeling there's a more clever way to do this printing...
```
cristianoc, line 717: This code looks fine. No need to be too clever. It just reflects the internal representation of poly variants.

analysis/src/CompletionBackEnd.ml
```diff
@@ -992,16 +1002,35 @@ let rec extractRecordType ~env ~package (t : Types.type_expr) =
     | _ -> None)
   | _ -> None
 
-let rec extractVariantType ~env ~package (t : Types.type_expr) =
+type extractedType =
+  | Declared of QueryEnv.t * Type.t Declared.t
+  (* Polyvariants seem to be "inlined" in the type system, so we extract what we need from them here. *)
```
cristianoc, line 1007: It's because they're structural types, so you don't have go chase type declarations the way one would do with nominal types.

analysis/src/CompletionBackEnd.ml
```diff
@@ -992,16 +1002,35 @@ let rec extractRecordType ~env ~package (t : Types.type_expr) =
...
     match References.digConstructor ~env ~package path with
-    | Some (env, ({item = {kind = Variant constructors}} as typ)) ->
-      Some (env, constructors, typ)
     | Some (env, {item = {decl = {type_manifest = Some t1}}}) ->
-      extractVariantType ~env ~package t1
+      extractType ~env ~package t1
+    | Some (env, typ) -> Some (Declared (env, typ))
     | _ -> None)
+  | Tvariant {row_fields} ->
+    (* Polyvariants seem "inlined" into the type system, so we extract just
```
cristianoc, line 1020: I've seen this comment before.

analysis/src/SharedTypes.ml
```diff
@@ -425,11 +428,18 @@ module Completable = struct
...
 
-  type typedContext = NamedArg of string
+  type namedArg = {
+    (* What the user has already started writing, if anything *)
+    prefix: string;
+    (* Name of the argument *)
+    name: string;
+  }
+
+  type typedContext = NamedArg of namedArg
```
cristianoc, line 438: You can use inlined records.
Also `NamedArg` is good but does not covey the fact this is an application context. Perhaps specifying that in the comments on the type definition is the way to go.

analysis/src/SharedTypes.ml
```diff
@@ -425,11 +428,18 @@ module Completable = struct
     | CPObj of contextPath * string
     | CPPipe of contextPath * string
 
-  type typedContext = NamedArg of string
+  type namedArg = {
+    (* What the user has already started writing, if anything *)
+    prefix: string;
+    (* Name of the argument *)
+    name: string;
```
cristianoc, line 435: Perhaps `label` as in labelled argument, instead of `name`? The thing is `prefix` is also going to be the name of something, in a slightly different sense.

cristianoc (commented):
That's quite a lot of progress.

analysis/src/SharedTypes.ml
cristianoc: Perhaps the name suggestion will resolve itself once pattern matching contexts are added.

.vscode/settings.json
```diff
@@ -5,5 +5,9 @@
 	"typescript.preferences.quoteStyle": "single",
 	"editor.codeActionsOnSave": {
 		"source.fixAll.eslint": true
+	},
+	"ocaml.sandbox": {
+		"kind": "opam",
```
cristianoc, line 10: forgot why we have settings checked in at all

analysis/src/SharedTypes.ml
```diff
@@ -425,6 +428,80 @@ module Completable = struct
...
+      "JsxProp(<" ^ (componentPath |> ident) ^ " " ^ propName ^ "=" ^ " />)"
+    | NamedArg {label; contextPath} ->
+      "NamedArg(" ^ label ^ ", " ^ (contextPath |> contextPathToString) ^ ")"
+
+  type typedContextMeta = {
+    (* What the user has already started writing, if anything. *)
+    prefix: string option;
+    (* Record fields already written by the user, etc.
+       This is contextual of course, but putting it here in the general type to simplify things. *)
+    alreadySeenIdents: string list option;
```
cristianoc, line 502: why the option

analysis/src/SharedTypes.ml
```diff
@@ -425,6 +428,80 @@ module Completable = struct
...
+      ^ ")"
+    | CPArray -> "array"
+    | CPId (sl, completionContext) ->
+      completionContextToString completionContext ^ list sl
+    | CPField (cp, s) -> contextPathToString cp ^ "." ^ str s
+    | CPObj (cp, s) -> contextPathToString cp ^ "[\"" ^ s ^ "\"]"
+    | CPPipe (cp, s) -> contextPathToString cp ^ "->" ^ s
+
+  (* Temporary while I figure things. Pretty sure this should be integrated into contextPath when this is all finished *)
+  type howToRetrieveSourceType =
```
cristianoc, line 473: Not too surprising that there's a list of possible patters to be filled.
This is not very different from how now one extract a `contextPath` as "instructions" to be executed by the back-end.

cristianoc (commented):
A  few very high level comments.

analysis/tests/src/expected/BrokenParserCases.res.txt
```diff
@@ -0,0 +1,62 @@
+Dump AST src/BrokenParserCases.res 1:22
+
+Source:
```
cristianoc, line 3: If the source is repeated, how about also indicating what line number this is?

analysis/tests/src/expected/BrokenParserCases.res.txt
```diff
@@ -0,0 +1,62 @@
...
+// let _ = ({someProp: , otherProp: 123}, 123)
+//                    ^ast
+
+<*>Pstr_value(
```
cristianoc, line 7: What does `<*>` indicate? Seems redundant information-wise, but I guess it follows some existing convention of something.

zth: No existing convention, just something I made up. Here's a legend:
- `<*>` means it has the cursor
- `<x>` means it has an empty loc

It's redundant at the top level `Pstr_value` etc because the value itself would never get printed unless the cursor was there. But, I was confused several times myself _not_ seeing `<*>`, so I added it there too. Imo having it there too makes it easier to scan for "anchor points" of the cursor, like "there's no cursor (but a broken loc) inside of the structure item, but the structure item itself has the cursor, so we can use that to figure out the context".

analysis/tests/src/expected/BrokenParserCases.res.txt
zth: Yes, for sure! I need to re-do this part though, it's very hacky and breaks for a lot of cases. I did however find it quite convenient to be able to just scan the resulting txt and have everything I need to think about the AST there.

analysis/tests/src/expected/BrokenParserCases.res.txt
cristianoc: Got it. How about the ^ symbol instead of * so it's consistent with other tests.

Also question on terminology: is this a hit test? So everything containing the cursor has it, not a single node?

zth: Sure, that sounds reasonable.

If I understand what you mean then yes, it's a hit test in the sense that the structure item printed needs to have it, but nothing else inside of the structure item technically needs to have the cursor for it to hit.

cristianoc: https://en.wikipedia.org/wiki/Hit-testing

zth: Ahhh yes. Now I understand. The answer is yes, it's a hit test, and everything containing the cursor has it, not just a single node.

zth:
This work is superseeded by a bunch of other PRs. Closing.

Issue #386, opened 2022-04-07
parseCompilerLogOutput crashes when first error is graphql deprecation warning


AlexMoutonNoble:
Hi Folks. 
Found a case here.
given a codebase using rescript-apollo, with a deprecated field warning, the parser will fail its "File " branch based on the next line not being Warning, and end up in its last case where it attempts to push a follow up line to an empty parsedDiagnostics array.

I can see some pushback on the apollo output here maybe, but would it be better to make the "File " branch more permissive, or allow a new parsed diagnostic to be created there in that non-empty-line case?

Thanks
Alex

```
File "/Users/alexmouton/Documents/code/noble/web/src/packs/reactor/PackCapabilityRunSweepUpdateHf.res", line 45, characters 8-26:
45 | ........entDataRecordId
45 |   .....
Warning 22 [preprocessor]: Field "parentDataRecordId" has been deprecated. Reason: Prefer parent_data_record
File "/Users/alexmouton/Documents/code/noble/web/src/packs/reactor/PackCapabilityRunSweepUpdateHf.res", line 42, characters 40-53:
42 | ........................................ectedInput
42 |   .......
```

zth:
Good catch! I'll soon be working on adding a test suite for a few of these things, I'll keep you posted here so we can add this as a test case at a later point.

AlexMoutonNoble:
@zth I can make a PR if you have a preference for a fix?

zth:
@AlexMoutonNoble sorry, I blanked on this! Turns out I won't be doing a test setup just now, so if you feel like taking a stab at fixing this and PR:ing, that'd be great!

Do you want me to give some background/try and guide you to where to fix it?

AlexMoutonNoble:
I remember it being pretty clearcut but happy to hear any guidance you have
that will make the PR go easier

zth:
Yeah, re-reading your initial details it looks like you already have a clear idea of what's needed. Feel free to ping me otherwise! 😄

zth:
@AlexMoutonNoble did you end up having a look at this?

AlexMoutonNoble:
I havnt yet but still have it in mind

aspeddro:
I got this error here is the log

```sh
cat lib/bs/.compiler.log
```

```
#Start(1660943070627)
File "/home/pedro/Desktop/rescript-apollo-client/EXAMPLES/src/fragmentsUsage/Fragments.res", lines 11-13, characters 6-7:
[...]

PR #495, opened 2022-07-15
Bugfix: add safeguards for array push in server

steinararnason:
I hit a pretty annoying error with the plugin where the server would crash upon opening one of the rescript projects at work with the following error

image

I've added this to my local version and the error subsided

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -575,9 +575,11 @@ export let parseCompilerLogOutput = (
       //   10 ┆
     } else if (line.startsWith("  ")) {
       // part of the actual diagnostics message
-      parsedDiagnostics[parsedDiagnostics.length - 1].content.push(
-        line.slice(2)
-      );
+      if (parsedDiagnostics[parsedDiagnostics.length - 1] !== undefined) {
+        parsedDiagnostics[parsedDiagnostics.length - 1].content.push(
+          line.slice(2)
+        );
+      }
     } else if (line.trim() != "") {
       // We'll assume that everything else is also part of the diagnostics too.
       // Most of these should have been indented 2 spaces; sadly, some of them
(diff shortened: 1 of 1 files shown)


cristianoc:
Actually that makes it impossible to debug what's going on, an fail silently.
Would you be able so send a repro instead?

steinararnason:
I would if I knew what exactly caused this, I tried figuring it out but felt like I shouldn't spend too much time of it due to other projects/deadlines.
If you feel this would be a change for the worse feel free to close it down, but I find it strange that it tries to update something that doesn't exist in the first place 😅 
Without this the plugin in my editor just crashes and stops working completely (same for other developers on the project).

cristianoc:
There is server restart.
But in general, It would be preferable to replace this with something that issues a message asking to report an issue. Instead of eating it up.

@zth any suggestion on how to report where?

steinararnason:
Probably wont be able to take a better look at this today but I'll see if I can some time over the weekend / early next week.
Thanks for the information 😄

zth:
> There is server restart. But in general, It would be preferable to replace this with something that issues a message asking to report an issue. Instead of eating it up.
> 
> @zth any suggestion on how to report where?

Yeah, sounds like a good idea to encourage reporting + what to report. I'll open a new issue and have a look at that soon.

@steinararnason I agree with @cristianoc, while pushing to something that doesn't exist is of course wrong, I think we'd rather let things error at this point and find a better way for people to report the issue. This crash does mean there's a bug somewhere in our parsing logic, and it's better long term to fix that rather than "suppressing" it, even if that unfortunately means crashing for end users.

zth:
@steinararnason seems like the same issue as https://github.com/rescript-lang/rescript-vscode/issues/386, right?

steinararnason:
yeah that looks like the same issue to me, would you like me to close this PR?

zth:
@steinararnason yes I think we can close this. Let us know if you'd be interested in taking a stab at fixing the underlying issue, and we can help you if needed.

PR #497, opened 2022-07-15
basic completion for variant constructors

zth:
This is a crude attempt at continuing https://github.com/rescript-lang/rescript-vscode/pull/493 by completing variant constructors in labelled argument assignments. I looked around and tried to follow roughly how things have been done before. It's not complete, but it's a start, and it does actually do completion.


cristianoc:
Will look at it later. Need to cover basic earlier steps first.

Issue #496, opened 2022-07-15
New formatting producing syntax errors


remitbri:
Hello

I'm on a project using rescript 9.1.4, and I've been using the extension v1.3.0 for quite a while, with formatting on save enabled and without problems :) .
Since upgrading automatically to 1.4.0, I've noticed some minor new formatting (e.g `{...someRecord, foobar: foobar}` becoming `{...someRecord, foobar}`), which is surprising as I haven't migrated to a new rescript version. Painpoint: potentially having to reformat all my codebase without any "good reason" (quotes, quotes)? Hmm.

But much much more importantly, code like
```res
  @val
  external \"import": (
    @as("../SomeModule.mjs") _,
    unit,
  ) => Js.Promise.t<'a> = "import"
  let make = ReactLazy.make(() =>
    \"import"() |> Js.Promise.then_((module(SomeModule: SomeModuleType)) =>
      Js.Promise.resolve({"default": SomeModule.make})
    )
  )
```
is changed to
```res
 @val
  external import: (
    @as("../SomeModule.mjs") _,
    unit,
  ) => Js.Promise.t<'a> = "import"
  let make = ReactLazy.make(() =>
    import() |> Js.Promise.then_((module(SomeModule: SomeModuleType)) =>
      Js.Promise.resolve({"default": SomeModule.make})
    )
  )
```

and produces syntax errors 
1)
```
 `import` is a reserved keyword. Keywords need to be escaped: \"import"
 ``` 
(for the declaration of import)
 
2)
`I'm not sure what to parse here when looking at "import".` (for the use of import)

(Happy to help if more details needed, beta testing, etc)

cristianoc:
See release notes for the compiler.
import was an experimental syntax never officially supported and has now been removed.

zth:
I don't think this is related to the dedicated import syntax experiment that was removed. I think 2 things are at play here:

1) Formatting. A while back we introduced the ability to format files outside of ReScript projects. Formatting outside of a ReScript project uses the parser directly, which we vendor in the extension. This is because having no ReScript project setup means there's no ReScript binary to use for formatting, so we need to use something else.

The vendored parser is updated quite frequently, and is not tied to a specific ReScript version. It's instead typically a very recent, often unreleased version.

[...]

PR #498, opened 2022-07-17
Fix issue in finding bsc binary.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/496


cristianoc:
A string is a string is a string. But not all strings mean the same thing.

zth:
Lovely!

zth (approved):

PR #499, opened 2022-07-18
Test: Verify MARKETPLACE_TOKEN with vsce

fhammerschmidt:
@ryyppy what is the name of the environment in which the secrets are stored? Is it not `secrets` (i.e. did you create it with https://github.com/rescript-lang/rescript-vscode/settings/secrets/actions/new)?

It currently fails with this error: 
```
error: option '-p, --pat <token>' argument missing
```


zth:
Closing this since I believe this is unnecessary now that we merged the core functionality.

PR #501, opened 2022-07-19
Fix get binary dir path for monorepos

fhammerschmidt:
As already discussed in Discord, I would like to extend this PR with a test runner, to free us from future regressions. Any suggestions?

cc @cknitt please check if the monorepo example is realistic enough.


fhammerschmidt:
Or, you are free to merge this one and I deliver a test PR later.

cristianoc (changes requested):
We need a spec first. Discussed on the forum and agreed by all monorepo users.
Implementation comes later.

server/src/server.ts
```diff
@@ -67,7 +67,7 @@ let send: (msg: p.Message) => void = (_) => {};
 
 let getBinaryDirPath = (projectRootPath: p.DocumentUri) =>
   extensionConfiguration.binaryPath === null
-    ? path.join(projectRootPath, c.nodeModulesBinDir)
+    ? utils.findBinaryFromProjectRoot(projectRootPath)
```
cristianoc, line 70: Looks good.

cristianoc (approved):

PR #502, opened 2022-07-19
Complete JSX props via type

zth:
@cristianoc this sets up variant/polyvariant completion for JSX props. High level review appreciated. 

I think I'm seeing a few patterns starting to emerge that makes me want to go back and refactor/adapt the approach. But I'd like to do one or two more contexts in a crude way just to see if I can get even more information before refactoring. Sounds good?


analysis/src/CompletionBackEnd.ml
```diff
@@ -1362,17 +1362,102 @@ let getOpens ~rawOpens ~package ~env =
   (* Last open takes priority *)
   List.rev resolvedOpens
 
-let processCompletable ~debug ~package ~scope ~env ~pos ~forHover
-    (completable : Completable.t) =
-  let rawOpens = Scope.getRawOpens scope in
-  let opens = getOpens ~rawOpens ~package ~env in
-  let allFiles = FileSet.union package.projectFiles package.dependenciesFiles in
+(* Takes a context and a component path (SomeModule.SomeComponent) and extracts all labels *)
+let getLabelsForComponent ~package ~opens ~allFiles ~pos ~env ~scope
```
zth, line 1366: Function extracted for reuse.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1362,17 +1362,102 @@ let getOpens ~rawOpens ~package ~env =
...
+            _ )
+        when Path.name path = "React.componentLike" ->
+        getFields tObj
+      | _ -> []
+    in
+    typ |> getLabels
+  | None -> []
+
+(* This completes a type expression. Intended to be shared logic that can extract completions from a context, as soon as that the type is identified. *)
+let completeTypeExpr ~env ~package ~debug ~prefix (typeExpr : Types.type_expr) =
```
zth, line 1418: Extracted for reuse.

analysis/src/CompletionFrontEnd.ml
```diff
@@ -27,6 +27,38 @@ let positionToOffset text (line, character) =
     if bol + character <= String.length text then Some (bol + character)
     else None
 
+let getIdentFromExpr exp =
+  match exp.Parsetree.pexp_desc with
+  | Pexp_ident {txt} -> txt
+  | Pexp_construct ({txt}, _) -> txt
+  | Pexp_variant (label, _) -> Lident label
+  | Pexp_field (_, {txt}) -> txt
+  | _ -> Lident ""
```
zth, line 36: I guess this is right, because for anything else we'd like to complete from the entire context.

analysis/src/CompletionFrontEnd.ml
```diff
@@ -40,8 +72,16 @@ type jsxProps = {
   childrenStart: (int * int) option;
 }
 
+(* Checks if an expression is elgible for typed context completion. *)
+let isElgibleForTypedContextCompletion exp =
```
zth, line 76: Maybe this is just an artifact of the current approach being semi-wrong in some way, but I feel I need to have something like this that prevents me from "eating" expressions for type based completion, that shouldn't get completed. 

This essentially means that type based completions only work on one level and not deeper into the tree. I have a few ideas on how we can make this more robust and work in a more generalized way. But, as I said in the PR description, I'm letting that rest for just a little while longer.

analysis/src/Utils.ml
```diff
@@ -71,3 +71,42 @@ let flattenLongIdent ?(jsx = false) ?(cutAtOffset = None) lid =
   in
   let path, _ = loop lid in
   List.rev path
+
+let identifyPexp pexp =
```
zth, line 75: Just a debug util I tend to use. Can remove this later, or move it elsewhere.

analysis/tests/src/expected/TypeContextCompletion_Jsx.res.txt
```diff
@@ -1,31 +1,5581 @@
...
+    "kind": 12,
+    "tags": [],
+    "detail": "someVariant",
+    "documentation": null
+  }, {
+    "label": "Six",
+    "kind": 4,
+    "tags": [],
+    "detail": "Six(option<string>, int)\n\ntype someVariant =\n  | One\n  | Two\n  | Three\n  | Four\n  | Five(int)\n  | Six(option<string>, int)",
+    "documentation": null
```
zth, line 66: There's a bunch of duplication etc I'll need to deal with here. The top variants completed is the new work I'm doing, and the ones on the bottom comes from the general "complete from context". 

Also notice the detail is looking better on the ones completed from the general context. Not sure why, need to look into that.

analysis/src/CompletionFrontEnd.ml
cristianoc: Completion has a mode: field or id etc. If I remember correctly!
You might want to extract the mode.

analysis/src/CompletionFrontEnd.ml
cristianoc: I think it's true that you only want to proceed if an interesting type is found. Where interesting means: actionable for completion. If not, it's as if noting was seen.

That said, sub-expressions override outer expressions, so you don't need to hold back early on if you know something more precise will be found.

analysis/src/Utils.ml
cristianoc: This one is useful. Needed so many times.
Let's put it somewhere accessible.

cristianoc (commented):
Some high level comments.

analysis/src/CompletionFrontEnd.ml
cristianoc:
```ml
  (* Completion context *)
  type completionContext = Type | Value | Module | Field

```

cristianoc: Btw feel free to suggest different names for the concepts.
"Context" is becoming super overloaded.

cristianoc: Probably `idKind` is more fitting.

zth: I'll do a pass and suggest new names for a few things. I agree, context is overloaded, and there's a few similar-but-different concepts that has tripped me up a couple of times. I'll get back to you.

analysis/src/CompletionFrontEnd.ml
zth: I don't want to actually implement it now, but I want the final solution to make this kind of completion simple too:

```rescript
type someVariant = One | Two | Three

let someFunc = (~someParam: someVariant) => ...

someFunc(~someParam={
  if whatever == 123 {
     
//. ^com
})
```
If possible, I want the above to complete to `someVariant`, and I hope the rule could be as simple as "if this is the last expression in the braces body, check what type this block should return. Oh it's someVariant, complete from that".

analysis/src/Utils.ml
zth: Yeah I think I wrote it by hand like 5 times first before I realized I must save it somewhere 😄

analysis/src/CompletionBackEnd.ml
```diff
@@ -1757,53 +1803,33 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
...
+           Maybe that'd be confusing for the user though, if items in scope don't appear.
+           We should at a minimum be able to sort values according to type and the proximity
+           of them in the scope. *)
+        let completionsFromContext =
+          contextPath
+          |> getCompletionsForContextPath ~package ~opens ~rawOpens ~allFiles
+               ~pos ~env ~exact:forHover ~scope
+        in
+        completeTypeExpr ~env ~package ~debug ~prefix typeExpr
+        @ completionsFromContext)
```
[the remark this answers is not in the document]
zth: Sorry, this diff got really messy.

Correct, `completeTypeExpr` is the refactoring and `completionsFromContext` is the change. 

The concatenation is intentional. My reasoning is I need to bring in "general" completions too, otherwise I wouldn't complete for the following in the "typed" scenario:
* Variables you might have in scope that has the value
* Other modules you might want to use something from to get what you need.

Makes sense?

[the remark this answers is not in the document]
cristianoc: If the type is known (and interesting, e.g. a variant), there's no need to complete based on scope unless I'm missing something.
Do you have an example of where this is not the case?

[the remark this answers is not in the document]
zth: We might be talking about two separate things here. Here are two examples of what I mean:

```rescript
type someVariant = One | Two

let someValue = One

let someFunc = (~value: someVariant) => {...

// Here what I want is to use `someValue`. That's a local value I need to bring into the completion, or there's no way for me to access it via completing.
someFunc(~value=
```

And same as above, but with the value in another module/file.
```rescript
// Foo.res
type someVariant = One | Two

let someValue = One

let someFunc = (~value: someVariant) => {...

// Bar.res
// Here what I want is to use `Foo.someValue`. So I need to be able to complete for `Foo`, so I then can complete for `someValue`.
Foo.someFunc(~value=
```

Are you following what I mean?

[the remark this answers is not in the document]
cristianoc: You should not care where it is defined. The type definition gives a list of variants and you use them.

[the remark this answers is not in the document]
cristianoc: There are no general completions in those examples, as neither uses local bindings. They both use the type instead.
Local bindings are not restricted to variants of that type, but will return anything that is in scope.

I interpret that list append as: things from the type + everything in scope.

zth:
Merging this into the other branch so I can continue exploration from there.

PR #504, opened 2022-07-20
Prevent dir from 'eating' the projectRootPath

### LINKED ISSUE #503: Issue in finding bsc binary
Hi. This is a follow-up to rescript-lang/rescript-vscode/issues/496
I've upgraded to v1.4.1 and it seems it still doesn't find my bcc binary, so that it formatted for 10.x instead of 9.1.4

In one of the projects, I managed to get VS Code opening a warning window saying "Can't find Rescript binary in the directory null"

Screenshot 2022-07-19 at 23 39 35

fhammerschmidt:
Fixes #503

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,9 @@ If there's no ReScript build running already in the opened project, the extensio
 
 #### How does it find my ReScript binary?
 
-The extension will look for the existence of a `/node_modules/.bin/bsc` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. If it does not find it, the path can be overwritten with the setting `rescript.settings.binaryPath`.
+The extension will look for the existence of a `/node_modules/.bin/rescript` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level.
+
+To override this lookup process, the path can be configured explicitly using the setting `rescript.settings.binaryPath`.
 
 ### Hide generated files
 
--- a/README.md
+++ b/README.md
@@ -99,6 +99,10 @@ You'll find all ReScript specific settings under the scope `rescript.settings`. 
 
 If there's no ReScript build running already in the opened project, the extension will prompt you and ask if you want to start a build automatically. You can turn off this automatic prompt via the setting `rescript.settings.askToStartBuild`.
 
+#### How does it find my ReScript binary?
+
+The extension will look for the existence of a `/node_modules/.bin/bsc` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. If it does not find it, the path can be overwritten with the setting `rescript.settings.binaryPath`.
+
 ### Hide generated files
 
 You can configure VSCode to collapse the JavaScript files ReScript generates under its source ReScript file. This will "hide" the generated files in the VSCode file explorer, but still leaving them accessible by expanding the source ReScript file they belong to.
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -67,7 +67,7 @@ let send: (msg: p.Message) => void = (_) => {};
 
 let getBinaryDirPath = (projectRootPath: p.DocumentUri) =>
   extensionConfiguration.binaryPath === null
-    ? utils.findBinaryFromProjectRoot(projectRootPath)
+    ? utils.findBinaryDirPathFromProjectRoot(projectRootPath)
     : extensionConfiguration.binaryPath;
 
 let findRescriptBinary = (projectRootPath: p.DocumentUri) =>
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -65,9 +65,28 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
+let findBinaryDirPathFromProjectRoot = (
+  directory: p.DocumentUri // This must be a directory and not a file!
+): null | p.DocumentUri => {
+  let binaryDirPath = path.join(directory, c.nodeModulesBinDir);
+  let binaryPath = path.join(binaryDirPath, c.rescriptBinName);
+
+  if (fs.existsSync(binaryPath)) {
+    return binaryDirPath;
+  }
+
+  let parentDir = path.dirname(directory);
+  if (parentDir === directory) {
+    // reached the top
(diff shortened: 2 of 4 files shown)


server/src/utils.ts
```diff
@@ -39,14 +39,14 @@ export let findProjectRootOfFile = (
 };
 
 export let findBinaryFromProjectRoot = (
-  source: p.DocumentUri
+  projectRootPath: p.DocumentUri // This must be a directory and not a file!
 ): null | p.DocumentUri => {
-  let dir = path.dirname(source);
-  let bscNativeReScriptPath = path.join(dir, c.nodeModulesBinDir);
+  let dir = path.dirname(projectRootPath);
```
cristianoc, line 44: The style mistake is still there: `dir` is defined too early. It's very easy to change this and introduce a bug again.

cristianoc:
I think the current search for the binary needs to be documented.
And added to the changelog too.
So we know what it's supposed to do and  a user knows too.

server/src/utils.ts
```diff
@@ -39,19 +39,25 @@ export let findProjectRootOfFile = (
 };
 
 export let findBinaryFromProjectRoot = (
```
cristianoc, line 41: `findBinaryPath...` right?

cristianoc: Not the actual binary.

server/src/utils.ts
```diff
@@ -39,19 +39,25 @@ export let findProjectRootOfFile = (
...
+  let bscNativeReScriptPath = path.join(
+    projectRootPath,
+    c.nodeModulesBinDir,
+    c.bscBinName
+  );
 
   if (fs.existsSync(bscNativeReScriptPath)) {
-    return bscNativeReScriptPath;
-  } else if (dir === source) {
+    return path.dirname(bscNativeReScriptPath);
```
cristianoc, line 51: This does not look nice.
First create a path, then strip the last element.
Another recipe for future bugs.

fhammerschmidt:
I fixed the style mistakes now.

Furthermore, I successfully tested it on: 
* https://gist.github.com/remitbri/7ca472afe4b64a6af167b20250d58c5c
* A vanilla project created with `rescript init`
* All of the examples in the `analysis/examples` folder
* A couple of yarn workspace and non-workspace projects at work

server/src/utils.ts
cristianoc: Indeed. Looking at the caller:
```ml
let getBinaryDirPath = (projectRootPath: p.DocumentUri) =>
  extensionConfiguration.binaryPath === null
    ? utils.findBinaryFromProjectRoot(projectRootPath)
```

this should be instead:
```ml
let getBinaryDirPath = (projectRootPath: p.DocumentUri) =>
  extensionConfiguration.binaryPath === null
    ? utils.findBinaryDirPathFromProjectRoot(projectRootPath)

```

cristianoc:
Asking for 1 more review just to be sure.

server/src/utils.ts
```diff
@@ -38,20 +38,23 @@ export let findProjectRootOfFile = (
...
+  let binaryPath = path.join(binaryDirPath, c.bscBinName);
 
-  if (fs.existsSync(bscNativeReScriptPath)) {
-    return bscNativeReScriptPath;
-  } else if (dir === source) {
+  if (fs.existsSync(binaryPath)) {
+    return binaryDirPath;
+  }
+
+  let dir = path.dirname(projectRootPath);
```
cristianoc, line 51: `parentDir`

server/src/utils.ts
```diff
@@ -38,20 +38,23 @@ export let findProjectRootOfFile = (
   }
 };
 
-export let findBinaryFromProjectRoot = (
-  source: p.DocumentUri
+export let findBinaryDirPathFromProjectRoot = (
+  projectRootPath: p.DocumentUri // This must be a directory and not a file!
```
cristianoc, line 42: This is called `projectRootPath` but it's not true anymore in the recursive call.

server/src/server.ts
```diff
@@ -67,7 +67,7 @@ let send: (msg: p.Message) => void = (_) => {};
 
 let getBinaryDirPath = (projectRootPath: p.DocumentUri) =>
```
cristianoc, line 68: This is the only called. Feels like the two functions are tightly coupled and should not live in different files.
Too easy to change one without adapting the other.

cristianoc: In fact why do we need 2 functions to begin with.

server/src/utils.ts
```diff
@@ -38,20 +38,23 @@ export let findProjectRootOfFile = (
...
 ): null | p.DocumentUri => {
-  let dir = path.dirname(source);
-  let bscNativeReScriptPath = path.join(dir, c.nodeModulesBinDir);
+  let binaryDirPath = path.join(projectRootPath, c.nodeModulesBinDir);
+  let binaryPath = path.join(binaryDirPath, c.bscBinName);
```
cristianoc, line 45: Is there a special reason for looking for `bsc` and not `rescript`? The latter seems more future-proof.

fhammerschmidt:
> I think the current search for the binary needs to be documented. And added to the changelog too. So we know what it's supposed to do and a user knows too.

Yes, but let's do this in another PR.

server/src/utils.ts
cristianoc: Notice it's explicitly non-backwards-compatible with `bs-platform`. In the changelog.

cristianoc: So `bsc` might in fact find the wrong dir by mistake. If `bs-platform` is around.

cristianoc: This is also why this needs to be documented and put in the release notes.
So it's exactly clear where and how the binary is searched for.

server/src/utils.ts
cristianoc: directory or dir I would say -- more specific

cristianoc:
I think I'm done with the review.
Leaving it to @cknitt for a final look.

@zth a README update still missing before the release.

README.md
```diff
@@ -101,7 +101,7 @@ If there's no ReScript build running already in the opened project, the extensio
 
 #### How does it find my ReScript binary?
 
-The extension will look for the existence of a `/node_modules/.bin/bsc` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. If it does not find it, the path can be overwritten with the setting `rescript.settings.binaryPath`.
+The extension will look for the existence of a `/node_modules/.bin/rescript` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. If it does not find it, the path can be overwritten with the setting `rescript.settings.binaryPath`.
```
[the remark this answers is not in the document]
cristianoc: `If it does not find it, the path can be overwritten with the setting `rescript.settings.binaryPath`.

This is not correct. The setting predates (hence overrides) any searches.

cknitt (approved):
Looks good to me! 👍

PR #505, opened 2022-07-20
Fix commit SHA in artifact name for PR builds

cknitt:
This fixes the commit SHA used for the VSIX artifact name in PR builds.

Now it matches the last commit in the PR (9fe9e91 in this case).


zth (approved):
Good stuff, thank you!

PR #510, opened 2022-07-22
Add: Workspace symbols

### LINKED ISSUE #292: Add navigate to symbols
This plugin does not provide a way to find or navigate to symbols if you know their name.

When I use the go to any bar with `#` or `@` to find a symbol in the whole project or in the current file, there are no symbols provided.

If you point me to the documentation of th VSCode API and the place to do it in this plugin maybe I can implement it myself, but I need some guidance here.

aspeddro:
Close #292


server/src/utils.ts
```diff
@@ -156,6 +156,24 @@ export let runAnalysisAfterSanityCheck = (
   return JSON.parse(stdout.toString());
 };
 
+export let runAnalysis = (args: Array<any>) => {
+  let binaryPath;
+  if (fs.existsSync(c.analysisDevPath)) {
```
cristianoc, line 161: This logic should not be duplicated.
There's a function called `getAnalysisBinaryPath`.

server/src/server.ts
```diff
@@ -501,6 +502,20 @@ function documentSymbol(msg: p.RequestMessage) {
   return response;
 }
 
+function workspaceSymbol(msg: p.RequestMessage) {
+  // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol
+  let result: p.WorkspaceSymbol[] | null = utils.runAnalysis(
```
cristianoc, line 507: Why is this not called like all the other comments? Why does it need its own `runAnalysis` function?
Is there some special difference?

analysis/src/Symbols.ml
```diff
@@ -134,19 +134,42 @@ let command ~path =
...
+                 uri = Uri.toString (Uri.fromPath path);
+                 range = Utils.cmtLocToRange loc;
+               };
+             kind = kindNumber kind;
+           }
+         in
+         symbol)
+
+let workspace ~dir =
+  let open FindFiles in
```
cristianoc, line 153: There are a bunch of places in the codebase with this kind of code.
Wondering if it's similar enough to grant some refactoring.
Or if the differences are too big to gain something.

E.g. the functionality to read `sourceDirectories` could be factored out.

analysis/tests/src/expected/DocumentSymbol.res.txt
```diff
@@ -1,84 +1,67 @@
 DocumentSymbol src/DocumentSymbol.res
-[
-{
+[{
```
cristianoc, line 2: Why is there a change here?

cristianoc:
No tests for the new command yet right?

server/src/utils.ts
aspeddro: This function was client side, add it to `utils.ts`

server/src/server.ts
aspeddro: The `workspace/symbol` request parameters do not have any `textDocument.uri`. The function to perform the analysis requires a `filePath` argument.

analysis/tests/workspace-tests.txt
```diff
@@ -0,0 +1 @@
+workspaceSymbol
```
aspeddro, line 1: The tests for workspace are defined here.

analysis/tests/workspace/expected/workspaceSymbol.txt
```diff
@@ -0,0 +1,1637 @@
+[{
```
aspeddro, line 1: All symbols in `tests/src` dir

server/src/utils.ts
```diff
@@ -129,31 +139,32 @@ export let formatCode = (
...
     return null;
   }
-  let options: childProcess.ExecFileSyncOptions = {
-    cwd: projectRootPath || undefined,
-    maxBuffer: Infinity,
-  };
-  let stdout = childProcess.execFileSync(binaryPath, args, options);
-
-  return JSON.parse(stdout.toString());
+  return runAnalysis(args, undefined, projectRootPath == null ? undefined : projectRootPath)
```
aspeddro, line 167: The second argument is `undefined` so `runAnalysis` will found the `binaryPath`

analysis/tests/workspace/expected/workspaceSymbol.txt
cristianoc: This is going to be noisy when new test files are added. Also this output is big.
How about using a couple of small files and a subdirectory.

analysis/tests/test.sh
```diff
@@ -16,6 +16,17 @@ for file in not_compiled/*.res; do
   fi
 done
 
+# Workspaces Tests are defined in workspace-tests.txt. Each line is name of test to
```
cristianoc, line 19: not sure about "defined"?
isn't that file where the output goes?

cristianoc: Actually let me check -- I think I misunderstood.

server/src/utils.ts
```diff
@@ -139,18 +139,18 @@ export let formatCode = (
   }
 };
 
-export let runAnalysis = (args: Array<any>, binaryPath?: string, cwd?: string) => {
-  let analysisPath = binaryPath != null ? binaryPath : getAnalysisBinaryPath()
+export let runAnalysis = (args: Array<any>, cwd?: string) => {
+  let binaryPath = getAnalysisBinaryPath()
```
cristianoc, line 143: 👍

cristianoc:
@zth not necessarily about this PR, but we should consider beginning to add tests for the TypeScript part of the extension. That's where most breakages have been recently (and where we have no tests).
Instead of going top down, test framework, overhead, unclear utility, i.e. using "best practices". How about doing a quick review of recent breakages and add the least amount of infra required to prevent those from re-occurring. Then build out from there in future.

zth:
> @zth not necessarily about this PR, but we should consider beginning to add tests for the TypeScript part of the extension. That's where most breakages have been recently (and where we have no tests). Instead of going top down, test framework, overhead, unclear utility, i.e. using "best practices". How about doing a quick review of recent breakages and add the least amount of infra required to prevent those from re-occurring. Then build out from there in future.

Sounds good to me. Let's discuss this separately then.

analysis/src/Symbols.ml
zth: Should we add a separate issue to track this @cristianoc ?

server/src/server.ts
```diff
@@ -63,7 +64,7 @@ let projectsFiles: Map<
 let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 
 // will be properly defined later depending on the mode (stdio/node-rpc)
-let send: (msg: p.Message) => void = (_) => {};
+let send: (msg: p.Message) => void = (_) => { };
```
zth, line 67: Maybe revert this formatting change?

analysis/src/Symbols.ml
cristianoc: I think this can be assessed now, and any follow-on work postponed.

analysis/tests/workspace/expected/workspaceSymbol.txt
cristianoc: This is the one thing still holding the PR from being merged.

server/src/server.ts
```diff
@@ -4,14 +4,14 @@ import * as v from "vscode-languageserver";
 import * as rpc from "vscode-jsonrpc/node";
 import * as path from "path";
 import fs from "fs";
-import os from "os";
```
aspeddro, line 7: `os` is not used

aspeddro:
@zth @cristianoc could you disable the CI? It's in an infinite loop

package.json
```diff
@@ -179,7 +188,7 @@
 		"vscode:prepublish": "npm run clean && npm run compile",
 		"compile": "tsc -b",
 		"watch": "tsc -b -w",
-		"postinstall": "cd server && npm i && cd ../client && npm i && cd ../analysis/tests && npm i && cd ../reanalyze/examples/deadcode && npm i && cd ../termination && npm i"
+		"postinstall": "cd server && npm i && cd ../client && npm i && cd ../analysis/reanalyze/examples/deadcode && npm i && cd ../termination && npm i"
```
aspeddro, line 191: Removed `cd analysis/tests && npm i`. No longer has a `package.json` file

server/src/server.ts
aspeddro: Fixed.

aspeddro:
There were the following changes in the tests folder:

- `tests/src/`, `tests/not_compiled`, `tests/package.json` and `tests/bsconfig.json` moved to `tests/document/` folder.

cristianoc:
The last changes slowed down tests by about 3 seconds each clean build+test.
10 features would slow down 30 seconds, so looking into a more lightweight approach

cristianoc:
Let me test something.

cristianoc:
Oops commit history is gone.
Anyhow, it's just about putting tests into `analysis/tests/src/workplace` and change nothing in the structure.

cristianoc:
This way we also avoid to merge conflict any other PR currently ongoing.

cristianoc:
> The last changes slowed down tests by about 3 seconds each clean build+test.
> 10 features would slow down 30 seconds, so looking into a more lightweight approach

Even if that count is inaccurate, we don't want those 3 seconds.
If anything, faster would be better.

analysis/src/Symbols.ml
zth: @aspeddro will you have a look at this? After that, this should be good for merging.

aspeddro:
> Anyhow, it's just about putting tests into `analysis/tests/src/workspace` and change nothing in the structure.

Placing tests in `analysis/tests/src/workspace` will get all symbols in `analysis/tests/src/`. The `workspaceSymbols` command needs to read `bsconfig.json` and get `sources`.

When putting in `tests/workspaces` I isolate the files in `analysis/tests/src`

cristianoc:
> > Anyhow, it's just about putting tests into `analysis/tests/src/workspace` and change nothing in the structure.
> 
> Placing tests in `analysis/tests/src/workspace` will get all symbols in `analysis/tests/src/`. The `workspaceSymbols` command needs to read `bsconfig.json` and get `sources`.
> 
> When putting in `tests/workspaces` I isolate the files in `analysis/tests/src`

Reading bsconfig is always tricky, as there are a bunch of rules, and might change in future.
For example, global rename does not look into bsconfig.

Related question: what about dependencies?

About tests: there's quite a lot of freedom how test commands can be structured, and surely one can pass a folder as parameter or some other restriction to avoid noisy results.

cristianoc:
Making this run on CI: https://github.com/rescript-lang/rescript-vscode/pull/519

PR #513, opened 2022-07-24
Code Lens

zth:
This implements code lens for function definitions. It's only implemented for function definitions because we also have inlay hints, which fill much of the same purpose as code lenses do, but IMO in a better way. Function definitions are however much better in the code lens format, since the type of the function definition tends to be quite verbose, and that fits better in a code lens rather than an inlay hint.

It's hidden behind a setting + marked as experimental.

image

I also intend this PR to be an educational example of how to implement new commands in the analysis bin. Will do fairly elaborate comments on what I've done.


analysis/src/Hint.ml
```diff
@@ -119,4 +119,58 @@ let inlay ~path ~pos ~maxLength ~debug =
...
+  |> List.filter_map (fun (range : Protocol.range) ->
+         match Cmt.fullFromPath ~path with
+         | None -> None
+         | Some full -> (
+           match
+             References.getLocItem ~full
+               ~pos:(range.start.line, range.start.character + 1)
+               ~debug
+           with
+           | Some {locType = Typed (_, typeExpr, _)} ->
```
mattdamon108, line 160: I have zero knowledge about analysis opt-in in the vscode. Please forgive me about maybe nonsense question.

I always want to have a feature of type checking on the fly, I mean `type check as you type`. This line of code hits me. The analysis can check the type? I mean it has an opt-in type checker already?

cristianoc: > I have zero knowledge about analysis opt-in in the vscode. Please forgive me about maybe nonsense question.
> 
> I always want to have a feature of type checking on the fly, I mean `type check as you type`. This line of code hits me. The analysis can check the type? I mean it has an opt-in type checker already?

Quick background:
- When the editor integration starts running, the compiler has already run (there are exceptions for unsaved code, but let's ignore this)
- The compiler artifacts contain type information.

That type information is used by the analysis as-you-type, as long as the type information was already present before you start typing (i.e. last time the file was saved).

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 #### :rocket: New Feature
 
 - Inlay Hints (experimetal). `rescript.settings.inlayHints.enable: true`
+- Code Lenses for functions (experimetal). `rescript.settings.codeLens: true`
```
cristianoc, line 18: Should is say somewhere that it's off by default?

README.md
```diff
@@ -117,6 +117,15 @@ rescript.settings.inlayHints.enable: true
 rescript.settings.inlayHints.maxLength: 25
 ```
 
+### Code Lens (experimental)
+
+This tells the editor to add code lenses to function definitions, showing its full type as an annotation above the definition.
```
cristianoc, line 122: drop "as an annotation"? that word is overloaded

analysis/src/Hint.ml
```diff
@@ -119,4 +119,58 @@ let inlay ~path ~pos ~maxLength ~debug =
...
\ No newline at end of file
+             | None -> None)))
+
+let codeLens ~path ~debug =
+  let lenses = ref [] in
+  let push loc =
+    let range = Utils.cmtLocToRange loc in
+    lenses := range :: !lenses
+  in
+  (* Code lenses are only emitted for functions right now. So look for value bindings that are functions,
```
cristianoc, line 130: What about interface files?

analysis/src/Protocol.ml
```diff
@@ -1,6 +1,8 @@
 type position = {line: int; character: int}
 type range = {start: position; end_: position}
 type markupContent = {kind: string; value: string}
+type command = {title: string; command: string}
```
cristianoc, line 4: Is this a generic command, or only for codelenses?

analysis/src/Shared.ml
```diff
@@ -49,12 +49,12 @@ let declToString ?(recStatus = Types.Trec_not) name t =
 let cacheTypeToString = ref false
 let typeTbl = Hashtbl.create 1
 
-let typeToString (t : Types.type_expr) =
+let typeToString ?(lineWidth = 60) (t : Types.type_expr) =
```
cristianoc, line 52: It's the second time there's `60`.
First: unique source of truth.
Second: do we have slightly redundant printing functions? Why 2?

analysis/tests/src/expected/Dce.res.txt
```diff
@@ -1,3 +1,3 @@
 DCE src/Dce.res
-issues:243
+issues:249
```
cristianoc, line 2: Would you add an issue reminding to solve this?
This change is pure noise w.r.t. the current PR.

analysis/tests/src/expected/Debug.res.txt
```diff
@@ -4,8 +4,9 @@ Dependencies: @rescript/react
...
+Source files: ./src/Auto.res ./src/CodeLens.res ./src/CompletePrioritize1.res ./src/CompletePrioritize2.res ./src/Completion.res ./src/Component.res ./src/Component.resi ./src/CreateInterface.res ./src/Cross.res ./src/Dce.res ./src/Debug.res ./src/Definition.res ./src/DefinitionWithInterface.res ./src/DefinitionWithInterface.resi ./src/Div.res ./src/DocumentSymbol.res ./src/Fragment.res ./src/Highlight.res ./src/Hover.res ./src/InlayHint.res ./src/Jsx.res ./src/Jsx.resi ./src/LongIdentTest.res ./src/Object.res ./src/Patterns.res ./src/RecModules.res ./src/RecordCompletion.res ./src/RecoveryOnProp.res ./src/References.res ./src/ReferencesWithInterface.res ./src/ReferencesWithInterface.resi ./src/Rename.res ./src/RenameWithInterface.res ./src/RenameWithInterface.resi ./src/TableclothMap.ml ./src/TableclothMap.mli ./src/TypeDefinition.res ./src/Xform.res
```
cristianoc, line 7: Same here -- pure noise.

cristianoc: Also, this is testing the debug, which might be overkill.

client/src/extension.ts
```diff
@@ -240,7 +240,10 @@ export function activate(context: ExtensionContext) {
   // language client, and because of that requires a full restart.
   context.subscriptions.push(
     workspace.onDidChangeConfiguration(({ affectsConfiguration }) => {
-      if (affectsConfiguration("rescript.settings.inlayHints")) {
+      if (
+        affectsConfiguration("rescript.settings.inlayHints") ||
+        affectsConfiguration("rescript.settings.codeLens")
```
cristianoc, line 245: How do we know that `codeLens` config change requires a restart and other configurations don't?
Is there an easy rule one can follow to quickly figure that out? Or does one need to know the code of the entire extension plus the internals of the API for extensions, the spec?

server/src/server.ts
```diff
@@ -30,6 +31,7 @@ interface extensionConfiguration {
     enable: boolean;
     maxLength: number | null;
   };
+  codeLens: boolean;
```
cristianoc, line 34: Wondering whether extension config should go to its own file.

cristianoc (approved):
This looks great. Left some minor comments.

I was wondering whether the existence of a feature could slow down the extension: does vscode even know that the feature is turned off?

server/src/server.ts
```diff
@@ -38,6 +40,7 @@ let extensionConfiguration: extensionConfiguration = {
     enable: false,
     maxLength: 25
   },
+  codeLens: false,
```
cristianoc, line 43: If we change the default to true, do we need to change the value in 2 places?
If so, a comment should go here reminding about that.

cristianoc: Or is this going to be quickly overridden anyway, in which case, some comment sating it's a temporary placeholder.

analysis/src/Hint.ml
zth: Not emitted for interface files, they're redundant there imo since the definition in the interface file will be almost identical to what would've been printed in the code lens anyway.

zth: I'll add a comment on why.

analysis/src/Protocol.ml
zth: Generic command from the protocol. I'll add a reference to it in the spec.

analysis/src/Shared.ml
zth: Right, I don't really need the default itself here. Fixing.

analysis/tests/src/expected/Dce.res.txt
zth: Done https://github.com/rescript-lang/rescript-vscode/issues/515

server/src/server.ts
zth: You mean the type definition?

server/src/server.ts
zth: The latter; I added a clarifying comment.

client/src/extension.ts
zth: I added a clarifying comment.

analysis/tests/src/expected/Debug.res.txt
zth: Issue logged.

analysis/src/Shared.ml
zth: Done.

CHANGELOG.md
zth: Done.

zth:
@cristianoc I think I addressed all of your comments. Ready for another round.

zth:
> This looks great. Left some minor comments.
> 
> I was wondering whether the existence of a feature could slow down the extension: does vscode even know that the feature is turned off?

Yes it knows. Whether VSCode does anything from a LS feature or not is controlled by whether the server tells the client that it's capable of receiving requests for that feature or not. If the server doesn't say it supports a feature, the client won't ask for it nor do anything for it. So we should always keep potentially resource intensive features behind settings that are off by default, to ensure the default experience isn't bloated or potentially slow.

cristianoc:
Good to merge!

analysis/src/Commands.ml
```diff
@@ -391,6 +395,9 @@ let test ~path =
             let line_end = 6 in
             print_endline ("Inlay Hint " ^ path ^ " " ^ string_of_int line_start ^ ":" ^ string_of_int line_end);
             inlayhint ~path ~pos:(line_start, line_end) ~maxLength:"25" ~debug:false)
+          | "cle" ->
+            print_endline ("Code Lens " ^ path);
+            codeLens ~path ~debug:false
```
zth, line 400: A "gotcha" that I needed some time to figure out is that these commands that you add to the code can only be 3 chars. I had more than 3 chars and couldn't figure out why it didn't pick them up in the tests.

analysis/src/Hint.ml
```diff
@@ -119,4 +119,60 @@ let inlay ~path ~pos ~maxLength ~debug =
...
+     and use the loc of the value binding itself so we can look up the full function type for our code lens. *)
+  let value_binding (iterator : Ast_iterator.iterator)
```
zth, line 132: This creates an AST iterator targeting _value bindings_, `let whatever = ...`. Since we're only looking for function definitions, we can target value bindings only, since that's the only place they'll appear.

analysis/src/Commands.ml
cristianoc: Feel free to generalise that.

analysis/src/Hint.ml
```diff
@@ -119,4 +119,60 @@ let inlay ~path ~pos ~maxLength ~debug =
...
+      (vb : Parsetree.value_binding) =
+    (match vb with
+    | {
+     pvb_pat = {ppat_desc = Ppat_var _; ppat_loc};
+     pvb_expr = {pexp_desc = Pexp_fun _};
+    } ->
```
zth, line 138: This pattern matches the value binding to target 2 things:
1. `Ppat_var` means it's a variable, like `let whatever`. Assignments can also be destructures `let {something} = ....`, but this ensures we only target variables.
2. `Pexp_fun` means it's a function definition.

`pvb_pat` is the _pattern_, which is essentially the left hand side of `let whatever = something`. `pvb_expr` is the _expression_ being assigned, so the right hand side after `=`. We ensure that it is indeed a variable being assigned `Ppat_var`, and that it's being assigned a function definition `Pexp_fun`.

analysis/src/Hint.ml
```diff
@@ -119,4 +119,60 @@ let inlay ~path ~pos ~maxLength ~debug =
...
+   let parser =
+     Res_driver.parsingEngine.parseImplementation ~forPrinter:false
+   in
+   let {Res_driver.parsetree = structure} = parser ~filename:path in
+   iterator.structure iterator structure |> ignore);
+  !lenses
+  |> List.filter_map (fun (range : Protocol.range) ->
+         match Cmt.fullFromPath ~path with
+         | None -> None
+         | Some full -> (
```
zth, line 156: This here looks up the compiler artifact for `path`, which will be the file we're extracting code lenses from. The compiler artifacts are what holds the type information as the compiler has compiled successfully.

analysis/src/Hint.ml
```diff
@@ -119,4 +119,60 @@ let inlay ~path ~pos ~maxLength ~debug =
...
+           match
+             References.getLocItem ~full
+               ~pos:(range.start.line, range.start.character + 1)
+               ~debug
+           with
+           | Some {locType = Typed (_, typeExpr, _)} ->
```
zth, line 162: We then use a helper to get the full type via the compiler artifact for the target file, using the positions where we found variables being assigned functions. This will lead us to the full, correct type definitions for all functions assigned to variables.

A type at a location can be a bunch of different things - we target `Typed` specifically here, because we know that's the only thing our function definition can be. In `Typed`, `typeExpr` is what we're interested in. `typeExpr` comes from the type checker, and we have helpers that can print those type definitions as strings.

PR #520, opened 2022-07-26
Ws symbols

cristianoc:
rebased on master


cristianoc:
Rebase on master of https://github.com/rescript-lang/rescript-vscode/pull/510

cristianoc:
linux: `30s` until fail
macOS: `40s` until fail
win: `57s` until fail

cristianoc:
On my dev machine `time make -j test` from scratch.
Master: about `12s`.
PR: under `10s`.

server/src/server.ts
```diff
@@ -569,6 +570,20 @@ function documentSymbol(msg: p.RequestMessage) {
   return response;
 }
 
+function workspaceSymbol(msg: p.RequestMessage) {
+  // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol
+  let result: p.WorkspaceSymbol[] | null = utils.runAnalysis(
+    ["workspaceSymbols", process.cwd()],
```
cristianoc, line 576: This does not look right: `process.cwd` is *not* the current project.

CC @zth

zth: Yeah this looks off. I wonder what to put there though, I guess we're looking for the current workspace root?

cristianoc: Current file I guess, if the extension gives it to you? Or whatever relevant info is given by vscode.

zth: The spec doesn't seem to give us anything of value: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceSymbolParams

I guess we might have the workspaces by looking at the opened file contents cache. But that feels a bit tricky as that means it won't work if no file has been opened. Not sure how to proceed here.

cristianoc: If no file open vscode does nothing at all.
Can't expect this feature to be different.

zth: @aspeddro just to be clear - this is working as intended when you run the actual extension, not just the tests? If so, have you tried it on various project setups?

Just curious given `process.cwd`. Maybe we've misunderstood something.

cristianoc: It's not necessarily about project setup, but about where vscode is started from (cwd).

cristianoc: The only possible choice I think is to look at the list of opened files, and get the list of opened projects from there.
Unless one can also have info about the current file.

What do other extensions do in that regard?

aspeddro: We can support `DidChangeWorkspaceFolders` notification and register a cache. ocaml-lsp supports this [notification](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/workspaces.ml#L32-L66).

aspeddro: > I think it can be done/tested separately, then integrated in the PR?

Sure

zth: Yes, that sounds like a great solution!

PR #521, opened 2022-07-26
Fix invocation of node wrapper scripts on windows

fhammerschmidt:


zth:
@fhammerschmidt is this not needed now with the recent merge?

fhammerschmidt:
Yes, it was hacky anyway.

Issue #522, opened 2022-07-26
Doc comment with comments inside is not shown properly on hover.


aspeddro:
```res
let a = Js.Re.exec_
```

Hover `.exec_` show:

```
(Js.Re.t, string) => option<Js.Re.result>
Executes a search on a given string using the given RegExp object. urns Some(Js.Re.result) if a match is found, None otherwise.

res example Match "quick brown" followed by "jumps", ignoring characters in between Remember "brown" and "jumps" Ignore case

re = %re("/quick\s(brown).+?(jumps)/ig") result = Js.Re.exec_(re, "The Quick Brown Fox Jumps Over The Lazy Dog")

egExp.prototype.exec()`]([https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec](vscode-file://vscode-app/opt/visual-studio-code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) MDN.
```

Docstring:

~~~
(**
  Executes a search on a given string using the given RegExp object.
  Returns `Some(Js.Re.result)` if a match is found, `None` otherwise.

  ```res example
  /* Match "quick brown" followed by "jumps", ignoring characters in between
   * Remember "brown" and "jumps"
   * Ignore case
   */

  let re = %re("/quick\s(brown).+?(jumps)/ig")
  let result = Js.Re.exec_(re, "The Quick Brown Fox Jumps Over The Lazy Dog")
  ```

  See
  [`RegExp.prototype.exec()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec)
  on MDN.
*)
external exec_ : t -> string -> result option = "exec" [@@bs.send] [@@bs.return null_to_opt]
~~~

The start of link **[`R** is cut.

~~~
[`RegExp.prototype.exec()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec)
~~~

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/pull/526 for a self-contained example.

cristianoc:
The parser handles the examples properly.
A verbatim print of what's stored in the `ns.doc` attribute shows the problem already.
Something must be happening between parsing and when the typed ast is created in the compiler.

PR #526, opened 2022-07-26
Fix issue where doc comment with nested comments inside is not shown properly on hover.

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/522


cristianoc:
As shown in the tests, the issue is with parsing nested comments.

PR #528, opened 2022-07-27
Add `workspace/didChangeConfiguration`

aspeddro:
#### Change

- We don't need to restart the server when change `rescript.settings`. Just send a notification and update the settings

#### Polish

- Also, no need an extra field in `initializationOptions`. This makes setup easier on other clients like Neovim

```diff
require'lspconfig'.rescriptls.setup = {
  cmd = {
    'node',
    '/home/pedro/Desktop/Projects/rescript-vscode/server/out/server.js',
    '--stdio',
  },
  init_options = {
-    extensionConfiguration = {
-      askToStartBuild = false
-    },
+    askToStartBuild = false
  },
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ You'll find all ReScript specific settings under the scope `rescript.settings`.
 | Prompt to Start Build  | If there's no ReScript build running already in the opened project, the extension will prompt you and ask if you want to start a build automatically. You can turn off this automatic prompt via the setting `rescript.settings.askToStartBuild`.                                                                                                                                                                                                                                    |
 | ReScript Binary Path           | The extension will look for the existence of a `/node_modules/.bin/rescript` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. To override this lookup process, the path can be configured explicitly using the setting `rescript.settings.binaryPath` |
 | Inlay Hints (experimental) | This allows an editor to place annotations inline with text to display type hints. Enable using `rescript.settings.inlayHints.enable: true` |
-| Code Lens (experimental) | This tells the editor to add code lenses to function definitions, showing its full type above the definition. Enable using `rescript.settings.codeLens: true` |
+| Code Lens (experimental) | This tells the editor to add code lenses to function definitions, showing its full type above the definition. Enable using `rescript.settings.codeLens.enable: true` |
 | Autostarting the Code Analyzer | The Code Analyzer needs to be started manually by default. However, you can configure the extension to start the Code Analyzer automatically via the setting `rescript.settings.autoRunCodeAnalysis`.                                                                                                                                                                                                                                                                                |
 
 **Default settings:**
@@ -129,7 +129,7 @@ You'll find all ReScript specific settings under the scope `rescript.settings`.
 "rescript.settings.inlayHints.maxLength": 25
 
 // Enable (experimental) code lens for function definitions.
-"rescript.settings.codeLens": true
+"rescript.settings.codeLens.enable": true
 ```
 
 ## 🚀 Code Analyzer
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -96,9 +96,7 @@ export function activate(context: ExtensionContext) {
       // We'll send the initial configuration in here, but this might be
       // problematic because every consumer of the LS will need to mimic this.
       // We'll leave it like this for now, but might be worth revisiting later on.
-      initializationOptions: {
-        extensionConfiguration: workspace.getConfiguration("rescript.settings"),
-      },
+      initializationOptions: workspace.getConfiguration("rescript.settings"),
     };
 
     const client = new LanguageClient(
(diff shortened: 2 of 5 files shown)


server/src/server.ts
```diff
@@ -399,6 +403,15 @@ function hover(msg: p.RequestMessage) {
...
 
 function inlayHint(msg: p.RequestMessage) {
+
+  if (!initializationOptions.inlayHints?.enable) {
+    return {
+      jsonrpc: c.jsonrpcVersion,
+      id: msg.id,
+      result: null,
+    };
+  }
```
aspeddro, line 413: To clear inlay hints when setting is disabled.

server/src/server.ts
```diff
@@ -426,6 +439,15 @@ function sendInlayHintsRefresh() {
...
 function codeLens(msg: p.RequestMessage) {
+
+  if (!initializationOptions.codeLens?.enable) {
+    return {
+      jsonrpc: c.jsonrpcVersion,
+      id: msg.id,
+      result: null,
+    };
+  }
+
```
aspeddro, line 450: To clear codelens when setting is disabled

README.md
```diff
@@ -107,7 +107,7 @@ You'll find all ReScript specific settings under the scope `rescript.settings`.
...
 | ReScript Binary Path           | The extension will look for the existence of a `/node_modules/.bin/rescript` file and use its directory as the `binaryPath`. If it does not find it at the project root (which is where the nearest `bsconfig.json` resides), it goes up folders in the filesystem recursively until it either finds it (often the case in monorepos) or hits the top level. To override this lookup process, the path can be configured explicitly using the setting `rescript.settings.binaryPath` |
 | Inlay Hints (experimental) | This allows an editor to place annotations inline with text to display type hints. Enable using `rescript.settings.inlayHints.enable: true` |
-| Code Lens (experimental) | This tells the editor to add code lenses to function definitions, showing its full type above the definition. Enable using `rescript.settings.codeLens: true` |
+| Code Lens (experimental) | This tells the editor to add code lenses to function definitions, showing its full type above the definition. Enable using `rescript.settings.codeLens.enable: true` |
```
aspeddro, line 110: The settings in package.json is `rescript.settings.codeLens.enable`

server/src/server.ts
```diff
@@ -24,28 +25,31 @@ import { ChildProcess } from "child_process";
...
+interface initializationOptions {
   askToStartBuild: boolean;
   inlayHints: {
     enable: boolean;
     maxLength: number | null;
   };
-  codeLens: boolean;
+  codeLens: {
+    enable: boolean;
+  };
```
aspeddro, line 36: The config in package.json is `codeLens.enable`

client/src/extension.ts
```diff
@@ -96,9 +96,7 @@ export function activate(context: ExtensionContext) {
       // We'll send the initial configuration in here, but this might be
       // problematic because every consumer of the LS will need to mimic this.
       // We'll leave it like this for now, but might be worth revisiting later on.
-      initializationOptions: {
-        extensionConfiguration: workspace.getConfiguration("rescript.settings"),
-      },
+      initializationOptions: workspace.getConfiguration("rescript.settings"),
```
zth, line 99: While I appreciate wanting to keep thing simple, it's a choice whether we want to break user's configuration now or later. What if we decide we want more things sent to the server from the client at a later stage? Could be as simple as meta data around _what_ client it is, or whatever. If we make this change, we'd need to migrate back to the current solution again to support that, and that'd break all clients. I think the current approach is more future proof, even if it means sending an extra prop.

client/src/extension.ts
```diff
@@ -235,21 +233,14 @@ export function activate(context: ExtensionContext) {
...
-      // the server here. That will typically be any configuration that affects
-      // the capabilities declared by the server, since those cannot be updated
-      // on the fly, and require a full restart with new capabilities set when
-      // initializing.
       if (
-        affectsConfiguration("rescript.settings.inlayHints") ||
-        affectsConfiguration("rescript.settings.codeLens")
+        affectsConfiguration("rescript.settings")
       ) {
-        commands.executeCommand("rescript-vscode.restart_language_server");
```
zth, line 252: Did you test this in an actual editor? Since this affects what capabilities the LS itself is initiated with, I can't imagine turning features _on_ still works after this change, since you'd need to update the capabilities, and to do that you must restart the server.

client/src/extension.ts
```diff
@@ -235,21 +233,14 @@ export function activate(context: ExtensionContext) {
...
+        client.sendNotification(
+          "workspace/didChangeConfiguration",
+          { settings: workspace.getConfiguration("rescript.settings") }).catch((err) => { window.showErrorMessage(String(err)) })
```
zth, line 243: While I do like this touch of notifying the server on config changes, this in combination with removing the pulling breaks all other clients than VSCode, unless they _also_ implement this pushing. I'd rather pull and just have it work on all clients without the extra work needed.

server/src/server.ts
zth: Doing this, and not restarting the server, means the client will continue to send messages to the server for these features. The point of turning the features _off_ via a setting is also that they should be zero cost and not potentially slow down the extension. Even if this is just a null response, it'll still keep sending all of the messages back and forth. Might not hurt performance for inlay hints, but feature X, Y and Z that we'll add in the future might. Better to just turn it off completely via capabilities and a restart, like we're doing now. Then it's truly zero cost.

zth (changes requested):
To be frank, I don't see the point of this. Unless I'm mistaken it also subtly breaks a number of things. 

Maybe you could elaborate a bit more on why this is necessary, and what the trade offs are? It's typically much easier to review a PR that has a clearly stated goal and some reflection on trade offs, especially when it does seemingly "deep" refactors and changes the behavior behind the scenes, like this does.

PR #529, opened 2022-07-27
Call bsc.exe directly, when no binaryPath is set in settings

### LINKED ISSUE #514: Formatting after editting is failed with v1.4.2 but ok with v1.4.1.
rescript[v9.1.4]
vscode extension[v.1.4.2]
system [windows x64]
Failed to format the code with vscode extension of version 1.4.2

### LINKED ISSUE #530: Suggest installing a supported ReScript version when the binary is not found
When depending on `bs-platform@8.4.2`:
Screenshot 2022-07-27 at 21 23 44

- I'm not sure where "the directory null" comes from
- Maybe we could look for `bs-platform` in `node_modules` if `rescript` can't be found and suggest updating?

fhammerschmidt:
This should fix #514 and also will make formatting faster (at least on windows).
Supersedes #521 

I refactored the binary lookup a bit so that:
* we do not need to call existsSync twice on the same file
* the function returns the binary path itself

When no binaryPath is set, the extension will lookup and use the `bsc.exe` now (again).

Also the message when no binary has been found could have shown null as path. I simplified it for that case.

Edit: Fixes #530 now as well.

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -6,12 +6,8 @@ let platformDir =
 // See https://microsoft.github.io/language-server-protocol/specification Abstract Message
 // version is fixed to 2.0
 export let jsonrpcVersion = "2.0";
-export let bscNativeReScriptPartialPath = path.join(
-  "node_modules",
-  "rescript",
-  platformDir,
-  "bsc.exe"
-);
+export let platformPath = path.join("node_modules", "rescript", platformDir);
+export let bscNativeReScriptPartialPath = path.join(platformPath, "bsc.exe");
 
 export let analysisDevPath = path.join(
   path.dirname(__dirname),
@@ -29,6 +25,7 @@ export let analysisProdPath = path.join(
 export let rescriptBinName = "rescript";
 
 export let bscBinName = "bsc";
+export let bscExeName = "bsc.exe";
 
 export let nodeModulesBinDir = path.join("node_modules", ".bin");
 
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -6,8 +6,13 @@ let platformDir =
 // See https://microsoft.github.io/language-server-protocol/specification Abstract Message
 // version is fixed to 2.0
 export let jsonrpcVersion = "2.0";
-export let platformPath = path.join("node_modules", "rescript", platformDir);
-export let bscNativeReScriptPartialPath = path.join(platformPath, "bsc.exe");
+export let platformPath = path.join("rescript", platformDir);
+export let nodeModulesPlatformPath = path.join("node_modules", platformPath);
+export let bscExeName = "bsc.exe";
+export let bscNativeReScriptPartialPath = path.join(
+  nodeModulesPlatformPath,
+  bscExeName
+);
 
 export let analysisDevPath = path.join(
   path.dirname(__dirname),
@@ -25,7 +30,6 @@ export let analysisProdPath = path.join(
 export let rescriptBinName = "rescript";
 
 export let bscBinName = "bsc";
-export let bscExeName = "bsc.exe";
 
 export let nodeModulesBinDir = path.join("node_modules", ".bin");
 
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -79,14 +79,16 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
-let findBinaryDirPathFromProjectRoot = (
-  directory: p.DocumentUri // This must be a directory and not a file!
+let findBinaryFromProjectDirRec = (
+  directory: p.DocumentUri, // This must be a directory and not a file!
+  targetDir: p.DocumentUri,
+  binaryName: p.DocumentUri
 ): null | p.DocumentUri => {
-  let binaryDirPath = path.join(directory, c.nodeModulesBinDir);
-  let binaryPath = path.join(binaryDirPath, c.rescriptBinName);
+  let binaryDirPath = path.join(directory, targetDir);
+  let binaryPath = path.join(binaryDirPath, binaryName);
 
   if (fs.existsSync(binaryPath)) {
-    return binaryDirPath;
+    return binaryPath;
   }
 
   let parentDir = path.dirname(directory);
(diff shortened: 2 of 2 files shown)


server/src/server.ts
```diff
@@ -79,14 +79,16 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
-let findBinaryDirPathFromProjectRoot = (
-  directory: p.DocumentUri // This must be a directory and not a file!
+let findBinaryFromProjectDirRec = (
```
cristianoc, line 82: I find the logic inside this function difficult to follow, so error prone to any changes.
Looks like the caller has all the context, and this function does not stand on its own legs.
(This is only about style, not functionality).

server/src/server.ts
```diff
@@ -95,19 +97,22 @@ let findBinaryDirPathFromProjectRoot = (
...
-  extensionConfiguration.binaryPath == null
-    ? findBinaryDirPathFromProjectRoot(projectRootPath)
-    : extensionConfiguration.binaryPath;
-
 let findRescriptBinary = (projectRootPath: p.DocumentUri) =>
-  utils.findRescriptBinary(getBinaryDirPath(projectRootPath));
+  extensionConfiguration.binaryPath == null
+    ? findBinaryFromProjectDirRec(
+        projectRootPath,
+        c.nodeModulesBinDir,
```
cristianoc, line 107: This, and the next parameter, look very specific.
That's the kind of thing I would see directly inside `findBinaryFromProjectDirRec`, not a parameter if only we could reduce the number of callers to `findBinaryFromProjectDirRec` to be only one.
See below.

server/src/server.ts
```diff
@@ -95,19 +97,22 @@ let findBinaryDirPathFromProjectRoot = (
...
+        c.rescriptBinName
+      )
+    : utils.findRescriptBinary(extensionConfiguration.binaryPath);
 
 let findBscBinary = (projectRootPath: p.DocumentUri) =>
```
cristianoc, line 112: I think can just assume that the bsc binary is right next to the rescript binary.
As if it isn't, there's a big problem no matter what.
So this could be reflected in the logic: find the path to rescript, and adapt it to a path to bsc instead.
This would remove the additional call to `findBinaryFromProjectDirRec`

cristianoc (commented):
Added comments on how to make this simpler and more future-proof.

server/src/server.ts
fhammerschmidt: With this PR, in a simple project made with `rescript init`, the paths would default to
`projectRoot/node_modules/.bsc/rescript` and
`projectRoot/node_modules/rescript/{platformDir}/bsc.exe` 

So they are not in the same folder. Except you mean `rescript.exe` of course.

cristianoc: Let's avoid calling the 2 commands in different ways.
We have enough bugs already.

fhammerschmidt: I don't know what you mean? I should find the path to rescript.exe or rescript (the one that we use now)? I had to look for the actual `bsc.exe` and call it directly, because it won't work on Windows otherwise.

Or do you mean, if you have a rescript binary in `node_modules/.bsc` you probably have the `node_modules/rescript/{platformDir}/bsc.exe`?

cristianoc: I was for a second hoping one could call `rescript.exe` but the `rescript` node script actually does useful work, so I guess we'll have to call that.
OK so let me revise:
- use `node_modules/.bin/rescript` as source of truth to determine where ReScript is installed
- once that is found, assume, as an implementation detail that can be changed in future that the executable is in `node_modules/rescript/{platformDir}/bsc.exe` so just use that directly by massaging the path to `rescript` already found.

So a single, documented resolution logic for `rescript`, plus a little accessory function for `bsc`.

cristianoc: If we want to decouple the logic for finding `bsc.exe` in future, I think we can make `rescript` provide the path to `bsc.exe` as a specific command-line option, and use that once on project initialisation.

fhammerschmidt: > So a single, documented resolution logic for `rescript`, plus a little accessory function for `bsc`.
I think updated the PR to reflect what you mean.

server/src/constants.ts
```diff
@@ -6,11 +6,12 @@ let platformDir =
 // See https://microsoft.github.io/language-server-protocol/specification Abstract Message
 // version is fixed to 2.0
 export let jsonrpcVersion = "2.0";
+export let platformPath = path.join("rescript", platformDir);
```
cristianoc, line 9: TODO: check that `platformDir` computation is in sync with the compiler.

server/src/server.ts
```diff
@@ -79,14 +79,14 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
-let findBinaryDirPathFromProjectRoot = (
+let findBinaryPathFromProjectRoot = (
```
cristianoc, line 82: Add comment?

server/src/server.ts
```diff
@@ -95,19 +95,26 @@ let findBinaryDirPathFromProjectRoot = (
...
+  extensionConfiguration.binaryPath == null
+    ? findBinaryPathFromProjectRoot(projectRootPath)
+    : utils.findRescriptBinary(extensionConfiguration.binaryPath);
+
+let findBscBinary = (projectRootPath: p.DocumentUri) => {
+  let rescriptBinaryPath = findRescriptBinary(projectRootPath);
+  if (rescriptBinaryPath !== null) {
+    return path.join(
+      path.dirname(rescriptBinaryPath),
+      "..",
```
cristianoc, line 111: Is this OK for windows, symbolic links etc? Or should one manipulate the path instead?

server/src/server.ts
```diff
@@ -312,9 +319,10 @@ let openedFile = (fileUri: string, fileContent: string) => {
           method: "window/showMessage",
           params: {
             type: p.MessageType.Error,
-            message: `Can't find ReScript binary in the directory ${getBinaryDirPath(
-              projectRootPath
-            )}`,
+            message:
+              extensionConfiguration.binaryPath == null
+                ? "Can't find ReScript binary"
```
cristianoc, line 324: Starting from where?

cristianoc (commented):
This looks great thanks.
Just need confirmation from Windows users that this works.

server/src/constants.ts
fhammerschmidt: Not quite, in rescript-vscode the `platformDir` also allows for "linuxarm64", the compiler does not:
https://github.com/rescript-lang/rescript-compiler/blob/master/bsc#L6-L12
https://github.com/rescript-lang/rescript-compiler/blob/master/scripts/bin_path.js#L8-L11

cristianoc: In that case the compiler is a little behind and could be updated.

server/src/server.ts
fhammerschmidt: `..` definitely works on windows

server/src/constants.ts
fhammerschmidt: Or is the logic in rescript-vscode correct? The compiler has no directories for `win32arm64` or `linuxarm64`, and neither do the get built in CI?

cristianoc: It would be kind of future-proof, if one wants to build from source.
But does not matter much either way, as this is rarely done I guess.

cristianoc (approved):
Looks good to me.

cristianoc:
I think the README already aligns with this process to find the executable, and requires no change.

cristianoc:
Merge away any time.

PR #532, opened 2022-07-29
Code action: convert type to module

### LINKED ISSUE #430: Feature request: Code action to convert type to module
It's pretty common in ReScript that you want to add some method to a type and thus want to restructure it into its own submodule.

E.g.
```rescript
type readState = New | Unread | Read
```

to 
```
module ReadState = {
  type t = New | Unread | Read
}
```

I am not sure though, if we should deliberately not allow this action on types named `t` (which are supposedly already in their own module).

aspeddro:
Add basic support to convert type to module.

- Support types inside module?

Current state:

- Does not support types inside module
- Does not support recursive types.

**TODO**:

- [ ] Update changelog

Demo:

https://user-images.githubusercontent.com/16160544/181861064-e2754302-b65b-498c-be15-dd90665aaf2e.mp4

Close #430


zth (commented):
Awesome! 😄 I'll let @cristianoc look at the OCaml parts more in depth. I noticed there's a few refactors - might be good to add a comment for each of those and detail what they do/why.

analysis/src/Commands.ml
```diff
@@ -175,73 +175,9 @@ let references ~path ~pos ~debug =
 
 let rename ~path ~pos ~newName ~debug =
   let result =
-    match Cmt.fullFromPath ~path with
+    match Rename.command ~path ~pos ~newName ~debug with
+    | Some workspaceEdit -> workspaceEdit |> Protocol.stringifyWorkspaceEdit
     | None -> Protocol.null
```
aspeddro, line 180: Refactored to `Rename.command` because some code actions need renaming.

analysis/src/Xform.ml
```diff
@@ -110,9 +110,20 @@ module IfThenElse = struct
     | Some newExpr ->
       let range = rangeOfLoc newExpr.pexp_loc in
       let newText = printExpr ~range newExpr in
+      let uri = Uri.fromPath path |> Uri.toString in
```
aspeddro, line 113: To return `uri` scheme.

analysis/src/Xform.ml
```diff
@@ -249,14 +271,133 @@ module AddTypeAnnotation = struct
...
+          ~loc:typ.ptype_loc ~attrs:typ.ptype_attributes
+          ~params:typ.ptype_params ~cstrs:typ.ptype_cstrs ~kind:typ.ptype_kind
+          ~priv:typ.ptype_private
+      | Some manifest ->
+        Ast_helper.Type.mk
+          {txt; loc = typ.ptype_name.loc}
+          ~loc:typ.ptype_loc ~attrs:typ.ptype_attributes
+          ~params:typ.ptype_params ~cstrs:typ.ptype_cstrs ~kind:typ.ptype_kind
+          ~priv:typ.ptype_private ~manifest
+    in
```
aspeddro, line 310: This function is to change the name of type declaration. Is there another way to do this?

`typ.ptype_manifest` is a option, but `Ast_helper.Type.mk` `~manifest` not expected an option

cristianoc:
Haven't looked at the ocaml code that -- can do that soon.
I was first wondering about the big picture, perhaps @zth can help a bit there.
Trying to gauge what are going to be the actions of this kind supported in future, and how they play well with each other.
If we're just experiments and are happy to add/remove these kinds of things freely, then we can just go ahead, mark as experimental etc.

The questions that immediately come to mind are: is this too narrow a use case? Or is it general enough?
If we have 8 of these little things, how noisy is the UI going to get when lots of pieces of code trigger several actions etc.

So my question at the moment is: should we worry about these aspects already? Or just experiment first with a few features see how they feel and how they are used, knowing we might revise perhaps even heavily later?

cristianoc:
Also, should we have an inverse action?

cristianoc:
Some usage questions: this does not seem to trigger an action:

```rescript
type rec tree = Node({name: string, leaves: leaves}) | Empty
and leaves = {number: int, entries: array<tree>}
```

cristianoc:
And a bigger question: the most time consuming part is going around fix all the references to the type.
Should an attempt be made to automate this? As currently we automate a simple copy-paste.

cristianoc:
Notice we already have renaming working well across files, so fixing references might not be super difficult. Perhaps. One would have to try.

cristianoc:
Just as a simple test.

Before:
```rescript
type myType = This | That

let fun1 = (x: myType) => x

let fun2 = b => b ? This : That
```

Ohhh I was surprised to see that half of this is done automatically already:
```rescript
module MyType = {
  type t = This | That
}

let fun1 = (x: MyType.t) => x

let fun2 = b => b ? This : That
```

What's left is the final step:


```rescript
module MyType = {
  type t = This | That
}

let fun1 = (x: MyType.t) => x

let fun2 = b => b ? MyType.This : That
```

And one might or might not want to try to automate the final step.

cristianoc:
Naming: should this be called "move type definition into its own module"?

aspeddro:
Now support variant/record/object

```res
type state = New | Unread | Read
type refState = state
type person = {"age": int, "name": string}
type user = {
  name: string,
  age: int,
}
and response = Yes | No

type myType = This | That

let fun1 = (x: myType) => x
let fun2 = b => b ? This : That
let fun3 = b => b ? {name: "Lhs", age: 2} : {name: "Rhs", age: 3}
let fun4 = b => b ? Yes : No
let me: person = {
  "age": 5,
  "name": "Big ReScript",
}
```

```res
module State = {
  type t = New | Unread | Read
}
module RefState = {
  type t = State.t
}
module Person = {
  type t = {"age": int, "name": string}
}
module User = {
  type t = {
    name: string,
    age: int,
  }
  and response = Yes | No
}
module MyType = {
  type t = This | That
}

let fun1 = (x: MyType.t) => x
let fun2 = b => b ? MyType.This : That
let fun3 = b => b ? {User.name: "Lhs", age: 2} : {User.name: "Rhs", age: 3}
let fun4 = b => b ? User.Yes : No
let me: Person.t = {
  "age": 5,
  "name": "Big ReScript",
}
```

analysis/src/Xform.ml
```diff
@@ -390,11 +390,13 @@ module TypeToModule = struct
                           match documentChange with
                           | TextDocumentEdit textEdit ->
                             let textDocument = textEdit.textDocument in
-                            (* Ignores first edit as it is within range of code action *)
                             let edits =
-                              match textEdit.edits with
-                              | _ :: rest -> rest
-                              | _ -> []
+                              if textEdit.textDocument.uri = uri then
+                                (* Ignores first edit as it is within range of code action *)
```
cristianoc, line 388: This is pre-existing, but not sure I understand this comment.

cristianoc:
Looks like this is making steady progress.
Keep on going, and feel free to ping when more feedback is required.

zth:
I haven't forgotten about the high level questions @cristianoc , will get back with them soon.

zth:
Here are a few thoughts from me:
- I'm not worried about having too many code actions at this point. That's primarily because I think the way code actions are used is more about learning that certain actions exist in certain contexts, and then using the ones who fit the workflow you personally like. This might be a bit different if we end up having _a lot_ of code actions, and I don't think we should have a bunch of borderline useless actions just for having actions. But at the point we're at now, with code actions in the extension being in its infancy, I think we can afford to experiment some. Even if that means removing things eventually.
- For this particular code action, I'd like to ask ourselves this - is this a code pattern we'd recommend (modules with a type t to encapsulate related types/functions etc)? Personally, I'm inclined to say yes. `t` is a pretty strong convention, and also what pipe autocompletion is based on. If this is a pattern we recommend, making it as easy as possible to work with is a goal imo. And for this particular pattern, one of the annoying things of moving a type to its own module is the refactoring part, which this action goes some (all?) of the way to solve.

Thoughts?

cristianoc:
Agreed.
Just one small comment. Recommendation sounds a bit strong. In that you don't want a bureaucratic one module per type.
But, there are many cases where it makes sense.
So perhaps a bit more nuanced.

zth:
> Agreed. Just one small comment. Recommendation sounds a bit strong. In that you don't want a bureaucratic one module per type. But, there are many cases where it makes sense. So perhaps a bit more nuanced.

Yes, I agree. And I only really mean using t as the main type of a module. Anyways, with that said I think going ahead with this would be nice. Again, we can always call it experimental to start with.

cristianoc:
Great. Let's move on with this.
Seems pretty much ready already.

aspeddro:
It's ready for review.

analysis/src/Xform.ml
cristianoc: Forgot why this does not just the string it starts with?

analysis/src/Xform.ml
```diff
@@ -249,14 +271,144 @@ module AddTypeAnnotation = struct
...
+        in
+        result := Some (Ast_helper.Str.mk ~loc mod_expr, references, modName);
+        Ast_iterator.default_iterator.structure_item iterator si
+      | _ -> ()
+    in
+    {Ast_iterator.default_iterator with structure_item}
+
+  let xform ~path ~pos ~codeActions ~printStructureItem structure ~debug =
+    let result = ref None in
+    let newTypeName = "t" in
```
cristianoc, line 354: Feels like this logic should be factored out into its own easy to modify little local module.

analysis/src/Xform.ml
```diff
@@ -249,14 +271,144 @@ module AddTypeAnnotation = struct
...
+    let iterator = mkIterator ~pos ~result ~newTypeName in
+    iterator.structure iterator structure;
+    match !result with
+    | None -> ()
+    | Some (newStructureItem, references, modName) ->
+      let range = rangeOfLoc newStructureItem.pstr_loc in
+      let newText = printStructureItem ~range newStructureItem in
+      let uri = Uri.fromPath path |> Uri.toString in
```
cristianoc, line 362: Another instance of this.
If this does something useful, it should probably be moved into a standalone function inside Uri.

analysis/src/Xform.ml
```diff
@@ -310,5 +462,7 @@ let extractCodeActions ~path ~pos ~currentFile ~debug =
     AddTypeAnnotation.xform ~path ~pos ~full ~structure ~codeActions ~debug;
     IfThenElse.xform ~pos ~codeActions ~printExpr ~path structure;
     AddBracesToFn.xform ~pos ~codeActions ~path ~printStructureItem structure;
+    TypeToModule.xform ~path ~pos ~codeActions ~printStructureItem structure
```
cristianoc, line 465: We have quite a few of these.
Need a little sanity check:
- that they don't do repeat work in a way that hurts perf (iterating several times is probably ok while copying would not)
- that no work is done unless the action fires

This is just about reviewing the code and check that things are OK.

cristianoc (approved):
This looks great.
Left some comments, nothing substantial.
There's a "ready for review" comment deleted so not sure if more changes are planned.

analysis/src/Xform.ml
aspeddro: Was breaking code actions in neovim client because it was not returning uri scheme from LSP spec.

Example: Add type annotation
image

```json
[{"title": "Add type annotation", "kind": "refactor.rewrite", "edit": {"documentChanges": [{
  "textDocument": {
  "version": null,
  "uri": "/home/pedro/Desktop/learning-rescript/src/intro/Codeaction.res"
  },
  "edits": [{
  "range": {"start": {"line": 20, "character": 5}, "end": {"line": 20, "character": 5}},
  "newText": ": int"
  }]
  }]}}]
```

In vscode it works without problem

cristianoc: Sorry what I meant to ask is: what does this composition of functions do?
Let's put it into a single function with a name suggesting what it does, and use that function instead. So it's less likely that the same issue will pop up somewhere else by copy paste and change in future.

aspeddro:
It will take some time to get ready. There are more cases I need to check.

PR #534, opened 2022-07-30
Remove ".enable" from rescript.settings.codeLens in package.json

Minnozz:
The setting is called `rescript.settings.codeLens` instead of `rescript.settings.codeLens.enable`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
                     ],
                     "minimum": 0
 				},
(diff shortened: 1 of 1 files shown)


Minnozz:
Turns out I had it the wrong way around the first time, so I changed the PR.

zth:
Thank you!

PR #540, opened 2022-08-01
Inlay Hints: better support for destructuring

### LINKED ISSUE #536: Inlay hint incorrectly shows type of first binding of destructured record for entire record
Pre-release version `1.5.11`.

```res
type foo = {
  a: int,
  b: bool,
}

let bar = () => ({a: 0, b: true}, "123")

let ({a, b}, _) = bar()
```

inlay-hint-bug

To me, this incorrectly suggests `{a, b}` has type `int`.

aspeddro:
Fix #536 

image

### DIFF
--- a/analysis/src/Commands.ml
+++ b/analysis/src/Commands.ml
@@ -395,7 +395,7 @@ let test ~path =
           | "dia" -> diagnosticSyntax ~path
           | "hin" ->
             let line_start = 0 in
-            let line_end = 6 in
+            let line_end = 34 in
             print_endline
               ("Inlay Hint " ^ path ^ " " ^ string_of_int line_start ^ ":"
              ^ string_of_int line_end);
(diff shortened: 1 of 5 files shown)


analysis/src/Commands.ml
```diff
@@ -395,7 +395,7 @@ let test ~path =
           | "dia" -> diagnosticSyntax ~path
           | "hin" ->
             let line_start = 0 in
-            let line_end = 6 in
+            let line_end = 34 in
```
cristianoc, line 398: Just noticed this now -- would you add a comment about the line number hardcoded in the test?

cristianoc (approved):
Thanks for improving this.
Would you rebase on master to address merge conflicts. Then we can merge.

analysis/src/Commands.ml
aspeddro: Done

PR #539, opened 2022-08-01
Fix server crashers when open file is removed

aspeddro:
Fix #538

PR #543, opened 2022-08-02
Checking if a file is a res file

zth:
This is a continuation of https://github.com/rescript-lang/rescript-vscode/pull/539

### DIFF
--- a/analysis/src/Hint.ml
+++ b/analysis/src/Hint.ml
@@ -82,7 +82,7 @@ let inlay ~path ~pos ~maxLength ~debug =
     Ast_iterator.default_iterator.value_binding iterator vb
   in
   let iterator = {Ast_iterator.default_iterator with value_binding} in
-  (if Files.isResFile path then
+  (if Files.classifyFile path = Res then
    let parser =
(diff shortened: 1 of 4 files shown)


analysis/src/Hint.ml
```diff
@@ -82,7 +82,7 @@ let inlay ~path ~pos ~maxLength ~debug =
     Ast_iterator.default_iterator.value_binding iterator vb
   in
   let iterator = {Ast_iterator.default_iterator with value_binding} in
-  (if Files.exists path && Filename.check_suffix path ".res" then
+  (if Files.isResFile path then
```
cristianoc, line 85: I'm not sure that this does it.
The temptation could be `else ...` for `.resi` files which would leak to a crash.
Any other ideas?

zth: Check the latest commit, what do you think about that? Slightly harder at to accidentally do the wrong thing at least I think, but still not fool proof.

cristianoc (approved):
Great?
"classifySourceFile"

zth:
Fixed the name! Merging now.

Issue #537, opened 2022-08-01
Something went wrong parsing the json output of reanalyze: 'SyntaxError: Unexpected token W in JSON at position 4'


Minnozz:
```
Something went wrong parsing the json output of reanalyze: 'SyntaxError: Unexpected token W in JSON at position 4'
```
I can't find the output that causes this error in the logs.

cristianoc:
Two repros:
1) repro the issue
2) repro of the insufficient error message (should show the command being run)

It's the same repro needed but 2 aspects to it.

zth:
We've talked about this before - we should log to the extension console what the json that broke is. Should be easy enough.

cristianoc:
Can you try again from the latest beta extension from the marketplace? E.g. 1.5.17
It should give a better error helping reproducing the issue.

Minnozz:
I have updated to 1.5.17 and will see if it happens again.

zth:
Closing this meanwhile, feel free to reopen.

PR #544, opened 2022-08-02
More elaborate error message for users on reanalyze json error

zth:
Follow up for: https://github.com/rescript-lang/rescript-vscode/issues/537

This logs a more elaborate error for the user, including instructions on how to report the error, when reanalyze json is invalid for some reason. 

image

image

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -70,6 +70,11 @@ let client: LanguageClient;
 // });
 
 export function activate(context: ExtensionContext) {
+  let outputChannel = window.createOutputChannel(
+    "ReScript Language Server",
+    "rescript"
+  );
+
   function createLanguageClient() {
     // The server is implemented in node
     let serverModule = context.asAbsolutePath(
@@ -99,6 +104,7 @@ export function activate(context: ExtensionContext) {
       initializationOptions: {
         extensionConfiguration: workspace.getConfiguration("rescript.settings"),
       },
+      outputChannel,
     };
 
     const client = new LanguageClient(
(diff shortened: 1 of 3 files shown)


client/src/extension.ts
```diff
@@ -70,6 +70,11 @@ let client: LanguageClient;
 // });
 
 export function activate(context: ExtensionContext) {
+  let outputChannel = window.createOutputChannel(
+    "ReScript Language Server",
+    "rescript"
+  );
+
```
zth, line 77: Creating and passing around an explicit output channel lets us have a good place to log things to, that's also easy to reveal programatically.

cristianoc (approved):
This is great!
In addition, I was wondering whether we could issue a reproducible command. As that's required to produce the JSON in the first place.
That I guess would amount to pasting the command used to run the analysis.

Or even a few words on how to report an issue (which requires a repro project together wit the command).

zth:
image
What about this?

cristianoc:
This is great!
Ready to merge?

zth:
Yup!

Minnozz:
The error itself when trying to parse the JSON is no longer logged, is that intentional?

zth:
Yes, it's intentional. But it can be added again if it's considered helpful. I didn't think a random json error message would help anyone though as it's the accompanying json one is interested in to fix the issue, and the error itself can be reproduced by trying to parse the faulty json.

cristianoc:
How about an analogous one when there's an issue parsing the compiler log, as it's much more likely to happen and/or regress.

zth:
@cristianoc yes, this is a good idea. I can do that once I'm out of the woods on the current things I'm working on. Adding a new ticket to make sure it isn't lost.

PR #546, opened 2022-08-04
Fix isNamespaced check when namespace is string

fhammerschmidt:
Apparently, `Json.bool` returns `None` when the value is there but not a boolean.


cristianoc:
Good catch.

Issue #252, opened 2021-05-24
Signature help on function applications


amiralies:
https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp

cristianoc:
What's the idea -- different than hover?

amiralies:
It automatically opens when typing function applications (via `(` for example) and also users can config a hotkey to open the signature help while writing.
Another feature i see in ts extension is it highlights current parameter you're applying in signature help box.

amiralies:
https://user-images.githubusercontent.com/13261088/119499113-9f873f00-bd7b-11eb-9750-483fd7eaf893.mp4

This is an exmaple from ts extension

cristianoc:
How's that triggered in the example? By `(` ?
And any ideas of how the mechanism is different from normal auto-completion?

cristianoc:
This feature is quite compelling.

cristianoc:
This looks roughly like:
- `(` triggers signature help
- `,` re-triggers (for the n+1-th argument)

Needs a list of labels (the function parameters), and an index (of the current argument).

So:
```rescript
let add = (x,y) => x+y
```

then
`foo(3,`

would need to give:
- `[int, int]` (the list of parameters, perhaps named with conventional names, perhaps one can retrieve "x" and "y" from the ast)
- 1 (index of the second parameter)

cristianoc:
As usual some care is required about context:
`foo(3, (2,` should not trigger anything as it's about an inner tuple, not an argument to `foo`.

cristianoc:
For perf reasons, context scanning should happen *before* doing any other work that scans project structure etc. As context scanning is pure string parsing.

cristianoc:
In fact, the same perf comment apply to completion: https://github.com/rescript-lang/rescript-vscode/pull/255

oblitum:
The server publishes the characters that trigger this in _Server Capability_ `SignatureHelpOptions.triggerCharacters`.

oblitum:
This is must have, most servers provide this. Just having bare function names completed is a bit lacking. Also, a good chunk of servers also provide snippet expansion on top, meaning, not only is signatureHelp available, you can also expand the function call with all arguments filled with placeholders you can jump to.

zth:
Unless @cristianoc disagrees, this sounds like a good first issue for a contributor with some OCaml skills. Essentially, we should have all we need in the analysis bin now that we've integrated the full parser.

PR #547, opened 2022-08-06
Signature help

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/252

This implements signature help for function applications. Signature help is an "interactive" help hint you get when calling a function, that tells you what parameters the function has, as well as what parameter you're currently filling in, if any. Essentially a hover, but with the current parameter you're working on highlighted.

image

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp


zth:
It's ready for a general review, although I have a couple of more cases to handle here in larger functions.

zth:
After some testing, this has some ways to go for larger function bodies. I had another idea - instead of printing a dummy function, extracting types, and calculating offsets that way, what if I simply take the label (type string) that will be the signature help, parse that type definition with the parser again, and let the parser tell me all of the offsets. 

I think that should work, and also future proof/vastly simplify things.

cristianoc:
Sounds like a good thing to try.

analysis/src/Protocol.ml
```diff
@@ -170,6 +187,44 @@ let stringifyCodeLens (codeLens : codeLens) =
...
+  }|}
+    (Json.escape signatureInformation.label)
+    (signatureInformation.parameters
+    |> List.map stringifyParameterInformation
+    |> array)
+    (match signatureInformation.documentation with
+    | None -> ""
+    | Some docs ->
+      Printf.sprintf ",\n    \"documentation\": %s"
+        (stringifyMarkupContent docs))
```
zth, line 211: Poor mans optional prop stringification.

zth:
@cristianoc ready for your eyes.

analysis/src/SignatureHelp.ml
```diff
@@ -0,0 +1,213 @@
+type cursorAtArg = Unlabelled of int | Labelled of string
+
+let signatureHelp ~path ~pos ~currentFile ~debug =
+  let posBeforeCursor = (fst pos, max 0 (snd pos - 1)) in
```
cristianoc, line 4: This looks very specific -- copy paste from somewhere else right. Worth extracting into a function?

analysis/src/SignatureHelp.ml
```diff
@@ -0,0 +1,213 @@
...
+  let iterator = {Ast_iterator.default_iterator with expr} in
+  let parser = Res_driver.parsingEngine.parseImplementation ~forPrinter:false in
+  let {Res_driver.parsetree = structure} = parser ~filename:currentFile in
+  iterator.structure iterator structure |> ignore;
+  match !foundFunctionApplicationExpr with
+  | Some (argAtCursor, exp, extractedArgs) -> (
+    (* Not looking for the cursor position after this, but rather the target function expression's loc. *)
+    let pos = exp.pexp_loc |> Loc.end_ in
+    let completions, env, package =
+      let textOpt = Files.readFile currentFile in
```
cristianoc, line 60: This is where one switches to use the parser right?

analysis/src/SignatureHelp.ml
```diff
@@ -0,0 +1,213 @@
...
+                  | None -> None
+                  | Some docs -> Some {Protocol.kind = "markdown"; value = docs});
+              };
+            ];
+          activeSignature = Some 0;
+          activeParameter =
+            (match activeParameter with
+            | None -> Some (-1)
+            | activeParameter -> activeParameter);
+        }
```
cristianoc, line 211: Not going into too many comments on this function, rather some high level observations:

1 try to split up into smaller piece and make the toplevel logic short and clear

2 check if enough has been reused of existing logic, and avoid repetitions here -- not clear that there are repetitions, not clear either that there are not

server/src/server.ts
```diff
@@ -1054,6 +1076,10 @@ function onMessage(msg: p.Message) {
                 workDoneProgress: false,
               }
             : undefined,
+          signatureHelpProvider: {
+            triggerCharacters: ["("],
```
cristianoc, line 1080: Interesting -- I did not know this had its own entire config.

cristianoc (commented):
This looks great.
Only some high level comments -- the general structure and tests look good.

Still have to play with this, only looked at the code.

Question: should this be active by default or under a config? Any thought of which one it should be?

cristianoc:
Compared to other features, this is a little more elaborate i.e. maintenance cost for the future.
I'd say the ratio cost / value is OKish.

zth:
Will pick up this again and finish it soon.

analysis/src/Commands.ml
```diff
@@ -88,6 +88,15 @@ let hover ~path ~pos ~currentFile ~debug =
   in
   print_endline result
 
+let signatureHelp ~path ~pos ~currentFile ~debug =
```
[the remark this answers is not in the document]
zth: I think most of them can be inlined. I think the `Commands` file had a purpose before, but now there's a few commands that involve larger amounts of code, which probably needs to live in their own files anyway. So I think most of the functions in Commands don't serve any purpose as of now.

analysis/src/Files.ml
```diff
@@ -102,3 +102,9 @@ let classifySourceFile path =
   if Filename.check_suffix path ".res" && exists path then Res
   else if Filename.check_suffix path ".resi" && exists path then Resi
   else Other
+
+let writeTempFile ~prefix ~suffix contents =
```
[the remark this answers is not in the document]
zth: This was intended to be temporary as I figured out how we can modify the parser to accept string content rather than forcing you to supply a file it can read from. Will try and see if I can modify the parser to support that.

As for the question itself; no, I believe this can throw an exception that wouldn't be caught here. I'll fix that if I don't think modifying the parser to support raw content is feasible to fit into this diff.

analysis/src/Loc.ml
```diff
@@ -8,3 +8,5 @@ let toString (loc : t) =
   (if loc.loc_ghost then "__ghost__" else "") ^ (loc |> range |> Range.toString)
 
 let hasPos ~pos loc = start loc <= pos && pos < end_ loc
+
+let hasPosInclusive ~pos loc = start loc <= pos && pos <= end_ loc
```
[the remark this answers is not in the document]
zth: Yeah, I really don't know... 😄 I understand this shouldn't be needed, but it was the only way I found to make it work. I'll try again and see if we can maybe figure out how to avoid this together.

analysis/src/SignatureHelp.ml
zth: Correct, copy paste from autocomplete! Definitively worth extracting to its own function.

analysis/src/SignatureHelp.ml
zth: Correct.

analysis/src/SignatureHelp.ml
zth: Got it! Will revisit.

server/src/server.ts
zth: Yeah, it's quite elaborate really.

zth:
@cristianoc ready for your eyes again.

analysis/src/SignatureHelp.ml
```diff
@@ -0,0 +1,216 @@
...
+          | None -> "none"
+          | Some (Labelled name) -> "~" ^ name
+          | Some (Unlabelled index) -> "unlabelled<" ^ string_of_int index ^ ">");
+
+      (* The LS protocol wants us to send both the full type signature (label) that the end user sees as the signature help, and all parameters in that label
+         in the form of a list of start/end character offsets. We'll leverage the parser to figure the offsets out by parsing the label, and extract the
+         offsets from the parser. *)
+
+      (* Put together a label here that both makes sense to show to the end user in the signature help, but also can be passed to the parser. *)
+      let label = "let " ^ name ^ ": " ^ Shared.typeToString type_expr in
```
cristianoc, line 178: Can you walk through the reason for printing a type expr, then parse it back right away?
Presumably it reuse some existing logic, but is it really worth it (potential for mismatch between printing and parsing)?

cristianoc (commented):
Got some philosophical question.

Also how about when named and unnamed arguments are mixed in the same call: does that work at the moment?

zth:
> Got some philosophical question.
> 
> Also how about when named and unnamed arguments are mixed in the same call: does that work at the moment?

It should, there are a few tests for it. Although I haven't tested it extensively.

analysis/src/SignatureHelp.ml
zth: The editor wants 2 things to be able to show the signature help;
1) The _string_ to show. This should be the full type definition string for the function we're looking at.
2) Offsets for all available arguments inside of that exact string, and what argument (if any) is currently active.

Producing 1) is easy obviously. 2) is a bit more tricky, since I need the exact offsets for that particular type string. So, I'm using the parser to tell me the precise location of each argument (so I can calculate offsets). And I ensure those locs/offsets are correct by using the parser to parse the exact string with the type definition that the editor will show, so the parser can tell me the precise location of each argument inside of that.

Are you following what I mean?

cristianoc: Got it. Doing that in a more principled way would require instrumenting the printer, which sounds like more work and a one-off application.
Would you add a little line of comment there?

zth: This comment is a line above:
```
(* The LS protocol wants us to send both the full type signature (label) that the end user sees as the signature help, and all parameters in that label
         in the form of a list of start/end character offsets. We'll leverage the parser to figure the offsets out by parsing the label, and extract the
         offsets from the parser. *)
```

Is that enough or should I extend that more?

zth: Done.

zth:
@cristianoc would you mind having another look?

cristianoc (approved):
Would you update the change log before merging.

PR #549, opened 2022-08-08
Fix unsafe parsedDiagnostics array push

### LINKED ISSUE #548: Server crash with TypeError: Cannot read properties of undefined (reading 'content')
Extension version `v1.5.13`

Server crashes repeatedly:
```
TypeError: Cannot read properties of undefined (reading 'content')
    at Object.parseCompilerLogOutput (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/out/utils.js:531:61)
    at /nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/out/server.js:108:68
    at Map.forEach (<anonymous>)
    at sendUpdatedDiagnostics (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/out/server.js:105:19)
    at FSWatcher.<anonymous> (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/out/server.js:179:5)
    at FSWatcher.emit (node:events:390:28)
    at FSWatcher.emitWithAll (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/node_modules/chokidar/index.js:535:32)
    at FSWatcher._emit (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/node_modules/chokidar/index.js:626:8)
    at NodeFsHandler._handleFile (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/node_modules/chokidar/lib/nodefs-handler.js:400:14)
    at NodeFsHandler._addToNodeFs (/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/node_modules/chokidar/lib/nodefs-handler.js:629:21)
[Info  - 3:59:15 PM] Connection to server got closed. Server will restart.
/nix/store/9kqvjbaq67qb6zwb2y8gvnr1y7m31mvl-vscode-extension-chenglou92-rescript-vscode-1.5.13/share/vscode/extensions/chenglou92.rescript-vscode/server/out/utils.js:531
            parsedDiagnostics[parsedDiagnostics.length - 1].content.push(line);
                                                            ^
```

fhammerschmidt:
fixes #548 

Not sure what changed or why we did not encounter this before yet. 

Maybe we want it to crash here, because there is some message shape that we did not take into account?

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -575,7 +575,7 @@ export let parseCompilerLogOutput = (
       //   10 ┆
     } else if (line.startsWith("  ")) {
       // part of the actual diagnostics message
-      parsedDiagnostics[parsedDiagnostics.length - 1].content.push(
+      parsedDiagnostics[parsedDiagnostics.length - 1]?.content.push(
         line.slice(2)
       );
     } else if (line.trim() != "") {
@@ -585,7 +585,7 @@ export let parseCompilerLogOutput = (
       // messages not printing with indent). We used to get bug reports and fix
       // the messages, but that strategy turned out too slow. One day we should
       // revert to not having this branch...
-      parsedDiagnostics[parsedDiagnostics.length - 1].content.push(line);
+      parsedDiagnostics[parsedDiagnostics.length - 1]?.content.push(line);
     }
   }
 


cristianoc:
CC @zth

zth:
Yes, exactly, we haven't done this type of fix because this in essence means we've missed cases in parsing, and we don't want to swallow that, better have people report it.

I'm thinking we could do this fix (since not crashing is good) if we combine it with prompting the user that something went wrong and logging instructions for how to report the issue with relevant information. Kind of like we did in https://github.com/rescript-lang/rescript-vscode/pull/544.

What do you think @cristianoc @fhammerschmidt ?

cristianoc:
We are already giving better reports sincea recent PR.  I would say let's see how that goes first.

zth:
You mean for the reanalyze error message?

cristianoc:
> You mean for the reanalyze error message?

Forgot what the PR was for.
But in general I think we first want an actionable report, and let it crash. Then see if that experience is adequate or something should be added later.
Just would not want to make it too easy to ignore existing issues.

cristianoc:
We are particularly vulnerable with platforms not widely used, and it would be nice to make sure that corner case issues get reported.

zth:
I think we can make the prompt "annoying" if we want, but letting it crash will mean making the extension unusable for users as long as that diagnostic is reported. I think it would be nice to avoid that, but sure, we could let it happen to make it very sure that people report the issue.

cristianoc:
> I think we can make the prompt "annoying" if we want, but letting it crash will mean making the extension unusable for users as long as that diagnostic is reported. I think it would be nice to avoid that, but sure, we could let it happen to make it very sure that people report the issue.

I was thinking to making it annoying or things might not be reported. As a first step. Then learn from there.

cristianoc:
Also, if a more robust fix is desired, then it's better to move towards a more robust fix. Rather than sweep the issue under the carpet.

The more robust fix is not to paste random text on one side, coupled with random text parsing on the other side, and hope for the best. But to export in a specified JSON format, and read that.

fhammerschmidt:
So what do you suggest should I do with this PR? Obviously, the more robust fix is a bit out of scope.

cristianoc:
I suggest to drop it. It's not very different to adding -1 to the return of a wrong program that returns the results off by 1.

Issue #349, opened 2022-01-21
Possible bugs on resolving type alias to record type


cannorin:
Version: `1.2.0`
Related: #311 

I think I've found several bugs around type alias to record type.

Assume we have the following code:

```rescript
type foo<'a> = { content: 'a }
type bar = { age: int }
type foobar = foo<bar>

let x1: foo<bar> = { content: { age: 42 } }
let x2: foobar = { content: { age: 42 } }
```

1. When I hover on `content` in `let x1: foo<bar> = { content: { age: 42 } }`, it shows `'a` instead of `bar`.
    - This is kind of "correct" behavior, as it is just failing to substitute the type variable `'a`.
    - As a result of this, I don't get autocompletion when I enter `x1.content.`.

2. When I hover on `content` in `let x2: foobar = { content: { age: 42 } }`, it shows `foo<'a>` instead of `bar`.
    - This is not correct, as `content` does not have type `foo<'a>`.
    - I also don't get autocompletion when I enter `x2.content.`.
        - This is a bit strange since if it really thinks `x2.content` has type `foo<'a>` it should provide `content` as autocompletion.
    - What is interesting here is when I hover on `age` in `x2.content.age` it shows `int`, which is correct.

cannorin:
I think part of this issue is a duplicate of #310.

cristianoc:
Confirming this is still the case, with a test https://github.com/rescript-lang/rescript-vscode/commit/be92c6c47ac34c92031b0ff5921c9c9b8dfd6b3b

cristianoc:
The hovering part is done here: https://github.com/rescript-lang/rescript-vscode/pull/560

The completion part is still open.

cristianoc:
Completion part also done.

PR #561, opened 2022-08-31
Support type instantiation during autocompletion.

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/349


cristianoc:
This takes care of records, but likely one needs to look at other cases too.

cristianoc:
All cases should be handled now.

PR #562, opened 2022-09-01
Examples of hover where type expansion is not performed across files.

cristianoc:


zth:
This is fixed in recent commits.

PR #573, opened 2022-09-05
Add rescript.settings.platformPath and correctly determine its default value

### LINKED ISSUE #572: Formatting document fails in monorepo
Extension version: 1.7.6 (latest master)

The `Format document` feature uses the `bsc.exe` binary:
https://github.com/rescript-lang/rescript-vscode/blob/877f874c52ab51e88dacc73524ca4e5c16912eef/server/src/server.ts#L740-L742

The location of the `bsc.exe` binary is determined as follows:
https://github.com/rescript-lang/rescript-vscode/blob/877f874c52ab51e88dacc73524ca4e5c16912eef/server/src/server.ts#L112-L121

It is "reverse engineered" from the location of the `rescript` binary:
https://github.com/rescript-lang/rescript-vscode/blob/877f874c52ab51e88dacc73524ca4e5c16912eef/server/src/server.ts#L107-L110
https://github.com/rescript-lang/rescript-vscode/blob/877f874c52ab51e88dacc73524ca4e5c16912eef/server/src/utils.ts#L56-L60

The `node_modules/.bin/rescript` symlink exists both at the monorepo top level `node_modules` and the subdir `node_modules`, which makes `findRescriptBinary` return `src/monorepo/subdir/node_modules/.bin/rescript`.

However, the `node_modules/rescript` directory usually exists only at the monorepo top level, so when the path of the `rescript` binary is used to derive the location of `bsc.exe`, this results in the non-existent path `src/monorepo/subdir/node_modules/rescript/linux/bsc.exe`. It should actually return `src/monorepo/node_modules/rescript/linux/bsc.exe` (without `subdir`).

To make things more complicated, if `node_modules/rescript` *does* exist in the subdir, that directory should be used. This may be the case when multiple `rescript` versions are used in the same monorepo.

Minnozz:
This introduces the setting `rescript.settings.platformPath` and fixes the algorithm to find it if the setting is not specified. This should fix formatting in monorepos, which relies on `bsc.exe`.

Fixes #572


tsnobip:
I tested it on my monorepo and it did fix my autoformatting issue! Thanks @Minnozz!

fhammerschmidt:
Tried it on two monorepo projects. I had no issues before, but also did not find any regression with this build.

zth:
@Minnozz I'm inclined to merge this since it seems to fix setups that recently broke.

Would you be available to help fix unforeseen issues that might come from this? Not necessarily thinking there will be any issues, but a lot of the ad hoc monorepo fixes we've merged has broken someone else's monorepo setup. Hence my question.

Minnozz:
@zth Yes, I'll do my best to fix any issues

zth:
> @zth Yes, I'll do my best to fix any issues

Thank you!

PR #576, opened 2022-09-05
prompt the user whenever parsing errors/warning from the compiler log fails

zth:
...including instructions on how to report the issue.

image


cristianoc (approved):
This is great.

PR #578, opened 2022-09-08
Make tests compatible with JSX V4

cristianoc:


cristianoc:
Change tests so there's one component per module.
And temporarily rename `Jsx.res[i]` as there's an issue with the stdlib.

PR #580, opened 2022-09-10
Create interface file: add support for JSX V4.

cristianoc:


cristianoc:
@mattdamon108 see how the editor support needs to stay in sync with JSX.
In particular, it needs to reverse-engineer it starting from the types obtained after the ppx run.
So in case of future changes, this should be taken into account.

analysis/src/CreateInterface.ml
```diff
@@ -213,6 +236,60 @@ let printSignature ~extractor ~signature =
...
+          },
+          _ )
+      :: Sig_value (makeId (* make *), makeValueDesc)
+      :: rest
+      when Ident.name propsId = "props"
+           && getComponentTypeV4 makeValueDesc.val_type <> None
+           && optLbls |> List.mem "key" ->
+      (* PPX V4 component declaration:
+         type props = {..., key?: _}
+         let v = ...
```
cristianoc, line 253: This is a heuristic.
Can't be super precise when working with the results of running a PPX.

cristianoc:
This should be the last known item needed for https://github.com/rescript-lang/rescript-compiler/issues/5406
Anything else, people will find out when testing more thoroughly.

PR #581, opened 2022-09-10
Test JSX V4 on 10.1 alpha.

cristianoc:
CC @mattdamon108


cristianoc:
@mattdamon108 would you update your test repository for JSX V4 to install `rescript@next` which is now 10.1. No need to use master anymore.

mattdamon108:
> @mattdamon108 would you update your test repository for JSX V4 to install `rescript@next` which is now 10.1. No need to use master anymore.

Got it! https://github.com/mattdamon108/react-jsx-ppx-v4/commit/9e8cda9a8760f6cb1f3f855c884e8a46e0c68b8f

Issue #557, opened 2022-08-30
Expand type aliases one level when hovering


zth:
Currently, when hovering something where the type is a reference to some other type, that other type is not expanded. Example:
```rescript
SomeModule.Types.searchInput
type searchInput = AnotherModule.searchInput
```

We should expand that type reference one level (?), so that the resulting hover instead would be something along the lines of:
```rescript
AnotherModule.searchInput
type searchInput = {
  someField: int
}

type searchInput = AnotherModule.searchInput
```

Not 100% of how the representation should be, but we should include the full aliased type at least.

cristianoc:
Can we have a complete example (there are some details I have doubts about)? Also, is this only about references across different files?

cristianoc:
```
  type peps  = {x:int, y:string}

  let foo = () => {x:3, y:""}
// Hover: unit => peps
```

When hovering on `foo` instead of showing just `unit => peps` it could additionally print the type definition of `peps` as in the module hover:
```
  type peps = {x: int, y: string}
```

In general, when showing a type `t`, it could collect all the type names mentioned in `t`, and for each, if it is a type definition, expand it below.

PR #584, opened 2022-09-12
Expand one level of type definition on hover and fix optional field type.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/557

### DIFF
--- a/analysis/src/Hover.ml
+++ b/analysis/src/Hover.ml
@@ -98,25 +98,53 @@ let newHover ~full:{file; package} locItem =
          | Const_int64 _ -> "int64"
          | Const_nativeint _ -> "int"))
   | Typed (_, t, locKind) ->
+    let fromConstructorPath ~env path =
+      match References.digConstructor ~env ~package path with
+      | None -> None
+      | Some (_env, {name = {txt}; item = {decl}}) ->
+        if Utils.isUncurriedInternal path then None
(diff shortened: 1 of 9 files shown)


analysis/src/Hover.ml
```diff
@@ -98,25 +98,53 @@ let newHover ~full:{file; package} locItem =
...
-        | None -> (typeString, docstring)
-        | Some (extra, extraDocstring) ->
-          (typeString ^ "\n\n" ^ codeBlock extra, extraDocstring)
+        let envToSearch, typesToSearch =
+          match typ |> Shared.digConstructor with
+          | Some path -> (
+            let labelDeclarationsTypes lds =
+              lds |> List.map (fun (ld : Types.label_declaration) -> ld.ld_type)
+            in
+            match References.digConstructor ~env ~package path with
```
zth, line 120: Unrelated to this PR, but it's a little big confusing that both `References` and `Shared` has a `digConstructor`. Or rather, hard to at a glance understand what `Shared.digConstructor` does (`References.digConstructor` is pretty easy to guess it has to do with references at least 😄 )

zth (approved):
Awesome! 💯

analysis/src/Hover.ml
cristianoc: the shared stuff is leftover that should probably be moved elsewhere

PR #585, opened 2022-09-13
Clickable links in hovers

zth:
WIP. cc @cristianoc 

TODO before merge:
- [x] Replace hand rolled incomplete `encodeURIComponent` function with something robust/complete
- [x] Gate functionality behind some sort of flag provided by the language client, as other clients outside of VSCode might not be able to support this

image

### DIFF
--- a/analysis/src/Hover.ml
+++ b/analysis/src/Hover.ml
@@ -29,6 +29,7 @@ let encodeURIComponent text =
       | '=' -> Buffer.add_string buf "%3D"
       | '?' -> Buffer.add_string buf "%3F"
       | '@' -> Buffer.add_string buf "%40"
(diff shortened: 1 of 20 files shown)


analysis/src/Hover.ml
```diff
@@ -2,6 +2,36 @@ open SharedTypes
 
 let codeBlock code = Printf.sprintf "```rescript\n%s\n```" code
 
+let encodeURIComponent text =
```
zth, line 5: This needs to be replaced by something vastly more complete/robust. Hopefully there's something for OCaml we can find and vendor or similar.

cristianoc (approved):
Great work

cristianoc (approved):

PR #587, opened 2022-09-18
add debug utilities

zth:
This adds some (hopefully) useful debug utilities from this branch https://github.com/rescript-lang/rescript-vscode/tree/type-based-autocomplete

Reason being that branch still requires quite a lot of additional work before being mergeable, but these things are immediately useful.


cristianoc (approved):
This is great!

On the comment that this is incomplete: in what way is it incomplete.
If it's cases missing or something: does it print something saying "X is incomplete" so it's clear it needs to be filled in? Or is it some other kind of incompleteness?

zth:
Yes exactly, it's cases missing, and it does print that there's an unhandled case (and mostly even what that case is).

Issue #588, opened 2022-09-18
Enable the same hover content for hovers on unsaved content as on compiled content


zth:
Hovers can currently be produced in 2 ways - either 1) the location we're hovering has a type already (the content has compiled successfully), or 2) we use the same mechanism as autocomplete on unsaved content to trace back something being hovered to something that has compiled.

Hovering in case 1 vs case 2 is different, and we should unify the hovers so case 2 produces the same hover as case 1, if possible.

cristianoc:
Can you enumerate ways in which the hover is different? So we can partition into cases that are intrinsically different because of the lack of type info, and those that just happen to be so and can be changed.

PR #589, opened 2022-09-21
Expand types on autocomplete hover

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/588

This refactors the recently introduced type expansion in hovers to be easily reusable for other purposes.
It then applies that same type expansion on the hover contents produced when the hover is done via the autocomplete functionality.

https://github.com/rescript-lang/rescript-vscode/commit/825afc253c9c1e93842890c9743abde83746dc19 this commit is probably the most relevant one, showing how this changes what's produced when hovering something unsaved but where we can still find a type to show via autocomplete.


cristianoc (approved):
Sweet!

cristianoc (approved):

PR #593, opened 2022-09-28
Add example for dead code elimination for JSX v4.

cristianoc:
DCE does not seem to be working: probably JSX call does not use proper location for the make function.


cristianoc:
https://github.com/rescript-lang/syntax/pull/660

cristianoc:
This will be fixed when the new 10.1 release candidates comes out.

Issue #607, opened 2022-10-21
Understand paths in `-open` in `bsc-flags`


cristianoc:
E.g. `"-open ReScriptJs.Js"` is not recognised
while `"-open ReScriptJs", "-open Js"` is

cristianoc:
CC @zth

cristianoc:
The semantics of `"-open ReScriptJs", "-open Js"` is not the same as `"-open ReScriptJs.Js"`. To be accurate, this should be taken into account.

So this issue is not completely fixed yet.

zth:
This is now fixed.

PR #608, opened 2022-10-21
Fix issue where module paths in `-open` in `bsc-flags` such as "-open Re…

cristianoc:
…ScriptJs.Js" were not recognized

Fixes https://github.com/rescript-lang/rescript-vscode/issues/607


zth (approved):
⭐

Issue #621, opened 2022-11-03
Fix issue where doc comment is not shown on hover in case of shadowed identifier


cristianoc:
Example:
```res
module CC = {
  /** First Comment */
  let a = 10

  /** Second Comment */
  let a = a
}

let q = CC.a // does not show comment on hover
```

See https://github.com/rescript-lang/syntax/issues/715

PR #622, opened 2022-11-03
Add example showing how shadowing affects hovering on doc comments.

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/621


cristianoc:
When an id is shadowed, the compiler generates a module constraint for the module where the id is defined.
The constraint contains a type for the id without the doc comment (an attribute), which replaces the one with doc comment.

Issue #618, opened 2022-10-29
Signature help should account for pipe


zth:
<img width="565" alt="image" src="https://user-images.githubusercontent.com/1457626/198842921-9bd00a3e-837a-4e2c-b34d-27c3a71f479b.png">
Here, being after a pipe should skip the first argument.

PR #624, opened 2022-11-07
Signature help - handle pipes

zth:
Previously, the signature help wouldn't understand pipes, meaning signature help for piped expressions would always be off by one when having unlabelled arguments. This diff fixes the signature help to account for pipes when present.

Closes https://github.com/rescript-lang/rescript-vscode/issues/618

### DIFF
--- a/analysis/src/SignatureHelp.ml
+++ b/analysis/src/SignatureHelp.ml
@@ -154,11 +154,14 @@ let signatureHelp ~path ~pos ~currentFile ~debug =
   let posBeforeCursor = Pos.posBeforeCursor pos in
   let supportsMarkdownLinks = true in
   let foundFunctionApplicationExpr = ref None in
-  let setFound r = foundFunctionApplicationExpr := Some r in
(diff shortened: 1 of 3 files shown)


analysis/src/SignatureHelp.ml
```diff
@@ -154,9 +154,68 @@ let signatureHelp ~path ~pos ~currentFile ~debug =
   let posBeforeCursor = Pos.posBeforeCursor pos in
   let supportsMarkdownLinks = true in
   let foundFunctionApplicationExpr = ref None in
-  let setFound r = foundFunctionApplicationExpr := Some r in
+  let setFound r =
+    if !foundFunctionApplicationExpr = None then
+      foundFunctionApplicationExpr := Some r
+  in
+  let searchForArgWithCursor ~isPipeExpr ~args ~exp =
```
zth, line 161: This whole function body is just moved in here from below, with very minor changes.

cristianoc (approved):

PR #626, opened 2022-11-09
fix: Incorrect type hint for module type

### LINKED ISSUE #571: Incorrect type hint for module type
```rescript
module type A = {
  let f: unit => unit
}
```

When hovering on `A`, a type hint is shown with

```rescript
module A = {
  let f: unit => unit
}
```

(with `module` instead of `module type`).

aspeddro:
Close #571

### DIFF
--- a/analysis/src/SharedTypes.ml
+++ b/analysis/src/SharedTypes.ml
@@ -6,7 +6,7 @@ type modulePath =
   | File of Uri.t * string
   | NotVisible
   | IncludedModule of Path.t * modulePath
-  | ExportedModule of string * modulePath * bool
+  | ExportedModule of {name: string; modulePath: modulePath; isType: bool}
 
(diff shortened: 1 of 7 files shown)


analysis/src/SharedTypes.ml
```diff
@@ -6,7 +6,7 @@ type modulePath =
   | File of Uri.t * string
   | NotVisible
   | IncludedModule of Path.t * modulePath
-  | ExportedModule of string * modulePath
+  | ExportedModule of string * modulePath * bool
```
cristianoc, line 9: This is getting a bit confusing above, what about an inline record so the 3 components of the payload are named.

cristianoc (approved):
Thanks for fixing this!
A style suggestion.

analysis/src/SharedTypes.ml
aspeddro: Done

Issue #511, opened 2022-07-24
Inlay hints: Investigate performance on large files


zth:
Follow up for PR https://github.com/rescript-lang/rescript-vscode/pull/453

Tried TS inlay hints on a large TS file, and the performance was significantly better than our inlay hints on a large res file. Investigate where our bottleneck is, and what we can do about it.

PR #634, opened 2022-11-14
Polish: inlay hints and codelens performance

aspeddro:
- It was calling `Cmt.fullFromPath` for each hint.
- Removes inlay hints for parameters. It's not necessary as code lenses already do this job.

A noticeable change to `analysis/examples/large-project/src/res_core.res`

Is there any way to measure?

Part of #511

cc @zth

### DIFF
--- a/analysis/src/Hint.ml
+++ b/analysis/src/Hint.ml
@@ -77,7 +77,7 @@ let inlay ~path ~pos ~maxLength ~debug =
    in
    let {Res_driver.parsetree = structure} = parser ~filename:path in
    iterator.structure iterator structure |> ignore);
-  match Cmt.fullFromPath ~path with
+  match Cmt.loadFullCmtFromPath ~path with
   | None -> None
   | Some full ->
     let result =
@@ -141,7 +141,7 @@ let codeLens ~path ~debug =
    in
    let {Res_driver.parsetree = structure} = parser ~filename:path in
    iterator.structure iterator structure |> ignore);
-  match Cmt.fullFromPath ~path with
+  match Cmt.loadFullCmtFromPath ~path with
   | None -> None
   | Some full ->
     let result =
--- a/analysis/src/Hint.ml
+++ b/analysis/src/Hint.ml
@@ -41,16 +41,6 @@ let inlay ~path ~pos ~maxLength ~debug =
     if start_line <= range.end_.line && end_line >= range.start.line then
       hints := (range, kind) :: !hints
   in
-  let rec processFunction (exp : Parsetree.expression) =
-    match exp.pexp_desc with
-    | Pexp_fun (_, _, pat_exp, e) -> (
-      match pat_exp with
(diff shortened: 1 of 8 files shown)


zth:
Having a look at this soon, sorry for the wait!

zth:
@aspeddro we don't have any real benchmarks unfortunately. When I've tested I've clocked how long refreshing takes for us in that large file vs TS in a large TS file.

analysis/src/Hint.ml
```diff
@@ -156,30 +141,34 @@ let codeLens ~path ~debug =
...
-                          (* Code lenses can run commands. An empty command string means we just want the editor
-                             to print the text, not link to running a command. *)
-                          command = "";
-                          (* Print the type with a huge line width, because the code lens always prints on a
-                             single line in the editor. *)
-                          title = typeExpr |> Shared.typeToString ~lineWidth:400;
-                        };
-                  })
-           | _ -> None))
+  match Cmt.fullFromPath ~path with
```
zth, line 144: I think this is great. Question for the future: does it make sense to have some sort of cache inside of `Cmt.fullFromPath` directly?

analysis/tests/src/expected/InlayHint.res.txt
```diff
@@ -35,36 +35,12 @@ Inlay Hint src/InlayHint.res 1:34
...
-    "label": ": string",
-    "kind": 1,
-    "paddingLeft": true,
-    "paddingRight": false
-}, {
-    "position": {"line": 10, "character": 24},
-    "label": ": int",
-    "kind": 1,
-    "paddingLeft": true,
-    "paddingRight": false
```
zth, line 49: Why are these gone?

zth (commented):
I think this looks good. @cristianoc might want to have a glance. @aspeddro mind pulling in the latest master too? There's a conflict.

analysis/src/Hint.ml
cristianoc: The change looks good. Perhaps it's possible to rename `fullFromPath` to something that makes it clear it's an expensive operation. E.g. `loadFullCmtFromPath`.

analysis/tests/src/expected/InlayHint.res.txt
aspeddro: I remove inlay hints for parameters. Code Lens already shows the types of the parameters.

Before:
image


After:
image

analysis/src/Hint.ml
aspeddro: Done in 0f260537faf103aca33519ce1da93fae0a1e2479

analysis/tests/src/expected/InlayHint.res.txt
zth: Ah right, you said why already, sorry!

zth (approved):

PR #636, opened 2022-11-18
Test create interface file JSX V4.

cristianoc:


cristianoc:
Will open a less noisy PR.

Issue #617, opened 2022-10-29
Create interface file with JSX 4 component


cknitt:
Given a JSX 4 component defined in `Test.res`:

```rescript
@react.component
let make = (~a: string) => React.string(a)
```

I get the following `Test.resi` when using the "ReScript: Create interface file..." action:

```rescript
type props<'a> = {a: 'a}
let make: props<string> => React.element
```

I would have expected to get

```rescript
@react.component
let make: (~a: string) => React.element
```

(same as for JSX 3).

cristianoc:
This, and genType check, are finishing touches for the release. There are dependencies on the actual representation of things so only make sense to do when we know nothing else changes.
Basically, we need a final rc candidate first, then these.

fhammerschmidt:
The latest master extension still does not generate a `@react.component` for me when I use JSXV4. Regardless of mode, both `automatic` and `classic` give me the same result that Christoph mentions above.

@cristianoc something is still missing

cristianoc:
> The latest master extension still does not generate a `@react.component` for me when I use JSXV4. Regardless of mode, both `automatic` and `classic` give me the same result that Christoph mentions above.
> 
> @cristianoc something is still missing

What compiler version? Try `10.1.0-rc.4`.

fhammerschmidt:
I tried with both `10.1.0-rc.4`  and `10.1.0-rc.5`.

cristianoc:
@fhammerschmidt can you start from the project in this? https://github.com/rescript-lang/rescript-vscode/pull/636
I've switched to V4 and the interface file is create in CI. Should work from the editor directly too.

cristianoc:
OK I have some examples that don't work.

cristianoc:
@fhammerschmidt should be fixed now.

fhammerschmidt:
Thank you very much!

PR #637, opened 2022-11-18
Fix issue where create interface file does not work with V4

cristianoc:
See https://github.com/rescript-lang/rescript-vscode/issues/617


analysis/tests/src/expected/JsxV4.res.txt
```diff
@@ -16,3 +16,17 @@ Completable: Cjsx([M4], f, [first, f])
...
+module M4: {
+  @react.component
+  let make: (~first: string, ~fun: string=?, ~second: string=?) => React.element
+}
+module MM: {
+  @react.component
+  let make: unit => React.element
+}
+module Other: {
+  type props<'name> = {name: 'name}
```
cristianoc, line 29: CC @fhammerschmidt  this one still does not work

Issue #643, opened 2022-12-01
Autocomplete stops working after `evt->ReactEvent.Synthetic.preventDefault`


fhammerschmidt:
Very curious:

```res
let onClick = evt => {
  // Autocomplete works here, e.g.:
  // ReactEvent.
  evt->ReactEvent.Synthetic.preventDefault
  // Autocomplete does not work here, e.g.:
  // ReactEvent.

  Js.log("Hello")
}
```
<img width="581" alt="image" src="https://user-images.githubusercontent.com/18074327/205032851-0ee41653-5779-4bf9-a54e-3f27ccac6619.png">

<img width="473" alt="image" src="https://user-images.githubusercontent.com/18074327/205032761-d6642575-66dd-45e8-bb90-ea478603b0d8.png">

zth:
Can you create a reproducible example? Better yet a failing test in the analysis folder.

fhammerschmidt:
Done, and it got even weirder...

PR #644, opened 2022-12-01
Add failing completion test

fhammerschmidt:
Added the failing test mentioned in #643

The funny thing is, I tried to change the `Js.log("Hello")` to a `()` which made the second completion work as well. No idea what's going on there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@
 
 - Fix issue where create interface file would not work with certain JSX V4 components https://github.com/rescript-lang/rescript-vscode/issues/617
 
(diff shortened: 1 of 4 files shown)


cristianoc:
Looks like it's not what comes before, but what follows.

First
```
Completable: Cpath Value[ReactEvent, ""]
```

then
```
Completable: Cpath Value[ReactEvent, "", log]
```

so parser recovery somehow picks up `log` too.

cristianoc:
For context, this is a correct program:

```res
module S = {
type r = {xx:int}
}

let r = {S.xx:10}

let f = r.
S.xx
```

so playing with parser recovery is a little tricky in these cases.

PR #651, opened 2022-12-13
Redo pr 649

cristianoc:


cristianoc:
@aspeddro this should work, only needs cleaning up the changelog.

aspeddro:
I thought it was fixed in master. Thanks @cristianoc.

cristianoc:
These are 2 different issues:
1 references to elements defined in interface files
2 references from implementation files which also happen to have interface files

cristianoc:
1 in master
2 here

cristianoc:
All done now.

PR #658, opened 2022-12-18
add vscode sandbox

aspeddro:


cristianoc (approved):
What does this do in practice?

aspeddro:
This is to enable local sandbox switch when open VSCode.

I was getting this error, because VSCode use global switch.

image

PR #662, opened 2022-12-21
Pipe completion unknown/type parameter return types

zth:
This fixes pipe completion for functions returning a type not known until the code has compiled. `Array.reduce` being an example, which has the signature `('a, ('a, 'b) => 'a) => 'a`. It returns `'a` but we don't know what `'a` is unless it has compiled.

This makes it so that this now completes as `int` (after saving):
```
let _ = [123]->Belt.Array.reduce(0, (acc, curr) => acc + curr)->
```

Again, note this only has effect _after saving_ (compiling so that the type parameter is instantiated). But it at least enables pipe completion in this scenario, even if it requires saving, which used to not work at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,6 @@
 <p align="center">The Official VSCode plugin for ReScript</p>
 
 <p align="center">
-  <a href="https://marketplace.visualstudio.com/items?itemName=chenglou92.rescript-vscode">
-    <img src="https://vsmarketplacebadge.apphb.com/version/chenglou92.rescript-vscode.svg"/>
-    <img src="https://vsmarketplacebadge.apphb.com/installs/chenglou92.rescript-vscode.svg"/>
-  </a>
-</p>
-
-<p align="center">
   <img src="https://user-images.githubusercontent.com/1909539/101266821-790b1400-3707-11eb-8e9f-fb7e36e660e6.gif"/>
 </p>
 
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1275,7 +1275,7 @@ let rec getCompletionsForContextPath ~package ~opens ~rawOpens ~allFiles ~pos
                else None)
       | None -> [])
     | None -> [])
-  | CPPipe (cp, funNamePrefix) -> (
+  | CPPipe {contextPath = cp; id = funNamePrefix} -> (
     match
       cp
       |> getCompletionsForContextPath ~package ~opens ~rawOpens ~allFiles ~pos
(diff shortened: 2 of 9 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1157,6 +1157,17 @@ let completionsGetTypeEnv = function
   | {Completion.kind = Field ({typ}, _); env} :: _ -> Some (typ, env)
   | _ -> None
 
+let findReturnTypeOfFunctionAtLoc loc ~(env : QueryEnv.t) ~package ~debug =
+  match Cmt.loadFullCmtFromPath ~path:(env.file.uri |> Uri.toPath) with
```
cristianoc, line 1161: Since this seems to be intended for the current file,  we should have this info available since the `getCompletions` function. Perhaps pass around `~full` instead of `~package`, and when we get here it's already there.

analysis/src/SharedTypes.ml
```diff
@@ -479,7 +479,7 @@ module Completable = struct
     | CPId of string list * completionContext
     | CPField of contextPath * string
     | CPObj of contextPath * string
-    | CPPipe of contextPath * string
+    | CPPipe of {contextPath: contextPath; id: string; lhsLoc: Warnings.loc}
```
cristianoc, line 482: `Location.t` more memorable than `Warnings.loc`.

analysis/tests/src/expected/CompletionPipeChain.res.txt
```diff
@@ -239,3 +239,56 @@ posCursor:[58:5] posNoWhite:[58:4] Found expr:[57:8->0:-1]
 Completable: Cpath Value[Js, Array2, forEach](Nolabel, Nolabel)->
 []
 
+Complete src/CompletionPipeChain.res 62:5
+posCursor:[62:5] posNoWhite:[62:4] Found expr:[61:8->0:-1]
+Completable: Cpath Value[Belt, Array, reduce](Nolabel, Nolabel, Nolabel)->
```
cristianoc, line 244: Perhaps add a letter after `->` to cut down the result size of the test.

cristianoc (commented):
This seems to be a reasonable compromise.
I think the full solution would involve changing the case of application, which would have to find a way to instantiate function type parameters. That would need to get one's hands dirty with unification though, which is much more complex.

zth:
@cristianoc ready for another look through, I think I fixed all of your points.

README.md
```diff
@@ -4,13 +4,6 @@
 
 <p align="center">The Official VSCode plugin for ReScript</p>
 
-<p align="center">
```
cristianoc, line 7: What about these changes?

cristianoc (approved):

PR #665, opened 2022-12-25
Complete function argument values

zth:
The first in likely a long series of PRs bringing in and refining the work done in https://github.com/rescript-lang/rescript-vscode/pull/493, but little bit by bit.

This PR does the following:
- Completes the expressions of labelled arguments or unlabelled arguments in function applications.
- Currently completes booleans, options and variants. The rest of the relevant types (polyvariants, records, tuples, etc) will be completable as well in due time.

There are caveats that I've added comments for in the PR.

The next step will be getting support for snippets in completion items in, so we can intelligently move the cursor for the user into payloads of constructors etc. That'll come in a separate PR.
After that, completing polyvariants. 
Then the same work as in this PR, but for JSX props.

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1461,9 +1461,11 @@ let getArgs ~env (t : Types.type_expr) ~full =
     match t.desc with
     | Tlink t1 | Tsubst t1 | Tpoly (t1, []) ->
       getArgsLoop ~full ~env ~currentArgumentPosition t1
-    | Tarrow ((Labelled l | Optional l), tArg, tRet, _) ->
+    | Tarrow (Labelled l, tArg, tRet, _) ->
       (SharedTypes.Completable.Labelled l, tArg)
       :: getArgsLoop ~full ~env ~currentArgumentPosition tRet
+    | Tarrow (Optional l, tArg, tRet, _) ->
+      (Optional l, tArg) :: getArgsLoop ~full ~env ~currentArgumentPosition tRet
     | Tarrow (Nolabel, tArg, tRet, _) ->
       (Unlabelled {argumentPosition = currentArgumentPosition}, tArg)
       :: getArgsLoop ~full ~env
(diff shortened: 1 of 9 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1455,6 +1455,131 @@ let getOpens ~debug ~rawOpens ~package ~env =
   (* Last open takes priority *)
   List.rev resolvedOpens
 
+let getArgs ~env (t : Types.type_expr) ~full =
```
zth, line 1458: This one is mostly copied from a pre-existing fn, but with identifying what type of argument is found (labelled or unlabelled).

analysis/src/CompletionBackEnd.ml
```diff
@@ -1455,6 +1455,131 @@ let getOpens ~debug ~rawOpens ~package ~env =
...
+          |> List.filter_map (fun (constructor : Constructor.t) ->
+                 if
+                   prefix <> ""
+                   && not (Utils.startsWith constructor.cname.txt prefix)
+                 then None
+                 else
+                   Some
+                     (Completion.create
+                        ~name:
+                          (constructor.cname.txt
```
zth, line 1540: Bonus points @cristianoc - can we somehow figure out if we need to qualify the constructor here? Not something we need to do now, but something worth thinking about for the future.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1455,6 +1455,131 @@ let getOpens ~debug ~rawOpens ~package ~env =
...
+        items
+      | Some (Toption (env, t)) ->
+        let items =
+          [
+            Completion.create ~name:"None"
+              ~kind:(Label (t |> Shared.typeToString))
+              ~env;
+            Completion.create ~name:"Some(_)"
+              ~kind:(Label (t |> Shared.typeToString))
+              ~env;
```
zth, line 1564: I'm going to improve this a lot soon. Mainly 2 things:
1. `Some(_)` will eventually be a snippet `Some($1:_)`, which means that the cursor is automatically positioned in the option payload (so we can keep completing the option value)
2. If the inner type is "small" enough (bool, variant with few payloads, etc), we can expand those into the list directly, so that instead of `None/Some(_)`, it'd be `None/Some(true)/Some(false)` for bools, etc.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1455,6 +1455,131 @@ let getOpens ~debug ~rawOpens ~package ~env =
...
+        else
+          items
+          |> List.filter (fun (item : Completion.t) ->
+                 Utils.startsWith item.name prefix)
+      | Some (Tvariant {env; constructors; variantDecl; variantName}) ->
+        let items =
+          constructors
+          |> List.filter_map (fun (constructor : Constructor.t) ->
+                 if
+                   prefix <> ""
```
cristianoc, line 1533: Perhpaps factor out a filter function to be used here and above.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1817,6 +1942,26 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
...
+      | Some (typ, _env) -> typ |> getArgs ~full ~env
+      | None -> []
+    in
+    let targetLabel =
+      labels |> List.find_opt (fun (label, _) -> label = argumentLabel)
+    in
+    match targetLabel with
+    | None -> []
+    | Some (Labelled _, typ) ->
+      typ |> completeTypedValue ~env ~full ~prefix ~expandOption:true
```
cristianoc, line 1962: Why expand option on every labeled arg?

cristianoc (approved):

analysis/src/CompletionBackEnd.ml
zth: Trying to capture that you don't need to qualify optional values, but that's only for optional arguments, right? So this should be for optional, not labeled, arguments

zth: Fixed in a recent commit.

Issue #322, opened 2021-11-27
[Feature Request] Autocompletion for JSX variant/polymorphic variant props 


Lomand:
Greetings,

Having the ability to autocomplete props whose type is variant/polymorphic variant should be pretty neat.

Right now you can see the type of the prop

![image](https://user-images.githubusercontent.com/8079999/143719905-95562da4-9227-4589-a630-69cba5d46170.png)

but can't actually autocomplete it.

![image](https://user-images.githubusercontent.com/8079999/143719914-30bf7ab9-59dc-4be9-91ef-d8df8414b59d.png)

PS: Cristiano, thanks a lot for making autocompletion for objects 🙏

PR #667, opened 2022-12-29
complete JSX prop values

zth:
This is ready for an initial review @cristianoc . A few things to consider:
* Can we figure out if we need to qualify variant constructors as we complete them? That's a typical issue with JSX props IIRC, that you almost always have to qualify them so the compiler can find them.
* How can we figure out if we need to expand options for JSX props? With fn arguments it was "easy" - optional label should expand option. How do we do it in JSX, do we need to look up the underlying make function? Maybe we already have access to what we need and I just need a gentle push in the right direction to connect the dots.

Closes https://github.com/rescript-lang/rescript-vscode/issues/322

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1572,6 +1572,81 @@ let completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
   in
   completeTypedValueInner ~env ~full ~prefix ~expandOption
 
+let getJsxLabels ~componentPath ~findTypeOfValue ~package =
+  match componentPath @ ["make"] |> findTypeOfValue with
+  | Some (typ, make_env) ->
+    let rec getFieldsV3 (texp : Types.type_expr) =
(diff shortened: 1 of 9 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1572,6 +1572,81 @@ let completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
   in
   completeTypedValueInner ~env ~full ~prefix ~expandOption
 
+let getJsxLabels ~componentPath ~findTypeOfValue ~package =
```
zth, line 1575: This fn is mostly just extracted for reuse, and adapted so it also returns the relevant `env` for the labels produced.

cristianoc (approved):
I think variants are pretty much always unqualified in these completion cases driven by types, as the prop (or label in case of fun) determines the type the variant belongs to.
One counter example that comes to mind is when the prop has a generic type, but I guess in that case we're not able to do typed completion to begin with.

PR #668, opened 2022-12-29
Completion snippet support

zth:
This adds support for snippets in completion. Snippets lets us tell the editor where we want the cursor to move as the user tabs its way through the newly inserted text.
* Add snippet support
* Implement snippets for a few existing things like completion options, variant constructor arguments, etc
* Implement very basic completion for tuples, including with snippet support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - Add autocomplete for function argument values (booleans, variants and options. More values coming), both labelled and unlabelled. https://github.com/rescript-lang/rescript-vscode/pull/665
 - Add autocomplete for JSX prop values. https://github.com/rescript-lang/rescript-vscode/pull/667
+- Add snippet support in completion items. https://github.com/rescript-lang/rescript-vscode/pull/668
 
 #### :nail_care: Polish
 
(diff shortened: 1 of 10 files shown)


cristianoc (commented):
Is the support for snippets something configured from the server?
Not sure how that's supposed to work in practice. If it's a server side config or something else.

zth:
> Is the support for snippets something configured from the server? Not sure how that's supposed to work in practice. If it's a server side config or something else.

Basically, the client hooking up the server says whether it supports snippets or not. If it supports snippets, we're allowed to use the snippet syntax in `insertText` for our completion items. 

So, in practice, this is progressive enhancement. If the client using the LSP supports snippets we'll send them if wanted, and if not, we won't.

cristianoc:
Do we support clients that don't support snippets? Just wondering whether we should just have this on.

zth:
> Do we support clients that don't support snippets? Just wondering whether we should just have this on.

I'm pretty sure there are variants of vim people are using that at least requires extra configuration to support snippets. I think it's safer to leave it turned off unless the client asks for it. Which reminds me I need to actually hook things up properly here to what the client says, I always send `true` now.

zth:
@cristianoc fixed in the last commit.

cristianoc (approved):
Looks good as long as snippet support is hooked up from the .ts side.

PR #673, opened 2023-01-04
Adjust autocomplete when Belt is open

zth:
Adjusts what modules we complete from when `Belt` is open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,14 +12,18 @@
 
 ## master
 
-## v1.10.0
-
 #### :rocket: New Feature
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +21,12 @@
 
 #### :nail_care: Polish
 
+- Prefer opened `Belt` modules in autocomplete when `-open Belt` is detected in `bsconfig`. https://github.com/rescript-lang/rescript-vscode/pull/673
+
+## v1.10.0
```
zth, line 26: This was accidentally in the wrong place, so I put it in the correct location.

cristianoc (approved):

PR #674, opened 2023-01-04
add broken completion case for result

zth:
Accidentally completes from Pervasives when it should complete from Belt.Result.

 The resulting path we're completing from becomes `result.Pervasives`, when we're looking for just `result`. So the completion ends up being derived from `Pervasives`. 

What do you think is the best way to solve this @cristianoc?


cristianoc:
> Accidentally completes from Pervasives when it should complete from Belt.Result.
> 
> The resulting path we're completing from becomes `result.Pervasives`, when we're looking for just `result`. So the completion ends up being derived from `Pervasives`.
> 
> What do you think is the best way to solve this @cristianoc?

I guess that's because that's where `result` is defined.
The way to go is probably to intercept the path in the same way one intercepts the path to array, and reroute it where it makes sense.

cristianoc:
See if it behaves as expected now.

zth:
That's perfect! @cristianoc are there more built ins that might need this treatment that you can think of? I can't come up with any.

PR #678, opened 2023-01-05
remove redundant function name from signature help output

zth:
Cleans up the output some by getting rid of the redundant `let nameOfFn: ` part of what's shown to the end user.

### DIFF
--- a/analysis/src/SignatureHelp.ml
+++ b/analysis/src/SignatureHelp.ml
@@ -30,15 +30,15 @@ let findFunctionType ~currentFile ~debug ~path ~pos =
               file )))
   in
   match completables with
-  | Some ({kind = Value type_expr; name; docstring} :: _, env, package, file) ->
+  | Some ({kind = Value type_expr; docstring} :: _, env, package, file) ->
     let args, _ =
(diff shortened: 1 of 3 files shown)


analysis/src/SignatureHelp.ml
```diff
@@ -296,19 +306,24 @@ let signatureHelp ~path ~pos ~currentFile ~debug =
...
-          let parameters = extractParameters ~signature ~label in
+          let parameters =
+            extractParameters ~signature ~typeStrForParser ~labelPrefixLen
+          in
           if debug then
             Printf.printf "extracted params: \n%s\n"
               (parameters
               |> List.map (fun (_, start, end_) ->
-                     String.sub label start (end_ - start))
+                     String.sub fnTypeStr start (end_ - start))
```
cristianoc, line 326: Do we know these are safe bounds for the substring?

cristianoc (approved):

analysis/src/SignatureHelp.ml
zth: Yes, they are derived from the locations the parser tells us so they should be fine. This also only runs in the tests so an exception shouldn't be too damaging.

PR #681, opened 2023-01-07
complete for functions creating React.element when in a JSX context

zth:
Something I've wanted for a while. This will add `React.string`, `React.array` etc to the completion items for relevant types in pipe when in a JSX context. Being in a JSX context is defined as a) being in the body of a `make` function annotated with `@react.component`, or b) being inside of JSX.

Example:
Doing pipe completion on an `int` in JSX context will return `React.int` in the completion items, in addition to the builtin completions for `int`.

This accomplishes two things:
1. Convenience, since you otherwise have to first complete for `React`, and then the fn you're after.
2. Helps beginners with understanding that you in ReScript need to use the respective `React.element` producing functions in JSX.

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1477,7 +1477,7 @@ let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
                else None)
       | None -> [])
     | None -> [])
-  | CPPipe {contextPath = cp; id = funNamePrefix; lhsLoc} -> (
+  | CPPipe {contextPath = cp; id = funNamePrefix; lhsLoc; inJsx} -> (
     match
       cp
       |> getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
@@ -1584,7 +1584,7 @@ let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
         | None -> None
       in
       match completionPath with
-      | Some completionPath ->
+      | Some completionPath -> (
         let completionPathMinusOpens =
           completionPath
           |> removeRawOpens package.opens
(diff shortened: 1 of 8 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1193,6 +1193,46 @@ let findReturnTypeOfFunctionAtLoc loc ~(env : QueryEnv.t) ~full ~debug =
     | _ -> None)
   | _ -> None
 
+(** Pulls out a type we can complete from a type expr. *)
+let rec extractType ~env ~package (t : Types.type_expr) =
```
zth, line 1197: Just moved this, no changes.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1448,14 +1488,50 @@ let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
...
+              match typ |> extractType ~env ~package with
+              | Some (Tarray (_env, payload)) -> (
+                match
+                  payload |> getTypePath |> Option.map pathIdentToString
+                with
+                | Some "React.element" -> true
+                | _ -> false)
+              | _ -> false
+            in
+            if isReactElementArray then Some "array" else None
```
zth, line 1521: This only adds the array completions for arrays where we know for sure it contains `React.element`. The problem with that is that it requires compilation.

So, an alternative is to just add it to completions for all `array`, regardless of whether it's valid or not. That might produce a better experience.

analysis/src/CompletionFrontEnd.ml
```diff
@@ -774,6 +775,14 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
     if not !processed then
       Ast_iterator.default_iterator.structure_item iterator item
   in
+  let value_binding (iterator : Ast_iterator.iterator)
+      (value_binding : Parsetree.value_binding) =
+    if
+      locHasCursor value_binding.pvb_expr.pexp_loc
+      && Utils.isReactComponent value_binding
+    then inJsxContext := true;
```
cristianoc, line 783: where is this set back to its previous value (whether false or not)?

analysis/src/CompletionBackEnd.ml
cristianoc: Perhaps the simpler version that kicks in for all arrays is less surprising -- one less difference to observe.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1448,14 +1488,50 @@ let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
...
+                 {
+                   completion with
+                   name = completionName completion.name;
+                   env
+                   (* Restore original env for the completion after x->foo()... *);
+                 })
+        in
+        (* We add React element functions to the completion if we're in a JSX context *)
+        let forJsxCompletion =
+          match (inJsx, getTypePath typ) with
```
cristianoc, line 1503: Nit: it might be cleaner to check `inJsx` outside the match, as all the cases check for `true`.

analysis/tests/src/expected/CompletionJsx.res.txt
```diff
@@ -0,0 +1,389 @@
...
+posCursor:[15:21] posNoWhite:[15:20] Found expr:[14:4->34:10]
+posCursor:[15:21] posNoWhite:[15:20] Found expr:[15:7->34:10]
+posCursor:[15:21] posNoWhite:[15:20] Found expr:[15:7->15:21]
+Completable: Cpath Value[someString]->st <<jsx>>
+[{
+    "label": "React.string",
+    "kind": 12,
+    "tags": [],
+    "detail": "string",
+    "documentation": null
```
cristianoc, line 34: pergnhaps a few words of doc?

analysis/src/CompletionBackEnd.ml
zth: ✅

analysis/src/CompletionFrontEnd.ml
```diff
@@ -1115,6 +1115,7 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
           processed := true
         | _ -> ());
       if not !processed then Ast_iterator.default_iterator.expr iterator expr
+      else inJsxContext := oldInJsxContext
```
cristianoc, line 1118: The only case that sets `inJsxContext`, case `Pexp_apply`, does not set `processed` to true.
So the `else` branch would not be executed.
Instead of just fixing the branch, what about restoring it unconditionally so it's future-proof.

cristianoc (commented):
Restoring context seems wrong.

cristianoc (approved):
Go ahead merge after fixing the code for restoring the context.

PR #682, opened 2023-01-07
Nested expression completion

zth:
This extends the current expression completion to handle more cases (records, arrays, tuples, etc), and also handle nested expressions (record fields, constructor payloads, etc etc).

It's essentially a one-to-one port of the nested pattern completion, but for expressions instead.

After this lands, we can soon start refactoring and break up the giant CompletionFrontend/CompletionBackend a bit in my opinion.

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -2299,12 +2299,8 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
                  in which we send it. This fixes that by providing a sort text making the typed completions
                  guaranteed to end up on top. *)
               items
-              |> List.mapi (fun index (c : Completion.t) ->
-                     {
-                       c with
-                       sortText =
-                         Some ("A" ^ string_of_int (index + 1) ^ " " ^ c.name);
-                     })
+              |> List.map (fun (c : Completion.t) ->
+                     {c with sortText = Some ("A" ^ " " ^ c.name)})
             else items
           in
           items @ regularCompletions
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -2281,10 +2281,25 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
         match (prefix, completionContext) with
         | "", _ -> items
         | _, None ->
+          (* Completions for local things like variables in scope, modules in the project, etc. *)
           let regularCompletions =
             prefix
             |> getComplementaryCompletionsForTypedValue ~opens ~allFiles ~env
                  ~scope
+          in
+          let items =
+            if List.length regularCompletions > 0 then
+              (* The client will occasionally sort the list of completions alphabetically, disregarding the order
+                 in which we send it. This fixes that by providing a sort text making the typed completions
+                 guaranteed to end up on top. *)
+              items
+              |> List.mapi (fun index (c : Completion.t) ->
+                     {
+                       c with
+                       sortText =
+                         Some ("AAA" ^ string_of_int (index + 1) ^ " " ^ c.name);
+                     })
+            else items
           in
           items @ regularCompletions
         | _ -> items)))
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1717,6 +1717,7 @@ let rec extractType ~env ~package (t : Types.type_expr) =
   | Tconstr (Path.Pident {name = "array"}, [payloadTypeExpr], _) ->
     Some (Tarray (env, payloadTypeExpr))
   | Tconstr (Path.Pident {name = "bool"}, [], _) -> Some (Tbool env)
+  | Tconstr (Path.Pident {name = "string"}, [], _) -> Some (Tstring env)
   | Tconstr (path, _, _) -> (
     match References.digConstructor ~env ~package path with
     | Some (env, {item = {decl = {type_manifest = Some t1}}}) ->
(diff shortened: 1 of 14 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1780,29 +1780,29 @@ let getJsxLabels ~componentPath ~findTypeOfValue ~package =
     typ |> getLabels
   | None -> []
 
-(** This moves through a pattern via a set of instructions, trying to resolve the type at the end of the pattern. *)
-let rec resolveNestedPattern typ ~env ~package ~nested =
+(** This moves through a nested path via a set of instructions, trying to resolve the type at the end of the path. *)
+let rec resolveNested typ ~env ~package ~nested =
```
zth, line 1784: I changed this slightly so it can be used for both patterns and expressions.

analysis/src/CompletionBackEnd.ml
```diff
@@ -2153,16 +2156,20 @@ Note: The `@react.component` decorator requires the react-jsx config to be set i
...
-      typ
-      |> completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
-           ~expandOption:false ~includeLocalValues:true ~completionContext:None)
+    | Some (_, typ) -> (
+      match typ |> resolveNested ~env ~package:full.package ~nested with
+      | None -> []
+      | Some (typ, env, completionContext) ->
+        typ
+        |> completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
+             ~expandOption ~includeLocalValues:true ~completionContext))
```
cristianoc, line 2172: Not introduced here, but it might be simpler to use a function that expands the option before calling `completeTypedValue`, so less indirect action at a distance and recursive call to implement the "expand on only once" logic.

cristianoc (approved):

analysis/src/CompletionBackEnd.ml
```diff
@@ -1882,18 +1882,21 @@ let rec processCompletable ~debug ~full ~scope ~env ~pos ~forHover
...
-      typ
-      |> completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
-           ~expandOption:true ~includeLocalValues:true ~completionContext:None)
+    | Some (_, typ, env) -> (
+      match typ |> resolveNested ~env ~package:full.package ~nested with
+      | None -> []
+      | Some (typ, env, completionContext) ->
+        typ
+        |> completeTypedValue ~env ~envWhereCompletionStarted ~full ~prefix
+             ~expandOption:true ~includeLocalValues:true ~completionContext))
```
cristianoc, line 1899: Pre-existing: not sure why `expandOption` is always `true` here. Should there be a check that the label is optional?

zth: I'd love to, but I don't know how to figure that out in JSX. I'd need to parse the actual component and look whether the labelled arg in the make function is optional, right?

analysis/src/CompletionBackEnd.ml
zth: Fixed

analysis/src/CompletionBackEnd.ml
```diff
@@ -931,6 +931,30 @@ let findLocalCompletionsForValuesAndConstructors ~(localTables : LocalTables.t)
        (processLocalModule ~prefix ~exact ~env ~localTables);
   List.rev_append localTables.resultRev valuesFromOpens
 
+let findLocalCompletionsForValues ~(localTables : LocalTables.t) ~env ~prefix
```
zth, line 934: Slimmed down version of a pre-existing function (but that handles both this + constructors). We don't want constructors in typed completion, since they are handled by the typed completion itself.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1098,6 +1122,28 @@ let extractFunctionType ~env ~package typ =
   in
   loop ~env [] typ
 
+let getComplementaryCompletionsForTypedValue ~opens ~allFiles ~scope ~env prefix
```
zth, line 1125: Also a slimmed down version of a pre-existing fn.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1193,6 +1239,111 @@ let findReturnTypeOfFunctionAtLoc loc ~(env : QueryEnv.t) ~full ~debug =
     | _ -> None)
   | _ -> None
 
+let getJsxLabels ~componentPath ~findTypeOfValue ~package =
```
zth, line 1242: Moved, not changed.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1193,6 +1239,111 @@ let findReturnTypeOfFunctionAtLoc loc ~(env : QueryEnv.t) ~full ~debug =
...
+        when Path.name clPath = "React.componentLike"
+             && Path.last path = "props" ->
+        (* JSX V4 external or interface *)
+        getFieldsV4 ~path ~typeArgs
+      | _ -> []
+    in
+    typ |> getLabels
+  | None -> []
+
+let getArgs ~env (t : Types.type_expr) ~full =
```
zth, line 1317: Moved, not changed.

analysis/src/CompletionBackEnd.ml
cristianoc: Yes, this can be considered later I guess.

analysis/tests/src/expected/CompletionFunctionArguments.res.txt
```diff
@@ -25,7 +25,8 @@ Completable: Cexpression CArgument Value[someFn](~isOn)=t
     "kind": 4,
     "tags": [],
     "detail": "bool",
-    "documentation": null
+    "documentation": null,
+    "sortText": "AAA1 true"
```
cristianoc: Do we need `AAA`?
Do we need `true`?
What's bigger between `"1"` and `"10`"?

cristianoc (approved):
some minor comments

analysis/tests/src/expected/CompletionFunctionArguments.res.txt
zth: Changed to A + the actual name. Should be enough.

Issue #686, opened 2023-01-10
Semantic highlighting broken


fhammerschmidt:
Unfortunately, my latest fix (#685) broke something else. Namely the semantic highlighting stopped working, but I already know why. I proposed a fix here: https://github.com/microsoft/vscode-languageserver-node/pull/1168

fhammerschmidt:
Works again, closing...

PR #689, opened 2023-01-11
Fix semantic highlighting

fhammerschmidt:
This workaround brings back semantic highlighting until the referenced PR in #686 gets merged.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -1135,7 +1135,7 @@ function onMessage(msg: p.Message) {
               ],
               tokenModifiers: [],
             },
-            documentSelector: null,
+            documentSelector: [{ scheme: "file", language: "rescript" }],
             // TODO: Support range for full, and add delta support
             full: true,
           },


zth:
What does `documentSelector` actually do and why do we need it to be null? And why does it work as we want without setting it? 😄

fhammerschmidt:
A `documentSelector` is a filter to narrow the scope of the extension to e.g. a language. Which [we do on the client side ](https://github.com/rescript-lang/rescript-vscode/blob/master/client/src/extension.ts#L103) anyway. 

The most recent version of `vscode-languageserver-node` however broke the precedence between client and server configuration. According to its documentation the client options should be used if the server's is `null` but that is not the current behaviour. Until the PR in the comment gets merged at least.

So I figured, why not just comment it out on the server? And it works.

zth:
Can't we just use the client configuration in the server too then? Ie add what we have there instead of `null`?

zth (approved):

PR #690, opened 2023-01-11
expand options in typed expr/pattern completion

zth:
This is a QoL thing to make working with options easier in typed completion, allowing you to skip a step completing inside of `Some` and directly have access to the possible `Some` values.

### DIFF
--- a/analysis/tests/src/expected/CompletionExpressions.res.txt
+++ b/analysis/tests/src/expected/CompletionExpressions.res.txt
@@ -167,6 +167,14 @@ Completable: Cexpression CArgument Value[fnTakingRecord]($0)->recordField(nested
     "documentation": null,
     "insertText": "Some(${1:_})",
     "insertTextFormat": 2
+  }, {
+    "label": "Some({})",
+    "kind": 12,
+    "tags": [],
+    "detail": "otherRecord",
+    "documentation": null,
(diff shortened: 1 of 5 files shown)


analysis/tests/src/expected/CompletionExpressions.res.txt
```diff
@@ -167,6 +167,14 @@ Completable: Cexpression CArgument Value[fnTakingRecord]($0)->recordField(nested
     "documentation": null,
     "insertText": "Some(${1:_})",
     "insertTextFormat": 2
+  }, {
+    "label": "Some({})",
+    "kind": 12,
+    "tags": [],
+    "detail": "otherRecord",
+    "documentation": null,
+    "insertText": "Some({$0})",
```
cristianoc, line 176: is this intended in addition to the previous completion?

zth: Yes. Logic being that you might want to do one of three things:
- Add None
- Add Some but with any
- Add Some, plus whatever value for the nested type you're interested in

That's why I'm adding it in addition rather than fully replacing the completions.

cristianoc: OK seems sensible.

cristianoc (approved):

PR #695, opened 2023-01-13
support completing in/through inline records in variant payloads

zth:


analysis/src/CompletionBackEnd.ml
```diff
@@ -1793,29 +1793,35 @@ let printConstructorArgs argsLen ~asSnippet =
...
   else ""
 
-let rec completeTypedValue (t : Types.type_expr) ~env ~full ~prefix
+let rec completeTypedValue (t : SharedTypes.completionType) ~env ~full ~prefix
     ~completionContext =
-  match t |> extractType ~env ~package:full.package with
+  let extractedType =
+    match t with
+    | TypeExpr t -> t |> extractType ~env ~package:full.package
+    | InlineRecord fields -> Some (TinlineRecord {env; fields})
```
cristianoc, line 1801: I was a wondering why we need a variant called `InlineRecord` in two places in SharedTypes.
I guess the question can be rephrased here as: do we need `TypeExpr` and `InlineRecord` to exist at the same level here? Somehow an inline record is not a self-standing type.
Not sure I have something concrete to suggest, just an open-ended question.

cristianoc (approved):

analysis/src/CompletionBackEnd.ml
zth: It's a good question for sure, and I don't have a clear/good answer either. One extra bit of context is that I intend to add another constructor there as well soon, for handling completing from Type.t (via type annotations for example).

I guess I view this as the "type source" to derive what to complete from, and this was the fastest/easiest way to handle inline records since they have no type expression of their own. And I guess the same will be true for types derived from type annotations?

zth: Let's come back to this eventually. I'll merge now so I can move on to the next thing.

PR #697, opened 2023-01-14
break up CompletionBackend and CompletionFrontend into modules

zth:
I've just moved things around and tried to move things into sub modules. Should hopefully make the code a bit more approachable. More can be done of course, but I think this is a good start.


cristianoc (approved):
This is nice. I was wondering about what opportunities might arise for introducing code duplication without noticing. But this structure, I think, actually helps with that, as it's quicker to keep track of what functions exist.

Issue #253, opened 2021-05-25
Exhaustive pattern matching for variants and poly variants.


amiralies:
https://user-images.githubusercontent.com/13261088/119528876-d0c13880-bd96-11eb-8315-4b7f89b8a30a.mp4

Example from scala. I'm not sure how this should be done. scala has advantage of `match` keyworkd coming after target expression so can decide to show that snippet based on that. can we do this with opening brace?
for example 
```rescript
type user = Owner | Admin | User

let isSuperUser = (user: user) => switch user { // show snippet

```

cristianoc:
We have some templates in the extension.
@chenglou would templates give that ^ behaviour?

amiralies:
I don't think if those snippets now about type extensions. it should be done in analysis binary

cristianoc:
For poly variants, see suggestions in https://github.com/rescript-lang/rescript-vscode/issues/283.

Issue #684, opened 2023-01-08
Revamp snippets


zth:
Move producing snippets into the analysis bin so we can make sure they're only produced where it makes sense.

PR #699, opened 2023-01-15
implement exhaustive switch completion

zth:
This is a proposal to solve https://github.com/rescript-lang/rescript-vscode/issues/253.
It's also the first step of https://github.com/rescript-lang/rescript-vscode/issues/684, moving snippets into the analysis bin and only outputting them where they make sense.

exhaustive-switch
(this GIF isn't 100% accurate, I've done some cosmetic fixes to the completion list since recording it)

This PR does the following: When completing.. 
- an _ident_ after a _switch_
- and the switch has no cases
- and the type of the ident we're completing is something we can (reasonably) generate an exhaustive switch for
- ...add a separate completion item for the ident that, when chosen, inserts an exhaustive switch statement for that ident

Closes https://github.com/rescript-lang/rescript-vscode/issues/253


cristianoc (approved):
This  addition is pretty powerful.
Thoughts about the case where one already has a switch, with a catch-all ` | _ => ` and wants to complete the missing cases in place of `_`?

amiralies:
Nice job!

PR #700, opened 2023-01-15
add broken test for pipe completion on aliased types

zth:
Pipe completion does not work on aliased types. Wondering what the best way would be to make it work.

### DIFF
--- a/analysis/src/TypeUtils.ml
+++ b/analysis/src/TypeUtils.ml
@@ -148,6 +148,19 @@ let rec extractType ~env ~package (t : Types.type_expr) =
     Some (Tpolyvariant {env; constructors; typeExpr = t})
   | _ -> None
 
+let rec resolveTypeForPipeCompletion ~env ~package (t : Types.type_expr) =
+  match t.desc with
+  | Tlink t1 | Tsubst t1 | Tpoly (t1, []) ->
+    resolveTypeForPipeCompletion ~env ~package t1
+  (* Don't descend into types named "t". Type t is a convention in the ReScript ecosystem. *)
+  | Tconstr (path, _, _) when path |> Path.last = "t" -> (env, t)
+  | Tconstr (path, _, _) -> (
+    match References.digConstructor ~env ~package path with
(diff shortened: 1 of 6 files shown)


cristianoc:
It needs to start from the type environment and expand the definition.

analysis/src/TypeUtils.ml
```diff
@@ -148,18 +148,90 @@ let rec extractType ~env ~package (t : Types.type_expr) =
...
+    match typFromLoc with
+    | Some typFromLoc ->
+      typFromLoc |> resolveTypeForPipeCompletion ~lhsLoc ~env ~package ~full
+    | None ->
+      let rec digToRelevantType ~env ~package (t : Types.type_expr) =
+        match t.desc with
+        | Tlink t1 | Tsubst t1 | Tpoly (t1, []) ->
+          digToRelevantType ~env ~package t1
+        (* Don't descend into types named "t". Type t is a convention in the ReScript ecosystem. *)
+        | Tconstr (path, _, _) when path |> Path.last = "t" -> (env, TypExpr t)
```
cristianoc, line 226: What happens in practice if this case is omitted?
I was thinking of the different cases. E.g. when it's an alias to `array`, whether one wants to complete for `array`. And when it's a record definition, then one wants to treat is as the source of truth.
Just thinking about here, about what the convention could look like.

cristianoc (approved):
Nice one!

analysis/src/TypeUtils.ml
zth: Yeah, I added this case because it broke 2 tests in `Completion.res`, and looking at them I felt like the old behavior was better.

Yeah, I agree. I don't feel like there's a clear and good answer here, but I do lean a bit towards the current behavior, as in `type t` usually means the main type, and if it's in a module there's often functions in the module to work on it. But I agree it's not 100%. 

One idea could be to complete for the module with the type `t`, _and_ the aliased value. Unsure if that'd be messy to implement, but it might be good behavior. What do you think?

Merging since it doesn't change current behavior.

PR #701, opened 2023-01-18
complete function snippets

zth:
This completes function snippets in positions where a function is expected. I think this is a cool feature that can be quite useful.

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1119,45 +1119,58 @@ let rec completeTypedValue (t : SharedTypes.completionType) ~env ~full ~prefix
           ~env ();
       ]
     else []
-  | Some (Tfunction {env; typ; args}) ->
-    if mode = Pattern then []
-    else
-      let mkFnBody ~asSnippet =
-        match args with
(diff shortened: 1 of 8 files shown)


server/src/server.ts
```diff
@@ -1119,7 +1119,7 @@ function onMessage(msg: p.Message) {
           renameProvider: { prepareProvider: true },
           documentSymbolProvider: true,
           completionProvider: {
-            triggerCharacters: [".", ">", "@", "~", '"', "="],
+            triggerCharacters: [".", ">", "@", "~", '"', "=", "("],
```
zth, line 1122: I added this more as a test. I think we can leave it on in the pre-release version and see what people say. Easy to remove if it's annoying because it triggers completion quite often.

analysis/src/CompletionBackEnd.ml
```diff
@@ -1117,6 +1119,45 @@ let rec completeTypedValue (t : SharedTypes.completionType) ~env ~full ~prefix
           ~env ();
       ]
     else []
+  | Some (Tfunction {env; typ; args}) ->
+    if mode = Pattern then []
```
cristianoc, line 1123: This can be done in a separate match case perhaps? To avoid 1 indentation.

cristianoc (approved):

PR #709, opened 2023-01-21
Fix incorrect syntax in hover help for module

### LINKED ISSUE #708: Incorrect syntax in hover help for module
When hovering over `X` in

```rescript
module A = {
  let x = 42
}
```

the following hover help is shown:

```rescript
module A = {
  let x: int
}
```

However, this is not the correct syntax as it would have to be in the .resi file. There, a `:` would be needed instead of the `=`:

```rescript
module A: {
  let x: int
}
```

cknitt:
Fixes #708

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Make signature help work in calls nested inside of other calls. https://github.com/rescript-lang/rescript-vscode/pull/687
 - Fix pipe completion to work on aliased types. https://github.com/rescript-lang/rescript-vscode/pull/700
 - Fix issue with not finding type for function arguments. https://github.com/rescript-lang/rescript-vscode/pull/706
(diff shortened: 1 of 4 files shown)


zth (commented):
Awesome, looking great! Could we add separate tests that details this exact behavior? I don't see `module type` printed in the tests for example.

cknitt:
`module type` already occurs in Hover.res:
https://github.com/rescript-lang/rescript-vscode/blob/25caa496dc3c0e63dad288a66e009b2009bfb0c5/analysis/tests/src/Hover.res#L50
The output for module type is unchanged, as it was correct before.
So both cases are covered and I'm not sure if we actually need more tests.

zth (approved):

zth:
Changelog would be good, other than that good to go!

Issue #598, opened 2022-10-08
Autocomplete does not take escaped identifiers into account


cknitt:
E.g.

```rescript
module X = {
  let \"Foo" = "foo"
}

let x = X.
```

completes to `X.Foo` instead of `X.\"Foo"`.

cristianoc:
It would be nice to move in the opposite direction, even though it's not so obvious how exactly to do that.
There are cases where upper case ids, or non-standard ids, are necessary, such as for components.

Wondering whether there is a way to legitimise the legitimate use cases without the free-for-all quoted thing which is really not recommended unless there's no other way.

cristianoc:
For example, this
```res
let foo = (x,y) => x+y
```
does two things: it defines an anonymous function, and binds it to variable `foo`.
One could imagine some mechanism to determine the name of the function for `console.log(foo.name)`.

cknitt:
This affects hover, too, not just completion.

Hovering on the above module `X` incorrectly shows

```rescript
module X = {
  let Foo: string
}
```

Escaped identifiers also occur for prop names in both the React Native and the Material-UI bindings, e.g.

```rescript
<Mui.Dialog \"open"=true>
  ...
```

or

```rescript
<FlatList
  \"ItemSeparatorComponent"={_ => <Separator />}
  ...
```

where they are not completed correctly either.

PR #710, opened 2023-01-22
Exotic identifiers

cknitt:
Here are some test cases for #598.

So far, I was able to fix only a part of them and could use help with the rest.

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1090,7 +1090,10 @@ let rec completeTypedValue (t : SharedTypes.completionType) ~env ~full ~prefix
       |> List.filter (fun (field : field) ->
              List.mem field.fname.txt seenFields = false)
       |> List.map (fun (field : field) ->
-             Completion.create field.fname.txt
+             let fname =
+               PrintType.printIdentLike ~allowUident:false field.fname.txt
+             in
+             Completion.create fname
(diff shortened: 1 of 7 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1090,7 +1090,10 @@ let rec completeTypedValue (t : SharedTypes.completionType) ~env ~full ~prefix
       |> List.filter (fun (field : field) ->
              List.mem field.fname.txt seenFields = false)
       |> List.map (fun (field : field) ->
-             Completion.create field.fname.txt
+             let fname =
+               PrintType.printIdentLike ~allowUident:false field.fname.txt
+             in
+             Completion.create fname
```
zth, line 1096: One idea is to not use the `name` field for this, but instead the `insertText` field. That'll make the name appear normal in the completion list, but it's the escaped identifier that'll be insterted. What do you think?

zth: We can also consider whether this should be done inside of `Completion.create` directly. Probably not though, as there are quite a few instances where it shouldn't be escaped even when there are exotic characters in there (like with variant payloads, `Some(whatver)`).

zth (commented):
Looing good. A few small comments. Rest of the work is mostly just about finding all the places this needs to be done at.

PR #711, opened 2023-01-22
[WIP] Annotated completions

zth:
Completion via annotations. E.g: `let x: someVariant = O<com>` completes for the variant `someVariant`, even if the let assignment hasn't compiled previously.

Only handles "basic" types as of now. No handling of options etc.

Currently quite messy, but looking for your feedback before I continue @cristianoc .

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -599,7 +599,7 @@ let completionsGetCompletionType ~full = function
   | _ -> None
 
 let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
-    ~exact ~scope (contextPath : Completable.contextPath) =
+    ~exact ~scope ?(mode = `Regular) (contextPath : Completable.contextPath) =
   let package = full.package in
   match contextPath with
   | CPString ->
@@ -623,13 +623,37 @@ let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
           (Completion.Value
              (Ctype.newconstr (Path.Pident (Ident.create "float")) []));
     ]
-  | CPArray ->
+  | CPArray None ->
     [
       Completion.create "array" ~env
         ~kind:
           (Completion.Value
              (Ctype.newconstr (Path.Pident (Ident.create "array")) []));
     ]
+  | CPArray (Some cp) -> (
+    match mode with
+    | `Regular -> (
+      match
+        cp
+        |> getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos
+             ~env ~exact:true ~scope
+        |> completionsGetCompletionType ~full
+      with
+      | None -> []
+      | Some (typ, env) ->
+        [
+          Completion.create "dummy" ~env
+            ~kind:(Completion.ExtractedType (Tarray (env, typ), `Type));
+        ])
+    | `Pipe ->
+      (* Pipe completion with array just needs to know that it's an array, not
+         what inner type it has. *)
+      [
+        Completion.create "array" ~env
+          ~kind:
+            (Completion.Value
+               (Ctype.newconstr (Path.Pident (Ident.create "array")) []));
+      ])
   | CPOption cp -> (
     match
       cp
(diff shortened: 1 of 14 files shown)


cristianoc (commented):
Looks like it's proceeding well.
Unsure about the naming of some variants, e.g. `ResolvedType`. Perhaps just adding some comments to the definition would clarify.

analysis/src/CompletionBackEnd.ml
```diff
@@ -599,7 +599,7 @@ let completionsGetCompletionType ~full = function
   | _ -> None
 
 let rec getCompletionsForContextPath ~full ~opens ~rawOpens ~allFiles ~pos ~env
-    ~exact ~scope (contextPath : Completable.contextPath) =
+    ~exact ~scope ?(mode = `Regular) (contextPath : Completable.contextPath) =
```
cristianoc, line 602: why poly variants here? ``` `Regular ```

analysis/src/SharedTypes.ml
```diff
@@ -296,6 +294,36 @@ end
 
 type polyVariantConstructor = {name: string; args: Types.type_expr list}
 
+(** An type that can be used to drive completion *)
+type completionType =
```
cristianoc, line 298: How about factoring out `env`? Seems to be present in every case.

analysis/src/CompletionBackEnd.ml
zth: Pure laziness, I'll change it to a proper variant. Reason for having a mode overall is that I found a case where pipe completion breaks using my new approach, that I want to tackle in a separate PR/separate refactor.

analysis/src/CompletionBackEnd.ml
```diff
@@ -582,8 +584,22 @@ let completionsGetTypeEnv = function
   | {Completion.kind = Field ({typ}, _); env} :: _ -> Some (typ, env)
   | _ -> None
 
+let completionsGetCompletionType ~full = function
```
cristianoc, line 587: Every case seems to return the same `env`. Factor out?

analysis/src/SharedTypes.ml
```diff
@@ -296,6 +294,36 @@ end
...
+  | Tuple of QueryEnv.t * Types.type_expr list * Types.type_expr
+  | Toption of QueryEnv.t * completionType
+  | Tbool of QueryEnv.t
+  | Tarray of QueryEnv.t * completionType
```
cristianoc, line 302: Looking at `env` in the nested case, e.g. in the argument to `Tarray`.
It seems to me, from a cursory look, that the inner env might always be the same as the outer env.
If true in general, then factoring out the `env` in the type definition means it does not need to appear in the argument of `Tarray` etc.

I haven't really checked that this is true for all cases. If it is true, then the overall type would be:
`{env; desc}` where `desc` would be the recursive definition of the variant cases. So e.g. `Tarray of desc` would not include the `env`.

cristianoc (approved):
Looking great.
Left some comments about possible further refactoring opportunities.

analysis/src/SharedTypes.ml
zth: Yes, that's a good idea. I'll need to have a deep look at how env is passed around, where it's needed, etc. Ok with you if I do that in a separate PR?

analysis/src/CompletionFrontEnd.ml
```diff
@@ -490,6 +470,58 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
...
+           && CompletionExpressions.isExprHole pvb_expr -> (
+      (* Expression with derivable type annotation, when the expression is empty (expr hole).
+         E.g: let x: someRecord = <com> *)
+      match TypeUtils.contextPathFromCoreType coreType with
+      | Some ctxPath ->
+        setResult
+          (Completable.Cexpression
+             {contextPath = ctxPath; prefix = ""; nested = []})
+      | _ -> ())
+    | {pvb_pat; pvb_expr} when locHasCursor pvb_pat.ppat_loc -> (
```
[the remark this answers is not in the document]
zth: I'm *fairly* sure we're not depending on the ordering, but we're getting dangerously close...

PR #712, opened 2023-01-24
More completion for unsaved code.

zth:
Ready for a first round of feedback @cristianoc. I've commented on a few things I haven't figured out yet. So looking for feedback on that, plus on the general approach.

A surprising amount of quite juicy use cases already work though, which is exciting. You'll see in the tests.

### DIFF
--- a/analysis/src/CompletionFrontEnd.ml
+++ b/analysis/src/CompletionFrontEnd.ml
@@ -136,6 +136,7 @@ let findArgCompletables ~(args : arg list) ~endPos ~posBeforeCursor
          })
   | _ -> loop args
 
+(* TODO: Mimic pipe chain detection here? *)
 let rec exprToContextPath (e : Parsetree.expression) =
   match e.pexp_desc with
   | Pexp_constant (Pconst_string _) -> Some Completable.CPString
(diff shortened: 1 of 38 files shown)


analysis/src/CompletionFrontEnd.ml
```diff
@@ -768,15 +913,23 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
                 ~firstCharBeforeCursorNoWhite
             | None -> None
           in
-
-          setResultOpt argCompletable
+          match argCompletable with
+          | None -> (
+            match exprToContextPath funExpr with
```
cristianoc, line 918: Looks like this very same pattern match is done a few lines above.

analysis/src/CompletionFrontEnd.ml
```diff
@@ -804,8 +957,16 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
...
```
cristianoc, line 962: Same match a few lines above.
Also, this case looks nearly identical to the one above for `"|."`.

cristianoc:
Need to rebase on master.

cristianoc (approved):
Go go go

Issue #714, opened 2023-01-25
Add `?` when destructuring for optional fields


zth:
See discussion in https://github.com/rescript-lang/rescript-compiler/issues/5953

PR #715, opened 2023-01-25
special handling of optional fields when destructuring

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/714

cc @cknitt

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -1047,7 +1047,7 @@ let printConstructorArgs argsLen ~asSnippet =
   if List.length !args > 0 then "(" ^ (!args |> String.concat ", ") ^ ")"
   else ""
 
-type completionMode = Pattern of Completable.patternMode option | Expression
+type completionMode = Pattern of Completable.patternMode | Expression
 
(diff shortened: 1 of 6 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -1047,10 +1047,10 @@ let printConstructorArgs argsLen ~asSnippet =
   if List.length !args > 0 then "(" ^ (!args |> String.concat ", ") ^ ")"
   else ""
 
-type completionMode = Pattern | Expression
+type completionMode = Pattern of Completable.patternMode option | Expression
```
cknitt, line 1050: I am wondering if we can get rid of the `option` here by having

```ocaml
type patternMode = Default | Destructuring
```

or something like that?

cristianoc (commented):
Does this affect destructuring only, or also explicit switches?

zth:
Right now it's configured to only affect let destructures.

cristianoc (approved):

analysis/src/CompletionBackEnd.ml
cristianoc: Good point.

PR #718, opened 2023-01-31
Fix jsx wrapping in completion

zth:
Didn't account for nested values when deciding when to insert braces for JSX prop values when completing.


cristianoc (approved):
Good catch.

PR #719, opened 2023-01-31
Complete lowercase JSX elements

zth:
This adds completion for lowercase JSX elements, treating them like HTML elements.

Potential follow up for this - there's a heuristic in the hover mechanism that detects lowercase elements. Maybe we should leverage this new list of elements and descriptions in there too? It currently shows some fairly irrelevant things.

image


analysis/src/CompletionJsx.ml
```diff
@@ -487,6 +487,213 @@ let domLabels =
     ("suppressContentEditableWarning", bool);
   ]
 
+(* List and explanations taken from
+   https://www.tutorialrepublic.com/html-reference/html5-tags.php. *)
+let htmlElements =
+  [
+    ("a", "Defines a hyperlink.", false);
+    ("abbr", "Defines an abbreviated form of a longer word or phrase.", false);
```
cristianoc, line 495: How does this compare with the complete list currently supported? A subset? (E.g. aria).

cristianoc (approved):

analysis/src/CompletionJsx.ml
zth: Hmmm, not sure I follow what you mean..?

zth: No I think the type checker would accept anything, I think the JSX tag name is just turned into a string by the PPX. So this is strictly about providing completion/information for something that we don't really have (or care) about type information for right now. So it's strictly an addon/extra help.

Issue #707, opened 2023-01-21
JSX v4 auto completion


dkirchhof:
Hello,

I replaced the language server of my rescript-vim plugin to the latest one from this vs code plugin.
Unfortunatelly I have problems with the auto completion for props of a custom component.

In your "screenshot" (https://forum.rescript-lang.org/t/call-for-testing-more-autocomplete/4078#completion-for-jsx-prop-values-2) everything looks fine. 

In my editor it doesn't work:
![image](https://user-images.githubusercontent.com/2778643/213866104-ee5cb3e5-7213-442a-b982-8a3546c9aaff.png)

In normal html tags the auto completion works as expected:
![image](https://user-images.githubusercontent.com/2778643/213866333-a0820fae-0cf7-48bc-9bfa-796412835402.png)

Setup:
editor: neovim
rescript version: 10.1.0
server version: 1.10.0
jsx version: 4
jsx mode: automatic

zth:
I guess this is because you're not using the `@react.component` annotation. What happens when you try using?

We should probably support this pattern too now with JSX v4.

dkirchhof:
You're right. With `@react.component` and labeled arguments the auto completion shows all props (`text` and `key`).

zth:
Thanks. We'll look into supporting this case as well.

zth:
To be clear - this isn't actually about JSX v4, just that this pattern is now more expected than before. `@react.component` is still the recommended way with JSX v4, so this is about supporting this "alternate" pattern.

dkirchhof:
Oh, didn't know that it was possible before. Thanks for clarification.

PR #720, opened 2023-01-31
enable completions for handwritten JSX components

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/707

A bit unsure if this is the best approach. What do you think @cristianoc ?

### DIFF
--- a/analysis/src/CompletionJsx.ml
+++ b/analysis/src/CompletionJsx.ml
@@ -559,6 +559,19 @@ let getJsxLabels ~componentPath ~findTypeOfValue ~package =
              && Path.last path = "props" ->
         (* JSX V4 external or interface *)
         getFieldsV4 ~path ~typeArgs
+      | Tarrow (Nolabel, typ, _, _) -> (
+        (* Component without the JSX PPX, like a make fn taking a hand-written
+           type props. *)
+        let rec digToConstr typ =
+          match typ.Types.desc with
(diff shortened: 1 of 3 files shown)


analysis/src/CompletionJsx.ml
```diff
@@ -559,6 +559,19 @@ let getJsxLabels ~componentPath ~findTypeOfValue ~package =
              && Path.last path = "props" ->
         (* JSX V4 external or interface *)
         getFieldsV4 ~path ~typeArgs
+      | Tarrow (Nolabel, typ, _, _) -> (
+        (* Component without the JSX PPX, like a make fn taking a hand-written
```
cristianoc, line 563: Should one check anything about the return type? Or can it be anything?

cristianoc (approved):

analysis/src/CompletionJsx.ml
zth: We could check the return type for sure but I'm unsure if it really matters that much. Strictly though it needs to be `React.element`, or whatever JSX element is configured (not sure if it can be configured now, but it'll be possible later). For this to error, you'd need to use a JSX component with a make function in a module that takes a type named props, and return something else then React.element (which will error in compilation anyway).

I say we go without it for this case, and then fix it if need be?

cristianoc: Sounds good

PR #722, opened 2023-02-01
fall back to local completions when expr completion misses

zth:
This fixes a bug I introduced where we wouldn't include local completions in enough cases of typed expressions.

It essentially adds regular, local completions as fallback completions when typed expression completion misses for various reasons.


zth:
@cristianoc I missed that this added a ton of test output. Will clean up.

cristianoc (approved):

PR #726, opened 2023-02-05
treat signature help as a stable feature

zth:
This changes signature help to be a stable feature, turned on by default.


package.json
```diff
@@ -158,10 +158,10 @@
 					"default": false,
 					"description": "Enable (experimental) code lens for function definitions."
 				},
-				"rescript.settings.signatureHelp.enable": {
+				"rescript.settings.signatureHelp.enabled": {
```
zth, line 161: Changing the name of this setting so it resets for all users, and defaults to true.

cristianoc (approved):

Issue #354, opened 2022-01-27
[Feature] Publish analysis and  server  to npm


luooooob:
Are the analysis and server in vim-rescript copied and pasted from this project?

I think the "/analysis" and "/server" should be put in a "rescript-lsp-server" repository (or use monorepo ?) and published to npm, like [ocaml-lsp-server](https://github.com/ocaml/ocaml-lsp). It would be easier for people to develop plugins for more IDEs that support lsp

Do you have such a plan？

amiralies:
I think we are still not sure about this. lsp and analysis binary are implementation details for both vscode and vim extensions.
But I'm +1 for this. if we release an standalone language server we can have better support for other editors.

gasacchi:
+1 for this, is hard to install rescript language server when i want to use rescript in another machine, since i don't use (neo)vim / vscode & i use helix editor daily

zth:
Is the hard part that it's hard to access the lsp server and related binaries? If so, is it sure that an npm package would help with that, and would it help for all of the editors we want to support.

We build a vsix for VSCode for each release, and I'm pretty sure that's just a regular zip in disguise. I imagine downloading the latest vsix, unzipping and moving the server in place would be something that could be tried today as an experiment of whether it's doable, rather than publish a separate package.

gasacchi:
yes accessing lsp server and related binaries each time when changing machine that doesn't have it mean setting up $PATH or changing editor configuration file (in this case Helix) over again. npm package definitely help.

aspeddro:
Any plans for this?

I can work on it.

cristianoc:
@zth any thoughts?
Somehow related, people have reported about using the analysis in CI. That requires at least exposing the analysis binary (from which reanalyze is accessible as one of the commands).

zth:
I think it's a good idea, especially now that reanalyze lives in it too. But I think we need to think carefully about what parts of it we communicate is stable in it, to set the expectations right. So we don't end up in a situation where we can't make the changes we'd like without risking breaking things in an unexpected way.

aspeddro:
I think analysis should not be exposed. We should publish just one binary node to start the server, like `rescriptlsp`. In this case, the LSP is no longer an implementation detail.

PR #734, opened 2023-02-17
Publish LSP to npm

aspeddro:
- Publish a binary to npm as `rescriptls`. Support two channels `stdio` and `node-ipc`
- VSCode client keeps using `node-ipc` channel

Related #354


zth:
Great stuff! 

One thing I'd like to mull over a bit before going this route is that there might be a separate case for publishing just the analysis binary to npm, by itself. That would be because the analysis binary might in the future contain more general tooling, not just stuff for powering the LSP. Doc extraction is one example where it's possible that's the easiest way of distributing it.

This package will have the analysis bin too of course. So maybe there's a way to name this (and expose the analysis bin separately) so that they can be exposed via the same package, without it being fantastically confusing.. Let me think a bit. Any feedback welcome.

cristianoc:
Perhaps only expose the part of the analysis binary which is for public consumption. Whiteout even leaking the existence of the binary (helps keeping freedom to move stuff between binary and ts).

aspeddro:
If I understand correctly, as reported by users the main problem is that we don't publish an LSP like OCaml does to facilitate the setup in several editors. Just publishing the analysis binary does not resolve the reported issue.

> yes accessing lsp server and related binaries each time when changing machine that doesn't have it mean setting up $PATH or changing editor configuration file (in this case Helix) over again. npm package definitely help.
[Comment](https://github.com/rescript-lang/rescript-vscode/issues/354#issuecomment-1207222732)

One idea is to do what [`rust-analyzer`](https://github.com/rust-lang/rust-analyzer/blob/a6603fc21d50b3386a488c96225b2d1fd492e533/crates/rust-analyzer/src/cli/flags.rs#L9-L117) does. We publish a single binary with subcommands.

- `rescript-toolchain lsp [args]` - Start LSP
- `rescript-toolchain analysis [args]` - Call OCaml `analysis binary` commands

aspeddro:
Ready for review.

aspeddro:
The first version should be `0.0.1` or `1.20`?

zth (approved):
Awesome! I think `1.20.0` for the version is right, it's nice to have the same as the actual extension.

Fantastic to get this out.

PR #737, opened 2023-02-27
prettify optional fields in completion detail

zth:
This prettifies how optional fields are printed in the completion detail.

Before:
```rescript
someOptionalField: option<int>
```

After:
```rescript
someOptionalField?: int
```

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -221,7 +221,15 @@ let kindToDetail name (kind : Completion.kind) =
   | Label typString -> typString
   | Module _ -> "module"
   | FileModule _ -> "file module"
-  | Field ({typ}, s) -> name ^ ": " ^ (typ |> Shared.typeToString) ^ "\n\n" ^ s
+  | Field ({typ; optional}, s) ->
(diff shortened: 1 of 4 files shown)


analysis/src/CompletionBackEnd.ml
```diff
@@ -221,7 +221,15 @@ let kindToDetail name (kind : Completion.kind) =
   | Label typString -> typString
   | Module _ -> "module"
   | FileModule _ -> "file module"
-  | Field ({typ}, s) -> name ^ ": " ^ (typ |> Shared.typeToString) ^ "\n\n" ^ s
+  | Field ({typ; optional}, s) ->
+    (* Handle optional fields. Checking for "?" is because sometimes optional
+       fields are prefixed with "?" when completing, and at that point we don't
+       need to _also_ add a "?" after the field name, as that looks weird. *)
+    if optional && Utils.startsWith name "?" = false then
```
cristianoc, line 228: nit: little code duplication in the if and else branch, perhaps better to unify, perhaps not

cristianoc (approved):

analysis/src/CompletionBackEnd.ml
zth: I was being goal oriented 😉

PR #741, opened 2023-03-07
[Test] add @rescript/core to the tests.

cristianoc:
Just to see how things look like.
One comment is that the local file `Object.res` had to be renames as otherwise it would become unreachable. This is a side-effect of doing `-open RescriptCore` globally, which overrides the local scope. This won't be a problem when the library is integrated in the compiler.


analysis/tests/src/expected/CompletionFunctionArguments.res.txt
```diff
@@ -33,12 +37,20 @@ Completable: Cexpression CArgument Value[someFn](~isOn)=t
     "tags": [],
     "detail": "bool",
     "documentation": null
+  }, {
+    "label": "typeof",
+    "kind": 12,
+    "tags": [],
+    "detail": "'a => Core__Type.t",
```
cristianoc, line 44: Notice `Core__Typet` is leaked.

cristianoc: CC @zth

zth: Yeah, I think we'll need to special case Core. Good catch.

PR #749, opened 2023-03-11
leverage new completion features to provide hover on (some) unsaved code

zth:
This leverages the new unsaved completion code to get hovers on some unsaved code.

I would eventually like this to also work on assigned identifiers, like (all unsaved):
```rescript
let fff = makeSomeRecord()
//   ^hov should show `someRecord`, but it doesn't because it doesn't handle the actual assigned identifier
let ccc = fff
//               ^hov here it works though

```

But this is a win nonetheless so adding support for something like that can be a separate thing.

### DIFF
--- a/analysis/src/Hover.ml
+++ b/analysis/src/Hover.ml
@@ -125,7 +125,11 @@ let getHoverViaCompletions ~debug ~path ~pos ~currentFile ~forHover
     ~supportsMarkdownLinks =
   match Completions.getCompletions ~debug ~path ~pos ~currentFile ~forHover with
   | None -> None
-  | Some (completions, {file; package}) -> (
+  | Some (completions, ({file; package} as full), scope) -> (
+    let rawOpens = Scope.getRawOpens scope in
+    let allFiles =
+      FileSet.union package.projectFiles package.dependenciesFiles
+    in
     match completions with
     | {kind = Label typString; docstring} :: _ ->
       let parts =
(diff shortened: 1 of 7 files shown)


analysis/src/Hover.ml
```diff
@@ -125,31 +125,44 @@ let getHoverViaCompletions ~debug ~path ~pos ~currentFile ~forHover
     ~supportsMarkdownLinks =
   match Completions.getCompletions ~debug ~path ~pos ~currentFile ~forHover with
   | None -> None
-  | Some (completions, {file; package}) -> (
+  | Some (completions, ({file; package} as full), scope) -> (
+    let rawOpens = Scope.getRawOpens scope in
+    let allFiles =
```
cristianoc, line 130: Nit: there's a bit of redundancy.
`allFiles` comes from `package`.
There are functions that take both `allFiles` and `package` (or `full`). So they' don't really need all those args.
I guess `allFiles` can be either a function of `package`, or directly a field there. So it's computed on the spot when needed instead of being passed around.

analysis/src/Hover.ml
```diff
@@ -125,31 +125,44 @@ let getHoverViaCompletions ~debug ~path ~pos ~currentFile ~forHover
...
         (if typString = "" then [] else [Markdown.codeBlock typString])
         @ docstring
       in
       Some (Protocol.stringifyHover (String.concat "\n\n" parts))
-    | {kind = Field _; docstring} :: _ -> (
-      match CompletionBackEnd.completionsGetTypeEnv completions with
+    | {kind = Field _; env; docstring} :: _ -> (
+      let opens = CompletionBackEnd.getOpens ~debug ~rawOpens ~package ~env in
+      match
+        CompletionBackEnd.completionsGetTypeEnv2 ~debug ~full ~rawOpens ~opens
```
cristianoc, line 143: I had not noticed much the business of `completionsGetTypeEnv` vs `completionsGetTypeEnv2`. Is there ever a reason for using the former given the latter is more powerful?

cristianoc (approved):

analysis/src/Hover.ml
zth: Will fix in separate PR.

analysis/src/Hover.ml
zth: Probably not. Mostly a bridge for migrating to the new one. Will take care of this as we do our major refactor of all of this new functionality.

PR #752, opened 2023-03-22
Open Pervasives.JsxModules by default.

cristianoc:
See https://github.com/rescript-lang/rescript-compiler/pull/6091 where the jsx modules are moved inside Pervasives, to provide 2 different copies for curried and uncurried mode.


cristianoc:
One still gets e.g. `JsxC` and `JsxU` in the autocompletion list, just because they exist.
@zth I guess there are other similar issues elsewhere. I was wondering whether one could add some sort of annotation to some modules so they don't show up in autocomplete.

PR #763, opened 2023-04-25
Explore emit function token

aspeddro:


aspeddro:
Limitations: Does not highlight functions `b` and `f`.

The second case can be done through static analysis. The first requires type information.

image

PR #769, opened 2023-04-30
Use uncurried mode setting for outcome printer

cristianoc:
On hover,  uncurried type will now show without dot (in uncurried mode).

### DIFF
--- a/analysis/src/Packages.ml
+++ b/analysis/src/Packages.ml
@@ -34,7 +34,6 @@ let newBsPackage ~rootPath =
                Option.bind ns Json.bool
              in
              let uncurried = uncurried = Some true in
-             if uncurried then Config.uncurried := Uncurried;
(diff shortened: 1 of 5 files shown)


analysis/src/Packages.ml
```diff
@@ -29,6 +29,12 @@ let newBsPackage ~rootPath =
         | Some libBs ->
           Some
             (let namespace = FindFiles.getNamespace config in
+             let uncurried =
+               let ns = config |> Json.get "uncurried" in
+               Option.bind ns Json.bool
+             in
+             let uncurried = uncurried = Some true in
+             if uncurried then Config.uncurried := Uncurried;
```
cristianoc, line 37: This is a bit crude: triggered whenever a config file is read.
Should probably try to use the current config in the commands instead.

zth (approved):

PR #771, opened 2023-05-21
Print uncurried

zth:
- Vendors latest syntax stuff.
- Add tests for and fix various issues around finding and using uncurried functions.


zth:
Signature help is broken for uncurried too, but will fix in separate PR.

cristianoc (approved):
Sweet!

PR #774, opened 2023-05-23
Add syntax highlighting for raw JS code

fhammerschmidt:
image

image

implements https://github.com/rescript-lang/rescript-compiler/issues/6214

### DIFF
--- a/grammars/rescript.tmLanguage.json
+++ b/grammars/rescript.tmLanguage.json
@@ -511,46 +511,49 @@
   },
   "patterns": [
     {
-      "include": "#storage"
-    },
-    {
-      "include": "#ffi"
-    },
-    {
-      "include": "#constant"
-    },
-    {
-      "include": "#commentLine"
-    },
-    {
-      "include": "#commentBlock"
-    },
-    {
-      "include": "#character"
-    },
-    {
-      "include": "#typeParameter"
-    },
-    {
-      "include": "#string"
-    },
-    {
       "include": "#attribute"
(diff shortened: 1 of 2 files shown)


fhammerschmidt:
Note: I did not implement it for `""` strings, because I think that would break too easily.

zth (approved):
Nice! Mind adding to the changelog?

cristianoc:
Actually would you do it for `%ffi`?

zth:
> Actually would you do it for `%ffi`? Notice single `%`

Judging from the screenshot this already works, no?

cristianoc:
> > Actually would you do it for `%ffi`? Notice single `%`
> 
> 
> 
> Judging from the screenshot this already works, no?

Oh right I did not see the second one.

grammars/rescript.tmLanguage.json
```diff
@@ -481,12 +481,41 @@
...
+        }
+      ]
     }
   },
   "patterns": [
     {
       "include": "#storage"
     },
+    {
+      "include": "#ffi"
```
cristianoc, line 517: Should we alphabetize these?

cristianoc (approved):
This is great thanks!

grammars/rescript.tmLanguage.json
fhammerschmidt: I guess it does not hurt. Will do it.

cristianoc (approved):

PR #777, opened 2023-05-26
Tests: use the latest beta.

cristianoc:


cristianoc:
@zth this uses the latest beta, which has uncurried on by default.
The tests show things that still need to be taken care of.

cristianoc:
For a set of changes excluding uncurried, see here: https://github.com/rescript-lang/rescript-vscode/pull/778

zth:
I assume most of these have to do with us not accounting for not setting `uncurried` at all in `bsconfig.json` now implicitly means `uncurried: true`, where it used to mean `uncurried: false`?

cristianoc:
> I assume most of these have to do with us not accounting for not setting `uncurried` at all in `bsconfig.json` now implicitly means `uncurried: true`, where it used to mean `uncurried: false`?

Good point. I was not thinking about that. Maybe that's just it.

PR #778, opened 2023-05-26
Turn off uncurried

cristianoc:
Based on https://github.com/rescript-lang/rescript-vscode/pull/777 but where unhurried is turned off.

### DIFF
--- a/analysis/tests/src/expected/Completion.res.txt
+++ b/analysis/tests/src/expected/Completion.res.txt
@@ -370,7 +370,7 @@ Path Dep.c
     "label": "customDouble",
     "kind": 12,
     "tags": [1],
-    "detail": "int => int",
+    "detail": "(. int) => int",
     "documentation": {"kind": "markdown", "value": "Deprecated: Use customDouble instead\n\nSome doc comment"}
   }]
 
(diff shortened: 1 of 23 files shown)


analysis/tests/src/expected/Completion.res.txt
```diff
@@ -918,13 +918,13 @@ Path Js.n
     "label": "nullable",
     "kind": 22,
     "tags": [],
-    "detail": "type nullable<+'a>",
-    "documentation": {"kind": "markdown", "value": "\n  A value of this type can be undefined, null or 'a. This type is equivalent to Js.Null_undefined.t.\n"}
+    "detail": "type nullable<'a> = Value('a) | Null | Undefined",
```
cristianoc, line 921: @zth we lost the `+` here.

zth: Right, that's... weird. Don't think we changed the `+` in the implementation...

cristianoc: Looks like this is the intended printing, inherited from OCaml, which only prints the `+` when the type is abstract:

```ocaml
# type 'a t1;;
type 'a t1
# type +'a t2;;
type +'a t2
# type +'a t2 = | A;;
type 'a t2 = A
```

zth: Interesting! That's good to know 😀

PR #780, opened 2023-06-01
[Poc]: Doc extraction continuation (PR 732)

aspeddro:
This PR add more information to output:

1. Emit an `optional` field for optional fields in records.
2. Emit a `location` field for the location of the function, module or type in the source file.
3. Emit two field when `@deprecated` or `@@deprecated` attribute is used. 
   - `deprecated` is a bool
   - `deprecatedMessage` is a string.

Continuation of #732


aspeddro:
I removed the field `location`

PR #796, opened 2023-07-27
[DocGen]: Polish

aspeddro:
Emit all items from module alias.

```json
{
    "id": "RescriptCore.Core__Console",
    "kind": "moduleAlias",
    "docstrings": [],
    "signature": "module Console = Core__Console"
}, 
```

New Output:

```json
{
    "id": "RescriptCore.Core__Console",
    "name": "Console",
    "kind": "moduleAlias",
    "docstrings": [],
    "items": [
    {
      "id": "RescriptCore.Core__Console.assert_",
      "kind": "value",
      "name": "assert_",
      "signature": "let assert_: (bool, 'a) => unit",
      "docstrings": ["`assert_(assertion, value)` print a message to console if `assertion` evaluates `false`. Does nothing if it's `true`.\n\nSee [`console.assert`](https://developer.mozilla.org/en-US/docs/Web/API/console/assert)\non MDN.\n\n## Examples\n\n```rescript\nConsole.assert_(false, \"Hello World!\")\nConsole.assert_(n == 42, \"The answer\")\n```"]
    }, 
    {
      "id": "RescriptCore.Core__Console.assert2",
      "kind": "value",
      "name": "assert2",
      "signature": "let assert2: (bool, 'a, 'b) => unit",
      "docstrings": ["`assert2(v1, v2)`. Like `assert_`, but with two arguments.\n\n## Examples\n\n```rescript\nConsole.assert2(false, \"Hello\", \"World\")\nConsole.assert2(n == 42, [1, 2, 3], '4')\n```"]
    },
   ....
}
```

Full output for `ReScriptCore`: [Gist](https://gist.github.com/aspeddro/21bc14975b9b41aa9e465fce0c83bed6)

### DIFF
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -229,8 +229,6 @@ let typeDetail typ ~env ~full =
          })
   | _ -> None
 
-exception Invalid_file_type
-
 let makeId modulePath ~identifier =
   identifier :: modulePath |> List.rev |> SharedTypes.ident
 
@@ -239,7 +237,10 @@ let extractDocs ~path ~debug =
   if
     FindFiles.isImplementation path = false
     && FindFiles.isInterface path = false
-  then raise Invalid_file_type;
+  then (
+    Printf.printf "error: failed to read %s, expected an .res or .resi file\n"
+      path;
+    exit 1);
   let path =
     if FindFiles.isImplementation path then
       let pathAsResi =
@@ -255,7 +256,10 @@ let extractDocs ~path ~debug =
     else path
   in
   match Cmt.loadFullCmtFromPath ~path with
-  | None -> ()
+  | None ->
+    Printf.printf
+      "error: failed to generate doc for %s, try to build the project\n" path;
+    exit 1
   | Some full ->
     let file = full.file in
     let structure = file.structure in
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -165,7 +165,6 @@ let rec stringifyDocItem ?(indentation = 0) ~originalEnv (item : docItem) =
     stringifyObject ~startOnNewline:true ~indentation
       [
         ("id", Some (wrapInQuotes m.id));
-        ("name", Some (wrapInQuotes m.name));
         ("kind", Some (wrapInQuotes "moduleAlias"));
         ("name", Some (wrapInQuotes m.name));
         ("docstrings", Some (stringifyDocstrings m.docstring));
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -304,7 +304,9 @@ let extractDocs ~path ~debug =
                  | Module (Ident p) ->
                    (* module Whatever = OtherModule *)
                    let aliasToModule = p |> pathIdentToString in
-                   let modulePath = aliasToModule :: modulePath in
+                   let id =
+                     (modulePath |> List.rev |> List.hd) ^ "." ^ item.name
+                   in
                    let items =
                      match
                        ProcessCmt.fileForModule ~package:full.package
@@ -313,14 +315,14 @@ let extractDocs ~path ~debug =
                      | None -> []
                      | Some file ->
                        let docs =
-                         extractDocsForModule ~modulePath file.structure
+                         extractDocsForModule ~modulePath:[id] file.structure
                        in
                        docs.items
                    in
                    Some
                      (ModuleAlias
                         {
-                          id = modulePath |> List.rev |> SharedTypes.ident;
+                          id;
                           name = item.name;
                           items;
                           docstring = item.docstring |> List.map String.trim;
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -70,13 +70,13 @@ let stringifyDetail ?(indentation = 0) (detail : docItemDetail) =
     stringifyObject ~startOnNewline:true ~indentation
       [
         ("kind", Some (wrapInQuotes "record"));
-        ( "fieldDocs",
+        ( "items",
           Some
             (fieldDocs
             |> List.map (fun fieldDoc ->
                    stringifyObject ~indentation:(indentation + 1)
                      [
-                       ("fieldName", Some (wrapInQuotes fieldDoc.fieldName));
+                       ("name", Some (wrapInQuotes fieldDoc.fieldName));
                        ( "deprecated",
                          match fieldDoc.deprecated with
                          | Some d -> Some (wrapInQuotes d)
@@ -92,14 +92,14 @@ let stringifyDetail ?(indentation = 0) (detail : docItemDetail) =
     stringifyObject ~startOnNewline:true ~indentation
       [
         ("kind", Some (wrapInQuotes "variant"));
-        ( "constructorDocs",
+        ( "items",
           Some
             (constructorDocs
             |> List.map (fun constructorDoc ->
                    stringifyObject ~startOnNewline:true
                      ~indentation:(indentation + 1)
                      [
-                       ( "constructorName",
+                       ( "name",
                          Some (wrapInQuotes constructorDoc.constructorName) );
                        ( "deprecated",
                          match constructorDoc.deprecated with
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -6,13 +6,6 @@ type constructorDoc = {
   signature: string;
 }
 
-type docsForModuleAlias = {
-  id: string;
-  docstring: string list;
-  name: string;
-  signature: string;
-}
-
 type docItemDetail =
   | Record of {fieldDocs: fieldDoc list}
   | Variant of {constructorDocs: constructorDoc list}
@@ -32,7 +25,12 @@ type docItem =
           (** Additional documentation for constructors and record fields, if available. *)
     }
   | Module of docsForModule
-  | ModuleAlias of docsForModuleAlias
+  | ModuleAlias of {
+      id: string;
+      docstring: string list;
+      name: string;
+      items: docItem list;
+    }
 and docsForModule = {
   id: string;
   docstring: string list;
@@ -137,9 +135,15 @@ let rec stringifyDocItem ?(indentation = 0) ~originalEnv (item : docItem) =
     stringifyObject ~startOnNewline:true ~indentation
       [
         ("id", Some (wrapInQuotes m.id));
+        ("name", Some (wrapInQuotes m.name));
         ("kind", Some (wrapInQuotes "moduleAlias"));
         ("docstrings", Some (stringifyDocstrings m.docstring));
-        ("signature", Some (m.signature |> Json.escape |> wrapInQuotes));
+        ( "items",
+          Some
+            (m.items
+            |> List.map
+                 (stringifyDocItem ~originalEnv ~indentation:(indentation + 1))
+            |> array) );
       ]
 
 and stringifyDocsForModule ?(indentation = 0) ~originalEnv (d : docsForModule) =
@@ -257,14 +261,24 @@ let extractDocs ~path ~debug =
                    (* module Whatever = OtherModule *)
                    let aliasToModule = p |> pathIdentToString in
                    let modulePath = aliasToModule :: modulePath in
+                   let items =
+                     match
+                       ProcessCmt.fileForModule ~package:full.package
+                         aliasToModule
+                     with
+                     | None -> []
+                     | Some file ->
+                       let docs =
+                         extractDocsForModule ~modulePath file.structure
+                       in
+                       docs.items
+                   in
                    Some
                      (ModuleAlias
                         {
                           id = modulePath |> List.rev |> SharedTypes.ident;
-                          signature =
-                            Printf.sprintf "module %s = %s" item.name
-                              aliasToModule;
                           name = item.name;
+                          items;
                           docstring = item.docstring |> List.map String.trim;
                         })
                  | Module (Structure m) ->
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -267,7 +267,7 @@ let extractDocs ~path ~debug =
     let rec extractDocsForModule ?(modulePath = [env.file.moduleName])
         (structure : Module.structure) =
       {
-        id = modulePath |> ident;
+        id = modulePath |> List.rev |> ident;
         docstring = structure.docstring |> List.map String.trim;
         name = structure.name;
         deprecated = structure.deprecated;
--- a/analysis/src/ProcessAttributes.ml
+++ b/analysis/src/ProcessAttributes.ml
@@ -5,7 +5,7 @@ let rec findDocAttribute attributes =
   let open Parsetree in
   match attributes with
   | [] -> None
-  | ( {Asttypes.txt = "ocaml.doc" | "ns.doc" | "res.doc"},
+  | ( {Asttypes.txt = "ocaml.doc" | "ocaml.text" | "ns.doc" | "res.doc"},
       PStr
         [
           {
--- a/analysis/src/ProcessCmt.ml
+++ b/analysis/src/ProcessCmt.ml
@@ -593,9 +593,11 @@ and forStructure ~name ~env strItems =
       strItems []
   in
   let attributes =
-    match strItems with
-    | {str_desc = Tstr_attribute attribute} :: _ -> [attribute]
-    | _ -> []
+    strItems
+    |> List.filter_map (fun (struc : Typedtree.structure_item) ->
+           match struc with
+           | {str_desc = Tstr_attribute attr} -> Some attr
+           | _ -> None)
   in
   let docstring = attrsToDocstring attributes in
   let deprecated = ProcessAttributes.findDeprecatedAttribute attributes in
--- a/analysis/tests/src/expected/DocExtraction2.res.txt
+++ b/analysis/tests/src/expected/DocExtraction2.res.txt
@@ -21,7 +21,7 @@ preferring found resi file for impl: src/DocExtraction2.resi
     "docstrings": ["Makerz of stuffz."]
   }, 
   {
-    "id": "InnerModule.DocExtraction2",
+    "id": "DocExtraction2.InnerModule",
     "name": "InnerModule",
     "kind": "module",
     "items": [
--- a/analysis/tests/src/expected/DocExtraction2.resi.txt
+++ b/analysis/tests/src/expected/DocExtraction2.resi.txt
@@ -20,7 +20,7 @@ extracting docs for src/DocExtraction2.resi
     "docstrings": ["Makerz of stuffz."]
   }, 
   {
-    "id": "InnerModule.DocExtraction2",
+    "id": "DocExtraction2.InnerModule",
     "name": "InnerModule",
     "kind": "module",
     "items": [
--- a/analysis/tests/src/expected/DocExtractionRes.res.txt
+++ b/analysis/tests/src/expected/DocExtractionRes.res.txt
@@ -95,7 +95,6 @@ extracting docs for src/DocExtractionRes.res
     "items": [
     {
       "id": "DocExtractionRes.AnotherModule.SomeInnerModule",
-      "name": "LinkedModule",
       "kind": "moduleAlias",
       "name": "LinkedModule",
       "docstrings": ["This links another module. Neat."],
--- a/analysis/tests/src/expected/DocExtractionRes.res.txt
+++ b/analysis/tests/src/expected/DocExtractionRes.res.txt
@@ -14,13 +14,13 @@ extracting docs for src/DocExtractionRes.res
     "detail": 
     {
       "kind": "record",
-      "fieldDocs": [{
-        "fieldName": "name",
+      "items": [{
+        "name": "name",
         "optional": false,
         "docstrings": ["The name of the stuff."],
         "signature": "string"
       }, {
-        "fieldName": "online",
+        "name": "online",
         "optional": false,
         "docstrings": ["Whether stuff is online."],
         "signature": "bool"
@@ -55,19 +55,19 @@ extracting docs for src/DocExtractionRes.res
       "detail": 
       {
         "kind": "variant",
-        "constructorDocs": [
+        "items": [
         {
-          "constructorName": "Started",
+          "name": "Started",
           "docstrings": ["If this is started or not"],
           "signature": "Started(t)"
         }, 
         {
-          "constructorName": "Stopped",
+          "name": "Stopped",
           "docstrings": ["Stopped?"],
           "signature": "Stopped"
         }, 
         {
-          "constructorName": "Idle",
+          "name": "Idle",
           "docstrings": ["Now idle."],
           "signature": "Idle"
         }]
@@ -95,10 +95,11 @@ extracting docs for src/DocExtractionRes.res
     "items": [
     {
       "id": "DocExtractionRes.AnotherModule.SomeInnerModule",
+      "name": "LinkedModule",
       "kind": "moduleAlias",
       "name": "LinkedModule",
       "docstrings": ["This links another module. Neat."],
-      "signature": "module LinkedModule = SomeInnerModule"
+      "items": []
     }, 
     {
       "id": "DocExtractionRes.AnotherModule.callback",
@@ -123,9 +124,9 @@ extracting docs for src/DocExtractionRes.res
       "detail": 
       {
         "kind": "variant",
-        "constructorDocs": [
+        "items": [
         {
-          "constructorName": "SomeStuff",
+          "name": "SomeStuff",
           "docstrings": ["This has inline records..."],
           "signature": "SomeStuff"
         }]
--- a/analysis/tests/src/expected/DocExtractionRes.res.txt
+++ b/analysis/tests/src/expected/DocExtractionRes.res.txt
@@ -94,7 +94,7 @@ extracting docs for src/DocExtractionRes.res
     "kind": "module",
     "items": [
     {
-      "id": "DocExtractionRes.AnotherModule.SomeInnerModule",
+      "id": "DocExtractionRes.LinkedModule",
       "kind": "moduleAlias",
       "name": "LinkedModule",
       "docstrings": ["This links another module. Neat."],
--- a/analysis/tests/src/expected/DocExtractionRes.res.txt
+++ b/analysis/tests/src/expected/DocExtractionRes.res.txt
@@ -42,7 +42,7 @@ extracting docs for src/DocExtractionRes.res
     "docstrings": ["Stuff goes offline."]
   }, 
   {
-    "id": "SomeInnerModule.DocExtractionRes",
+    "id": "DocExtractionRes.SomeInnerModule",
     "name": "SomeInnerModule",
     "kind": "module",
     "items": [
@@ -89,7 +89,7 @@ extracting docs for src/DocExtractionRes.res
     }]
   }, 
   {
-    "id": "AnotherModule.DocExtractionRes",
+    "id": "DocExtractionRes.AnotherModule",
     "name": "AnotherModule",
     "kind": "module",
     "items": [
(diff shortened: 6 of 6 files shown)


aspeddro:
Run for entrypoint lib file `dune exec -- rescript-editor-analysis extractDocs ~/path/to/rescript-core/src/RescriptCore.res`

aspeddro:
Yes. I would like to test this in the compiler documentation.

aspeddro:
@zth, I added some error messages and an exit code.

Also renamed some fields, this simplifies writing the [decode](https://github.com/rescript-association/rescript-lang.org/blob/691ce2c1778259184363d8f37702dd5af1405775/scripts/docgen.res#L149-L188)

analysis/src/ProcessAttributes.ml
```diff
@@ -5,7 +5,7 @@ let rec findDocAttribute attributes =
   let open Parsetree in
   match attributes with
   | [] -> None
-  | ( {Asttypes.txt = "ocaml.doc" | "ns.doc" | "res.doc"},
+  | ( {Asttypes.txt = "ocaml.doc" | "ocaml.text" | "ns.doc" | "res.doc"},
```
aspeddro, line 8: The `Js.ml` module top level generate a `ocaml.text` attr. The docgen not find any docstring, see the docgen output for [Js.json](https://github.com/rescript-association/rescript-lang.org/blob/303fdf895fb2854f5616c389f15886044a50f59b/index_data/api/Js.json#L3)

The bsc typed tree for Js.ml

> Important, see the first attr [`bsconfig.config`](https://github.com/rescript-lang/rescript-compiler/blob/c799ba1fa5a7fda6c78fbd34e902f877fb1a58c0/jscomp/others/js.ml#L25)

```
[
  structure_item 
    Tstr_attribute "bs.config"
    [
      structure_item 
        Pstr_eval
        expression 
          Pexp_record
          [
            "flags" 
              expression 
                Pexp_array
                [
                  expression 
                    Pexp_constant PConst_string("-unboxed-types",None)
                  expression 
                    Pexp_constant PConst_string("-w",None)
                  expression 
                    Pexp_constant PConst_string("-49",None)
                ]
          ]
          None
    ]
  structure_item 
    Tstr_attribute "ocaml.text"
    [
      structure_item 
        Pstr_eval
        expression 
          Pexp_constant PConst_string("\n  The Js module mostly contains ReScript bindings to _standard
 JavaScript APIs_\n  like [console.log](https://developer.mozilla.org/en-US/docs/Web/API/Console/log),\
n  or the JavaScript\n  [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Glob
al_Objects/String),\n  [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_
Objects/Date), and\n  [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Globa
l_Objects/Promise)\n  classes.\n\n  It is meant as a zero-abstraction interop layer and directly expose
s JavaScript functions as they are. If you can find your API in this module, prefer this over an equiva
lent Belt helper. For example, prefer [Js.Array2](js/array-2) over [Belt.Array](belt/array)\n\n  ## Arg
ument Order\n\n  For historical reasons, some APIs in the Js namespace (e.g. [Js.String](js/string)) ar
e\n  using the data-last argument order whereas others (e.g. [Js.Date](js/date)) are using data-first.\
n\n  For more information about these argument orders and the trade-offs between them, see\n  [this blo
g post](https://www.javierchavarri.com/data-first-and-data-last-a-comparison/).\n\n  _Eventually, all m
odules in the Js namespace are going to be migrated to data-first though._\n\n  In the meantime, there 
are several options for dealing with the data-last APIs:\n\n  ```rescript\n  /* Js.String (data-last AP
I used with pipe last operator) */\n  Js.log(\"2019-11-10\" |> Js.String.split(\"-\"))\n  Js.log(\"ReSc
ript\" |> Js.String.startsWith(\"Re\"))\n\n  /* Js.String (data-last API used with pipe first operator)
 */\n  Js.log(\"2019-11-10\"->Js.String.split(\"-\", _))\n  Js.log(\"ReScript\"->Js.String.startsWith(\
"Re\", _))\n\n  /* Js.String (data-last API used without any piping) */\n  Js.log(Js.String.split(\"-\"
, \"2019-11-10\"))\n  Js.log(Js.String.startsWith(\"Re\", \"ReScript\"))\n  ```\n  ## Js.Xxx2 Modules\n
\n  Prefer `Js.Array2` over `Js.Array`, `Js.String2` over `Js.String`, etc. The latters are old modules
.\n",None)
    ]
```

analysis/src/ProcessCmt.ml
```diff
@@ -593,9 +593,11 @@ and forStructure ~name ~env strItems =
...
   in
   let attributes =
-    match strItems with
-    | {str_desc = Tstr_attribute attribute} :: _ -> [attribute]
-    | _ -> []
+    strItems
+    |> List.filter_map (fun (struc : Typedtree.structure_item) ->
+           match struc with
+           | {str_desc = Tstr_attribute attr} -> Some attr
+           | _ -> None)
```
aspeddro, line 600: Get all attrs and no longer the first.


```sh
./analysis/rescript-editor-analysis.exe extractDocs ~/Desktop/projects/rescript-compiler/lib/ocaml/js.ml | head
```
Now we have the docstring.
```
{
  "name": "Js",
  "docstrings": ["The Js module mostly contains ReScript bindings to _standard JavaScript APIs_\n  like [console.log](https://developer.mozilla.org/en-US/docs/Web/API/Console
/log),\n  or the JavaScript\n  [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String),\n  [Date](https://developer.mozilla.org/en-U
S/docs/Web/JavaScript/Reference/Global_Objects/Date), and\n  [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)\n  classes.\n
\n  It is meant as a zero-abstraction interop layer and directly exposes JavaScript functions as they are. If you can find your API in this module, prefer this over an equiva
lent Belt helper. For example, prefer [Js.Array2](js/array-2) over [Belt.Array](belt/array)\n\n  ## Argument Order\n\n  For historical reasons, some APIs in the Js namespace 
(e.g. [Js.String](js/string)) are\n  using the data-last argument order whereas others (e.g. [Js.Date](js/date)) are using data-first.\n\n  For more information about these a
rgument orders and the trade-offs between them, see\n  [this blog post](https://www.javierchavarri.com/data-first-and-data-last-a-comparison/).\n\n  _Eventually, all modules 
in the Js namespace are going to be migrated to data-first though._\n\n  In the meantime, there are several options for dealing with the data-last APIs:\n\n  ```rescript\n  /
* Js.String (data-last API used with pipe last operator) */\n  Js.log(\"2019-11-10\" |> Js.String.split(\"-\"))\n  Js.log(\"ReScript\" |> Js.String.startsWith(\"Re\"))\n\n  /
* Js.String (data-last API used with pipe first operator) */\n  Js.log(\"2019-11-10\"->Js.String.split(\"-\", _))\n  Js.log(\"ReScript\"->Js.String.startsWith(\"Re\", _))\n\n
  /* Js.String (data-last API used without any piping) */\n  Js.log(Js.String.split(\"-\", \"2019-11-10\"))\n  Js.log(Js.String.startsWith(\"Re\", \"ReScript\"))\n  ```\n  ##
 Js.Xxx2 Modules\n\n  Prefer `Js.Array2` over `Js.Array`, `Js.String2` over `Js.String`, etc. The latters are old modules."],
  "items": [
  {
    "id": "Js.t",
    "kind": "type",
    "name": "t",
    "signature": "type t<'a> = 'a constraint 'a = {..}",
```

analysis/src/ProcessAttributes.ml
zth: Nice catch!

zth:
> @zth, I added some error messages and an exit code.
> 
> Also renamed some fields, this simplifies writing the [decode](https://github.com/rescript-association/rescript-lang.org/blob/691ce2c1778259184363d8f37702dd5af1405775/scripts/docgen.res#L149-L188)

We should design the output so that we can use zero-cost modelling of the data via ReScript v11 and the new variants represenation. That should be possible with inline records. What do you think about that?

analysis/src/ProcessCmt.ml
zth: No idea why just one attribute was picked up previously. Maybe @cristianoc knows?

cristianoc: Not sure. I might be a good idea to run the tests and check that this does not change anything.

aspeddro:
I'm testing the decode in `rescript-lang` repo using version 10.1. I haven't explored the new variant representation. Does the current output make it impossible to use this feature?

analysis/src/DocExtraction.ml
```diff
@@ -225,8 +228,6 @@ let typeDetail typ ~env ~full =
          })
   | _ -> None
 
-exception Invalid_file_type
-
 let makeId modulePath ~identifier =
   identifier :: modulePath |> List.rev |> SharedTypes.ident
```
aspeddro, line 232: I think this is not correct, [Belt.json](https://github.com/rescript-association/rescript-lang.org/blob/51c4534499696b9ce7cc92f9e02878498ff73613/data/api_belt.res.json#L40)

aspeddro: Fixed e5937e09ec706db4a4295b4d52bd6072018d3919

```diff
diff --git a/data/api_belt.res.json b/data/api_belt.res.json
index 312255b..565cd1a 100644
--- a/data/api_belt.res.json
+++ b/data/api_belt.res.json
@@ -37,7 +37,7 @@
       "docstrings": ["`('key, 'id) cmparable` is a module of functions, here it only includes `cmp`.\n\n  Unlike normal functions, when created, it comes with a unique identity (guaranteed\n  by the type system).\n\n  It can be created using function [`comparableU`]() or [`comparable`]().\n\n  The idea of a unique identity when created is that it makes sure two sets would type\n  mismatch if they use different comparison function"]
     }, 
     {
-      "id": "MakeComparableU.Belt_Id.Belt",
+      "id": "Belt.Belt_Id.MakeComparableU",
       "name": "MakeComparableU",
       "kind": "module",
       "items": [
@@ -64,7 +64,7 @@
       }]
     }, 
     {
-      "id": "MakeComparable.Belt_Id.Belt",
+      "id": "Belt.Belt_Id.MakeComparable",
       "name": "MakeComparable",
       "kind": "module",
       "items": [
@@ -112,7 +112,7 @@
       "docstrings": ["`('key, 'id) hashable` is a module of functions, here it only includes `hash`, `eq`.\n\n  Unlike normal functions, when created, it comes with a unique identity (guaranteed\n  by the type system).\n\n  It can be created using function [`hashableU`]() or [`hashable`]().\n\n  The idea of a unique identity when created is that it makes sure two hash sets would type\n  mismatch if they use different comparison function"]
     }, 
     {
-      "id": "MakeHashableU.Belt_Id.Belt",
+      "id": "Belt.Belt_Id.MakeHashableU",
       "name": "MakeHashableU",
       "kind": "module",
       "items": [
@@ -146,7 +146,7 @@
       }]
     }, 
     {
-      "id": "MakeHashable.Belt_Id.Belt",
+      "id": "Belt.Belt_Id.MakeHashable",
       "name": "MakeHashable",
       "kind": "module",
       "items": [
diff --git a/data/api_js.ml.json b/data/api_js.ml.json
index df51e3a..bd87926 100644
--- a/data/api_js.ml.json
+++ b/data/api_js.ml.json
@@ -38,7 +38,7 @@
     }]
   }, 
   {
-    "id": "Internal.Js",
+    "id": "Js.Internal",
     "name": "Internal",
     "kind": "module",
     "items": [
@@ -3634,7 +3634,7 @@
       }
     }, 
     {
-      "id": "Kind.Js_json.Js",
+      "id": "Js.Js_json.Kind",
       "name": "Kind",
       "kind": "module",
       "items": [
@@ -4413,7 +4413,7 @@
       "docstrings": []
     }, 
     {
-      "id": "Type.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Type",
       "name": "Type",
       "kind": "module",
       "items": [
@@ -4426,7 +4426,7 @@
       }]
     }, 
     {
-      "id": "ArrayBuffer.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.ArrayBuffer",
       "name": "ArrayBuffer",
       "kind": "module",
       "items": [
@@ -4467,7 +4467,7 @@
       }]
     }, 
     {
-      "id": "S.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.S",
       "name": "S",
       "kind": "module",
       "items": [
@@ -4837,7 +4837,7 @@
       }]
     }, 
     {
-      "id": "Int8Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Int8Array",
       "name": "Int8Array",
       "kind": "module",
       "items": [
@@ -5256,7 +5256,7 @@
       }]
     }, 
     {
-      "id": "Uint8Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Uint8Array",
       "name": "Uint8Array",
       "kind": "module",
       "items": [
@@ -5675,7 +5675,7 @@
       }]
     }, 
     {
-      "id": "Uint8ClampedArray.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Uint8ClampedArray",
       "name": "Uint8ClampedArray",
       "kind": "module",
       "items": [
@@ -6094,7 +6094,7 @@
       }]
     }, 
     {
-      "id": "Int16Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Int16Array",
       "name": "Int16Array",
       "kind": "module",
       "items": [
@@ -6513,7 +6513,7 @@
       }]
     }, 
     {
-      "id": "Uint16Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Uint16Array",
       "name": "Uint16Array",
       "kind": "module",
       "items": [
@@ -6932,7 +6932,7 @@
       }]
     }, 
     {
-      "id": "Int32Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Int32Array",
       "name": "Int32Array",
       "kind": "module",
       "items": [
@@ -7374,7 +7374,7 @@
       "items": []
     }, 
     {
-      "id": "Uint32Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Uint32Array",
       "name": "Uint32Array",
       "kind": "module",
       "items": [
@@ -7793,7 +7793,7 @@
       }]
     }, 
     {
-      "id": "Float32Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Float32Array",
       "name": "Float32Array",
       "kind": "module",
       "items": [
@@ -8235,7 +8235,7 @@
       "items": []
     }, 
     {
-      "id": "Float64Array.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.Float64Array",
       "name": "Float64Array",
       "kind": "module",
       "items": [
@@ -8677,7 +8677,7 @@
       "items": []
     }, 
     {
-      "id": "DataView.Js_typed_array.Js",
+      "id": "Js.Js_typed_array.DataView",
       "name": "DataView",
       "kind": "module",
       "items": [
@@ -8956,7 +8956,7 @@
       "docstrings": []
     }, 
     {
-      "id": "ArrayBuffer.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.ArrayBuffer",
       "name": "ArrayBuffer",
       "kind": "module",
       "items": [
@@ -8997,7 +8997,7 @@
       }]
     }, 
     {
-      "id": "Int8Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Int8Array",
       "name": "Int8Array",
       "kind": "module",
       "items": [
@@ -9436,7 +9436,7 @@
       }]
     }, 
     {
-      "id": "Uint8Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Uint8Array",
       "name": "Uint8Array",
       "kind": "module",
       "items": [
@@ -9855,7 +9855,7 @@
       }]
     }, 
     {
-      "id": "Uint8ClampedArray.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Uint8ClampedArray",
       "name": "Uint8ClampedArray",
       "kind": "module",
       "items": [
@@ -10274,7 +10274,7 @@
       }]
     }, 
     {
-      "id": "Int16Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Int16Array",
       "name": "Int16Array",
       "kind": "module",
       "items": [
@@ -10693,7 +10693,7 @@
       }]
     }, 
     {
-      "id": "Uint16Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Uint16Array",
       "name": "Uint16Array",
       "kind": "module",
       "items": [
@@ -11112,7 +11112,7 @@
       }]
     }, 
     {
-      "id": "Int32Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Int32Array",
       "name": "Int32Array",
       "kind": "module",
       "items": [
@@ -11531,7 +11531,7 @@
       }]
     }, 
     {
-      "id": "Uint32Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Uint32Array",
       "name": "Uint32Array",
       "kind": "module",
       "items": [
@@ -11950,7 +11950,7 @@
       }]
     }, 
     {
-      "id": "Float32Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Float32Array",
       "name": "Float32Array",
       "kind": "module",
       "items": [
@@ -12369,7 +12369,7 @@
       }]
     }, 
     {
-      "id": "Float64Array.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.Float64Array",
       "name": "Float64Array",
       "kind": "module",
       "items": [
@@ -12788,7 +12788,7 @@
       }]
     }, 
     {
-      "id": "DataView.Js_typed_array2.Js",
+      "id": "Js.Js_typed_array2.DataView",
       "name": "DataView",
       "kind": "module",
       "items": [
```

analysis/src/ProcessCmt.ml
aspeddro: The tests were not changed

analysis/src/DocExtraction.ml
zth: Nice!

zth:
> I'm testing the decode in `rescript-lang` repo using version 10.1. I haven't explored the new variant representation. Does the current output make it impossible to use this feature?

Ahh, right! No, it looks fine. Here's a quick example of how bindings for most of the structures could look, zero cost: https://rescript-lang.org/try?version=v10.1.2&code=C4TwDgpgBAZglhANgEwCIHsDGUC8UDeAUFLAigHICGAthAFxQDOwATnAHYDmANMVMlmZsujBpRYtKIADxCOnAHy8SjOJ3aVgAVxb0mrebwC+hQqEhRM6dkK2Zg6Fhmx4iJKzdZ2HLKrQZyXMr8ggYiYhJSsmGKwarqmjp6gTyEJoQAAsCUnAAUAEQA1hzI+QCUZuDQyBDZcIi4UAA+UBmUjAW6ViylZVAAShDdyLn48EhoghGSMuMozgpGfS1tHfkAbuJwlOzA5VAAals7wKNzk5iiUOIz0ucLS6ZZOQXF7L2VFl1QcMAQ1Lg+Ct2gVzBBynwACpVUZ8EgkODIAIxYLwqAafz6YSpNEqNQabS6ZHY1HwgSXGJXG5RFJKOFQR4kYFrTaILTgiokI5siCw3E-JFYwz0kgY5IokVMfGJIlCoKS8kpKmRGS00mM5qtEH5ajoZBaRAQACCiG2jAhJAAsnqDcbTSC3LjEcThfyxS75fzFZTpjSYnT+b9-srbkHqAH4Y8gA

aspeddro:
I added one more change https://github.com/rescript-lang/rescript-vscode/pull/796/commits/70f19fa34f990e645a051946409decf937072d2f. `id` as valid module path, `Belt.Map.Int.checkInvariantInternal` instead `Belt.Belt_Map.Belt_MapInt.checkInvariantInternal`


```diff
diff --git a/data/api_belt.json b/data/api_belt.json
index 565cd1a..df12c7e 100644
--- a/data/api_belt.json
+++ b/data/api_belt.json
@@ -3,60 +3,60 @@
   "docstrings": ["The ReScript standard library.\n\nBelt is currently mostly covering collection types. It has no string or date functions yet, although Belt.String is in the works. In the meantime, use [Js.String](js/string) for string functions and [Js.Date](js/date) for date functions.\n\n## Motivation\n\nBelt provides:\n\n- The **highest quality** immutable data structures in JavaScript.\n- Safety by default: A Belt function will never throw exceptions, unless it is\n  indicated explicitly in the function name (suffix \"Exn\").\n- Better performance and smaller code size running on the JS platform.\n- Ready for [Tree Shaking](https://webpack.js.org/guides/tree-shaking/).\n\n## Usage\n\nTo use modules from Belt, either refer to them by their fully qualified name (`Belt.List`, `Belt.Array` etc.) or open the `Belt` module by putting\n\n```rescript\nopen Belt\n```\n\nat the top of your source files. After opening Belt this way, `Array` will refer to `Belt.Array`, `List` will refer to `Belt.List` etc. in the subsequent code.\n\nIf you want to open Belt globally for all files in your project instead, you can put\n\n```json\n\"bsc-flags\": [\"-open Belt\"]\n```\n\ninto your `bsconfig.json`.\n\n**Note**: this is the **only** `open` we encourage.\n\nExample usage:\n\n```rescript\nlet someNumbers = [1, 1, 4, 2, 3, 6, 3, 4, 2]\n\nlet greaterThan2UniqueAndSorted =\n  someNumbers\n  ->Belt.Array.keep(x => x > 2)\n  // convert to and from set to make values unique\n  ->Belt.Set.Int.fromArray\n  ->Belt.Set.Int.toArray // output is already sorted\n\nJs.log2(\"result\", greaterThan2UniqueAndSorted)\n```\n\n## Curried vs. Uncurried Callbacks\n\nFor functions taking a callback parameter, there are usually two versions\navailable:\n\n- curried (no suffix)\n- uncurried (suffixed with `U`)\n\nE.g.:\n\n```rescript\nlet forEach: (t<'a>, 'a => unit) => unit\n\nlet forEachU: (t<'a>, (. 'a) => unit) => unit\n```\n\nThe uncurried version will be faster in some cases, but for simplicity we recommend to stick with the curried version unless you need the extra performance.\n\nThe two versions can be invoked as follows:\n\n```rescript\n[\"a\", \"b\", \"c\"]->Belt.Array.forEach(x => Js.log(x))\n\n[\"a\", \"b\", \"c\"]->Belt.Array.forEachU((. x) => Js.log(x))\n```\n\n## Specialized Collections\n\nFor collections types like set or map, Belt provides both a generic module as well as specialized, more efficient implementations for string and int keys.\n\nFor example, Belt has the following set modules:\n\n- [Belt.Set](belt/set)\n- [Belt.Set.Int](belt/set-int)\n- [Belt.Set.String](belt/set-string)\n\n## Implementation Details\n\n### Array access runtime safety\n\nOne common confusion comes from the way Belt handles array access. It differs from than the default standard library's.\n\n```rescript\nlet letters = [\"a\", \"b\", \"c\"]\nlet a = letters[0] // a == \"a\"\nlet capitalA = Js.String.toUpperCase(a)\nlet k = letters[10] // Raises an exception! The 10th index doesn't exist.\n```\n\nBecause Belt avoids exceptions and returns `options` instead, this code behaves differently:\n\n```rescript\nopen Belt\nlet letters = [\"a\", \"b\", \"c\"]\nlet a = letters[0] // a == Some(\"a\")\nlet captialA = Js.String.toUpperCase(a) // Type error! This code will not compile.\nlet k = letters[10] // k == None\n```\n\nAlthough we've fixed the problem where `k` raises an exception, we now have a type error when trying to capitalize `a`. There are a few things going on here:\n\n- Reason transforms array index access to the function `Array.get`. So `letters[0]` is the same as `Array.get(letters, 0)`.\n- The compiler uses whichever `Array` module is in scope. If you `open Belt`, then it uses `Belt.Array`.\n- `Belt.Array.get` returns values wrapped in options, so `letters[0] == Some(\"a\")`.\n\nFortunately, this is easy to fix:\n\n```rescript\nopen Belt\nlet letters = [\"a\", \"b\", \"c\"]\nlet a = letters[0]\n\n// Use a switch statement:\nlet capitalA =\n  switch a {\n  | Some(a) => Some(Js.String.toUpperCase(a))\n  | None => None\n  }\n\nlet k = letters[10] // k == None\n```\n\nWith that little bit of tweaking, our code now compiles successfully and is 100% free of runtime errors!\n\n### A Special Encoding for Collection Safety\n\nWhen we create a collection library for a custom data type we need a way to provide a comparator function. Take Set for example, suppose its element type is a pair of ints, it needs a custom compare function that takes two tuples and returns their order. The Set could not just be typed as Set.t (int \\* int) , its customized compare function needs to manifest itself in the signature, otherwise, if the user creates another customized compare function, the two collection could mix which would result in runtime error.\n\nWe use a phantom type to solve the problem:\n\n```rescript\nmodule Comparable1 =\n  Belt.Id.MakeComparable(\n    {\n      type t = (int, int)\n      let cmp = ((a0, a1), (b0, b1)) =>\n        switch Pervasives.compare(a0, b0) {\n        | 0 => Pervasives.compare(a1, b1)\n        | c => c\n        }\n    }\n  )\n\nlet mySet1 = Belt.Set.make(~id=module(Comparable1))\n\nmodule Comparable2 =\n  Belt.Id.MakeComparable(\n    {\n      type t = (int, int)\n      let cmp = ((a0, a1), (b0, b1)) =>\n        switch Pervasives.compare(a0, b0) {\n        | 0 => Pervasives.compare(a1, b1)\n        | c => c\n        }\n    }\n  )\n\nlet mySet2 = Belt.Set.make(~id=module(Comparable2))\n```\n\nHere, the compiler would infer `mySet1` and `mySet2` having different type, so e.g. a `merge` operation that tries to merge these two sets will correctly fail.\n\n```rescript\nlet mySet1: t<(int, int), Comparable1.identity>\nlet mySet2: t<(int, int), Comparable2.identity>\n```\n\n`Comparable1.identity` and `Comparable2.identity` are not the same using our encoding scheme."],
   "items": [
   {
-    "id": "Belt.Belt_Id",
+    "id": "Belt.Id",
     "kind": "moduleAlias",
     "name": "Id",
     "docstrings": ["[`Belt.Id`]()\n\n  Provide utilities to create identified comparators or hashes for\n  data structures used below.\n\n  It create a unique identifier per module of\n  functions so that different data structures with slightly different\n  comparison functions won't mix"],
     "items": [
     {
-      "id": "Belt.Belt_Id.hash",
+      "id": "Belt.Id.hash",
       "kind": "type",
       "name": "hash",
       "signature": "type hash<'a, 'id>",
       "docstrings": ["`('a, 'id) hash`\n\n  Its runtime represenation is a `hash` function, but signed with a\n  type parameter, so that different hash functions type mismatch"]
     }, 
     {
-      "id": "Belt.Belt_Id.eq",
+      "id": "Belt.Id.eq",
       "kind": "type",
       "name": "eq",
       "signature": "type eq<'a, 'id>",
       "docstrings": ["`('a, 'id) eq`\n\n  Its runtime represenation is an `eq` function, but signed with a\n  type parameter, so that different hash functions type mismatch"]
     }, 
     {
-      "id": "Belt.Belt_Id.cmp",
+      "id": "Belt.Id.cmp",
       "kind": "type",
       "name": "cmp",
       "signature": "type cmp<'a, 'id>",
       "docstrings": ["`('a,'id) cmp`\n\n  Its runtime representation is a `cmp` function, but signed with a\n  type parameter, so that different hash functions type mismatch"]
     }, 
     {
-      "id": "Belt.Belt_Id.comparable",
+      "id": "Belt.Id.comparable",
       "kind": "type",
       "name": "comparable",
       "signature": "type comparable<'key, 'id> = module(Comparable with type identity = 'id)\\ntype t = 'key",
       "docstrings": ["`('key, 'id) cmparable` is a module of functions, here it only includes `cmp`.\n\n  Unlike normal functions, when created, it comes with a unique identity (guaranteed\n  by the type system).\n\n  It can be created using function [`comparableU`]() or [`comparable`]().\n\n  The idea of a unique identity when created is that it makes sure two sets would type\n  mismatch if they use different comparison function"]
     }, 
     {
-      "id": "Belt.Belt_Id.MakeComparableU",
+      "id": "Belt.Id.MakeComparableU",
       "name": "MakeComparableU",
       "kind": "module",
       "items": [
       {
-        "id": "Belt.Belt_Id.MakeComparableU.identity",
+        "id": "Belt.Id.MakeComparableU.identity",
         "kind": "type",
         "name": "identity",
         "signature": "type identity",
         "docstrings": []
       }, 
```

zth (approved):

PR #804, opened 2023-08-16
revamp insert missing cases code action

zth:
This revamps the current "Insert missing cases" code action by leveraging actual AST transforms to produce the updated switch code instead of relying on unstable and ad hoc string parsing. This is now possible thanks to the compiler printing the cases in ReScript syntax rather than OCaml syntax.

Example:
image
image
image

### DIFF
--- a/analysis/src/Commands.ml
+++ b/analysis/src/Commands.ml
@@ -392,6 +392,14 @@ let test ~path =
                                  Printf.printf "%s\nnewText:\n%s<--here\n%s%s\n"
                                    (Protocol.stringifyRange range)
                                    indent indent newText)))
+          | "c-a" ->
+            let hint = String.sub rest 3 (String.length rest - 3) in
+            print_endline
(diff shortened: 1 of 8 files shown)


analysis/src/Commands.ml
```diff
@@ -392,6 +392,14 @@ let test ~path =
                                  Printf.printf "%s\nnewText:\n%s<--here\n%s%s\n"
                                    (Protocol.stringifyRange range)
                                    indent indent newText)))
+          | "c-a" ->
```
cristianoc, line 395: What's c-a?

cristianoc: Unambiguous but getting a bit cryptic

cristianoc (approved):
Much more solid now.

analysis/src/Commands.ml
zth: It's "codemod - AddMissingCases" 😅 Might be make more sense to have the codemod variant derived from the rest of the string too...

PR #806, opened 2023-08-16
Code action: Wrap in Some

zth:
Allows for easily wrapping a pattern where an option is expected in `Some`. Useful as you change patterns from non-options to options, and need to update all parts of the pattern to be wrapped in `Some`.

Based on https://github.com/rescript-lang/rescript-vscode/pull/804 so that needs merging first.


cristianoc (approved):
Nit: this is currently based on master.
Not so easy to see the additional change.

PR #809, opened 2023-08-16
Fix signature help in uncurried mode

zth:
This wasn't working in uncurried mode.


cristianoc (approved):
It might be possible to use a utility function for function$, but that will kill pattern matching.

Issue #772, opened 2023-05-21
Create-Interface command in uncurried mode does not preserve `@react.component`


fhammerschmidt:
Consider the following example:

```rescript
@react.component
let make = (~name="") => <div />
```

In ReScript 11.0.0-alpha.6 with `"uncurried": false` when using the create-interface command, this is the resulting `.resi`:

```rescript
@react.component
let make: (~name: string=?) => Jsx.element
```

but when I set `"uncurried": true`, this is the result:

```rescript
type props<'name> = {name?: 'name}
let make: (. props<string>) => Jsx.element
```

I expected it to be the same as before.

cristianoc:
Indeed it will need to support uncurried types.

PR #810, opened 2023-08-16
Various uncurried fixes

zth:
This runs the tests in uncurried mode (+ latest beta) and fixes whatever issues surfaced.

The diff is hard to read because I can't get the printer to print uncurried without the dots. @cristianoc any idea? Getting that to work would make this much easier to follow.

Also, ignore `SignatureHelp` because fixes for that is in a separate PR.

Closes https://github.com/rescript-lang/rescript-vscode/issues/772


cristianoc:
Might need to turn on uncurried mode explicitly when the printer is called.

zth:
@cristianoc I think this was the issue: https://github.com/rescript-lang/rescript-vscode/pull/810/commits/99c0097e7d1c3518656bed4a0264be2c05b1e63b#diff-e4132b7ea215bee01eb9558296b4c196e3769d5766f09b61dce718ce5b1d8b1cR325

Compiler PR: https://github.com/rescript-lang/rescript-compiler/pull/6353

analysis/vendor/ml/printtyp.ml
```diff
@@ -49,10 +49,11 @@ let ident ppf id = pp_print_string ppf (ident_name id)
 (* Print a path *)
 
 let ident_pervasives = Ident.create_persistent "Pervasives"
+let ident_pervasives_u = Ident.create_persistent "PervasivesU"
```
zth, line 52: This should be backported to the compiler.

analysis/vendor/res_syntax/res_outcome_printer.ml
```diff
@@ -322,9 +322,7 @@ let rec printOutTypeDoc (outType : Outcometree.out_type) =
       ]
 
 and printOutArrowType ~uncurried typ =
-  let uncurried =
-    if !Config.uncurried <> Legacy then not uncurried else uncurried
-  in
+  let uncurried = Res_uncurried.getDotted ~uncurried !Config.uncurried in
```
zth, line 325: PR:ed to the compiler here: https://github.com/rescript-lang/rescript-compiler/pull/6353

analysis/tests/src/expected/Completion.res.txt
```diff
@@ -1597,14 +1599,8 @@ Resolved opens 3 Completion.res Completion.res pervasives
...
-Found type for function (~b: int) => int
-[{
-    "label": "b",
-    "kind": 4,
-    "tags": [],
-    "detail": "int",
-    "documentation": null
-  }]
+Found type for function int
+[]
```
zth, line 1603: This is right in uncurried mode I believe. The code is:
```rescript
type callback = (~a: int) => int

let withCallback: (~b: int) => callback = (~b) => { (); (~a) => a + b }
```

And the test is:
```rescript
// withCallback(~a)(~
//                                 ^com
```

In curried mode I guess the returned callback is "merged" into the main function definition, so `~a` works even if it's in the _callback_ and not the main function. but in uncurried it's required to explicitly call `withCallback` with the right parameter `~b` before you get the callback back and can invoke that.

analysis/tests/src/expected/SignatureHelp.res.txt
```diff
@@ -10,12 +10,11 @@ ContextPath Value[someFunc]
 Path someFunc
```
zth, line 10: Ignore this file, we checked it in the last PR. I just removed the dedicated uncurried test file here.

analysis/tests/src/expected/Definition.res.txt
```diff
@@ -8,10 +8,10 @@ Hover src/Definition.res 14:14
...
 
 Hover src/Definition.res 18:14
-{"contents": {"kind": "markdown", "value": "```rescript\n(Belt.List.t<'a>, 'a => 'b) => Belt.List.t<'b>\n```\n\n---\n\n```\n \n```\n```rescript\ntype Belt.List.t<'a> = list<'a>\n```\nGo to: [Type definition](command:rescript-vscode.go_to_location?%5B%22belt_List.mli%22%2C34%2C0%5D)\n\n\n\n  Returns a new list with `f` applied to each element of `someList`.\n\n  ```res example\n  list{1, 2}->Belt.List.map(x => x + 1) // list{3, 4}\n  ```\n"}}
+{"contents": {"kind": "markdown", "value": "```rescript\n(Belt.List.t<'a>, 'a => 'b) => Belt.List.t<'b>\n```\n\n---\n\n```\n \n```\n```rescript\ntype Belt.List.t<'a> = list<'a>\n```\nGo to: [Type definition](command:rescript-vscode.go_to_location?%5B%22belt_List.resi%22%2C35%2C0%5D)\n\n\n\n  Returns a new list with `f` applied to each element of `someList`.\n\n  ```res example\n  list{1, 2}->Belt.List.map(x => x + 1) // list{3, 4}\n  ```\n"}}
```
zth, line 11: Changes from ReScriptifying Belt. Looks fine.

analysis/tests/src/expected/CreateInterface.res.txt
```diff
@@ -1,6 +1,6 @@
 Create Interface src/CreateInterface.res
 type r = {name: string, age: int}
-let add: (~x: int, ~y: int) => int
+let add: (. ~x: int, ~y: int) => int
```
zth, line 3: Haven't figured out why this is printed like this yet.

zth:
@cristianoc this is ready for review. Should we run all tests in uncurried mode primarily moving forward, or should I revert back to curried mode? What do you think? 

I personally think it might be too soon to run in uncurried mode and optimize for that as the main case given so few have migrated to it yet, but that we should switch over and test the output then and then. 

There's the alternative of trying to set up 2 test suites, one for curried and one for uncurried. But I don't know.

analysis/tests/src/expected/Completion.res.txt
cristianoc: Indeed the two are not equivalent in uncurried mode.

analysis/vendor/res_syntax/res_outcome_printer.ml
cristianoc: Does that fix the issue?

cristianoc (approved):

analysis/vendor/res_syntax/res_outcome_printer.ml
zth: It fixes the printing issues in this repo, yes. There were dots in lots of places there shouldn't be in uncurried mode before this.

cristianoc:
> @cristianoc this is ready for review. Should we run all tests in uncurried mode primarily moving forward, or should I revert back to curried mode? What do you think?
> 
> I personally think it might be too soon to run in uncurried mode and optimize for that as the main case given so few have migrated to it yet, but that we should switch over and test the output then and then.
> 
> There's the alternative of trying to set up 2 test suites, one for curried and one for uncurried. But I don't know.

Two suites sounds heavy, unless you feel like having that.
Perhaps yes one can test back and forth, and switch to the prevalent one at the time. So no switch yet.

zth:
@cristianoc the test failures are interesting. They're different errors depending on whether it's Linux or Windows, but what they have in common is that "Go To Definition" seems to point at _different files_ depending on platform. This started happening with the later versions of v11 where more things have been ReScriptified. 

Any ideas?

zth:
Sorry I pushed another commit that muddied up what I commented on. If you want to have a look at the problem I described, you can check these two failed runs:
- https://github.com/rescript-lang/rescript-vscode/actions/runs/5893455531/job/15985036846
- https://github.com/rescript-lang/rescript-vscode/actions/runs/5893455531/job/15985037017

PR #812, opened 2023-08-18
Code actions for exhaustive switches

zth:
This introduces a code action for inserting an exhaustive switch. Works for variants, polyvariants, options, and bools. It can be triggered in the following ways:

### Putting the cursor on an identifier

https://github.com/rescript-lang/rescript-vscode/assets/1457626/3f0a2ec7-3e1c-44fc-ae25-8a1912a59e0c

### Selecting an expression

https://github.com/rescript-lang/rescript-vscode/assets/1457626/a9fcee90-5a5b-4bd7-b2fb-8e8b9315cfff

### DIFF
--- a/analysis/src/CompletionFrontEnd.ml
+++ b/analysis/src/CompletionFrontEnd.ml
@@ -222,7 +222,8 @@ let completePipeChain (exp : Parsetree.expression) =
     exprToContextPath exp |> Option.map (fun ctxPath -> (ctxPath, pexp_loc))
   | _ -> None
 
-let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
+let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor
+    ?findThisExprLoc text =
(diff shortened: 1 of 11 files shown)


analysis/src/CompletionFrontEnd.ml
```diff
@@ -777,6 +778,12 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor ~text =
           (Pos.toString posCursor) (Pos.toString posNoWhite)
           (Loc.toString expr.pexp_loc)
     in
+    (match findThisExprLoc with
+    | Some loc when expr.pexp_loc = loc -> (
+      match exprToContextPath expr with
+      | None -> ()
+      | Some ctxPath -> setResult (Cpath ctxPath))
+    | _ -> ());
```
zth, line 786: This is pretty messy to include like this in the existing autocomplete. But, it's a pretty valuable feature, and this part of the code base needs a big refactor soon anyway, so I think it's OK given what it enables.

cristianoc (approved):

PR #813, opened 2023-08-19
Make completion engine understand await

zth:
This makes the completion engine understand async/await.


cristianoc (approved):
This could be considered an instance of apply but it does not add much complexity so looks good!

PR #814, opened 2023-08-20
Instantiate variant types as needed

zth:
This fixes a bug where for example completion from `result` wouldn't work, because the variant's type variables wasn't instantiated (and result is linked to `Belt.Result.t`, so I guess the type variables got lost in the linkage).


cristianoc (approved):
nice fix !

PR #816, opened 2023-09-11
Make function template async when fn returns a promise

zth:
Small quality of life improvement - when a function returns a promise, make the suggested function template async.


cristianoc (approved):
Nice touch

PR #824, opened 2023-10-04
Account for "public" setting in bsconfig

zth:
This accounts for the "public" setting in `bsconfig`. Seems like a fairly hidden property, but we should still support it imo given that trying to use something not listed in public (when public is set) is an error, even though items appear in autocomplete.

Example from the forum: https://forum.rescript-lang.org/t/how-to-only-expose-one-module-from-a-lib/4744/2

### DIFF
--- a/analysis/src/FindFiles.ml
+++ b/analysis/src/FindFiles.ml
@@ -105,6 +105,18 @@ let getNamespace config =
     either fromString fromName |> Option.map nameSpaceToName
   else None
 
+module StringSet = Set.Make (String)
+
+let getPublic config =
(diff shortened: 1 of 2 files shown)


cristianoc (approved):
Is this documented anywhere?
Do you want to keep this feature?

zth:
Not documented I think, at least not anymore. I don't understand why though, because it seems like a great feature to get around that you can't really hide top level modules.

I'd definitively want to keep the feature, at least OTOH. But I think posting on the forum about it might be good. Maybe someone remembers if this was deprecated at some point and if so why, given that it isn't documented from what I can see.

PR #828, opened 2023-10-10
Support rescript.json

fhammerschmidt:
Gave it a try, not sure if it is the right approach though.


zth:
Looks good to me! And you tested that everything worked with both rescript.json and bsconfig.json, including the code analyzer etc?

cc @cristianoc

cristianoc:
> Looks good to me! And you tested that everything worked with both rescript.json and bsconfig.json, including the code analyzer etc?
> 
> cc @cristianoc

Same comment. Looks good if everything is tested.

fhammerschmidt:
Now I tested it and of course found some missing checks. Should work now for both monorepos and normal ones.

cknitt:
Tested against a big monorepo project of ours, worked fine, even with rescript.json and bsconfig.json mixed across packages.

zth:
Excellent! Merging and we can fix in follow ups.

PR #830, opened 2023-10-12
Touch ups to LSP and Tools packages

zth:
* Changes to a more uniform naming of distributed binaries
* Adds an example of how we could bind to the structure the docgen emits zero cost using ReScript v11

These things are fine to do because neither of the packages have had their initial versions published yet. We'll hold off publishing the initial tools version at least until we've figured out these last details.

cc @aspeddro 

Also, @aspeddro did you have a basic generator from this format to markdown in some default format? Maybe something you've done when experimenting on the website?


aspeddro:
There was a basic generator for markdown, but it wasn't correct.

I am exploring generating the documents through `getStaticProp`. I had to do some transformations in JSON

Here is the latest deploy: https://rescript-lang-fn6212fyj-rescript-association.vercel.app/docs/manual/next/api/js

PR: https://github.com/rescript-association/rescript-lang.org/pull/703

It is on the `next` route. The paths are the same as the module path. So the module `Belt.SortArray.Int` -> [`belt/sortarray/int`](https://rescript-lang-fn6212fyj-rescript-association.vercel.app/docs/manual/next/api/belt/sortarray/int)

PR #832, opened 2023-10-13
Upgrade `ocamlformat@0.26.1` 

aspeddro:
This allows using latest `ocaml-lsp-server`


zth:
Is upgrading the LSP blocked because of an older ocaml-format version? Seriously?

cknitt:
Yes, see also https://github.com/rescript-lang/rescript-compiler/pull/6360.

We cannot upgrade the version in the compiler repo though because we also install and run it in CI for format checks, and it is not in the OPAM repo used for Windows...

I think we should avoid using a different version in the rescript-vscode repo than in the compiler repo.

zth:
Right, got it. Just perplexed that upgrading a _formatter_ (implicitly meaning potentially dealing with a bunch of reformatting conflicts for all outstanding branches) is allowed to block upgrading the LSP itself.

aspeddro:
Ready for merge

PR #834, opened 2023-10-17
Add (failing) test for hover on aliased record

fhammerschmidt:
At work we are using aliases a lot (see screenshot). I think we should not lose hover info when they are aliased. So I added a failing test as a first step towards that.

image


zth:
Superseded by https://github.com/rescript-lang/rescript-vscode/pull/881

PR #837, opened 2023-10-21
Show Docstrings for completion items

### LINKED ISSUE #469: Show docs for file modules on type hint
Related to module level docstring support (see [syntax issue](https://github.com/rescript-lang/syntax/issues/579)).

When e.g. type hinting `Belt_Array`, I get `file module` as docs:
image

Ideally, it would show the module level docs instead.

aspeddro:
Neovim:
image

VSCode:
image

Related #469
Close #469

### DIFF
--- a/analysis/src/CompletionBackEnd.ml
+++ b/analysis/src/CompletionBackEnd.ml
@@ -55,7 +55,7 @@ let resolveOpens ~env opens ~package =
     (* loop(previous) *)
     [] opens
 
-let completionForExporteds iterExported getDeclared ~prefix ~exact ~env ~package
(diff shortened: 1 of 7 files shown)


zth:
I like this in general! But, I worry about perf. Does this mean that the cmt is processed for every file module that can be found? If so, it's important to test the cases where we complete and get tons of file modules back. Like when completing for "nothing" at the top level for example.

Could you bench this somehow? Preferably in a larger project. We could for example consider doing the lookups for only the first X file modules, and do them lazily as needed.

zth:
@aspeddro could we check the perf here somehow?

aspeddro:
I'll check this week

aspeddro:
I changed it to draft bc the implementation is wrong

PR #841, opened 2023-11-06
Add pre-release to `@rescript/language-server`

aspeddro:
`npm version preminor --preid next` increase minor version.

Example:
```diff
- "version": "1.22.0",
+ "version": "1.23.0-next.0"
```

install using `next` tag: `npm i -g @rescript/language-server@next`


zth:
I guess this doesn't hurt to have, although I wonder how many will actually use it...? (thanks for the effort btw @aspeddro, as always 😄 )

aspeddro:
I think it will be used more by Vim/Neovim users.

I used the pre-release Github release pages a lot.

aspeddro:
@zth can we get on with this? I would like to test the pre-release as you are adding a lot of cool features.

I would also like to mention this in the post for vim/neovim that I will make.

zth:
Let's give it a shot!

PR #842, opened 2023-11-11
Fix issue with ordering of module opened via -open

zth:
This fixes a very annoying issue where you wouldn't get autocomplete for "second order opens". I haven't added any tests because it'd be an incredibly noisy diff, but here's roughly what was wrong and what it caused:
1. Say you have an open in `bsc-flags`, like `-open RescriptCore`. Everything works as expected
2. The `RescriptCore` open exposes a bunch of modules like `Array`, `List`, `Result` etc. All of these are available in autocomplete. So far, so good.
3. Now, let's say you were to add an explicit open in a file for a module exposed via `-open RescriptCore`, like `open Array`.
4. The content of `Array` _won't_ be exposed to autocomplete in this case.

As can be seen by the minimal diff, this was a case of ordering. The list is reversed (correctly), but the order of `rawOpens` (coming from the local file) and `packageOpens` (coming from the package, bsconfig/rescript.json) was off, so the package opens actually ended up _after_ the local opens when resolving autocomplete. This in turn lead to the contents of the opened modules not being exposed.

Phew.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## master
 
+#### :bug: Bug Fix
+
+- Fix issue `open` on submodules exposed via `-open` in bsconfig.json/rescript.json, that would cause the content of those `open` modules to not actually appear in autocomplete. https://github.com/rescript-lang/rescript-vscode/pull/842
+
 ## 1.22.0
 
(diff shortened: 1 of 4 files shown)


cristianoc (approved):
Seems legit.

PR #844, opened 2023-11-11
Dont complete for illegal file module names

zth:
We handled namespace file modules before, but other "exotic" file modules could still appear (example: `SomeTest.test.res`) which can't be used anyway. This filters anything with illegal characters in the name, for file modules specifically.

### DIFF
--- a/analysis/src/Utils.ml
+++ b/analysis/src/Utils.ml
@@ -222,7 +222,7 @@ let cutAfterDash s =
   | n -> ( try String.sub s 0 n with Invalid_argument _ -> s)
   | exception Not_found -> s
 
-let hasUnallowedChars s =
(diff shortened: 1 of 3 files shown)


analysis/src/Utils.ml
```diff
@@ -216,3 +216,10 @@ let rec lastElements list =
 let lowercaseFirstChar s =
   if String.length s = 0 then s
   else String.mapi (fun i c -> if i = 0 then Char.lowercase_ascii c else c) s
+
+let hasUnallowedChars s =
```
cristianoc, line 220: Unclear this is about file names by just looking at the code.

cristianoc (approved):

PR #847, opened 2023-11-13
feat: cli option to prevent 'start build' message

dkirchhof:
Hey,

when I updated the language server manually for my Neovim config, I always set the askToStartBuild option in the code was to false.
Since I don't have to extract the server code from the vscode extension anymore (many thanks for the npm package) I wouldn't like to change the code manually, so an option for the cli would be usefull.

Without overcomplicating the code, I just added a second optional param to the cli.

### DIFF
--- a/server/src/cli.ts
+++ b/server/src/cli.ts
@@ -12,15 +12,16 @@ Options:
 
 --stdio               Use stdio
 --node-ipc            Use node-ipc
+--dont-ask            Don't ask to start build (use it as second parameter)
 -v, --version         Print version
 -h, --help            Print help`;
 
 (() => {
   switch (args[0]) {
     case '--stdio':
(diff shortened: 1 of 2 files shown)


aspeddro:
How is your setup configuration?

aspeddro:
I need to document this, but you can pass the configuration at startup

```lua
require'lspconfig'.rescriptls.setup {
    cmd = { 'rescript-language-server', '--stdio' },
    init_options = {
    extensionConfiguration = {
      askToStartBuild = false
    },
  }
}
```

dkirchhof:
Oh great. Didn't know that. So I will close my PR. Would be nice, if it will be documented somewhere in the lspconfig docs?!

aspeddro:
yes

PR #850, opened 2023-11-14
More robust handling of namespaces in pipe completion

zth:
This is a more robust implementation of https://github.com/rescript-lang/rescript-vscode/pull/843 that should cover all cases instead of just a few. Here, instead of checking for namespaces when filtering the completion items, we flatten any namespace items to their accessible paths before filtering. This allows the regular filter mechanism to just do its work like it's used to, and the filtering will work for namespaced items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## master
 
+#### :bug: Bug Fix
+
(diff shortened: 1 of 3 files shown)


zth:
@cristianoc it would be nice to do the same in the printer, we can still get namespaced type names there:
type RescriptBun Globals Request t

cristianoc (approved):

PR #851, opened 2023-11-14
Showcase stack overflow bug with polyvariants

zth:
More info on the issue here, including a variant of that code snippet that compiles (thanks to explicit type annotations I guess) here: https://github.com/rescript-association/rescript-lang.org/issues/716#issuecomment-1729913023


cristianoc:
> Fatal error: exception Stack overflow
+Raised at Ident.find_same in file "vendor/ext/ident.ml", line 156, characters 6-21
+Called from Printtyp.ident_name in file "vendor/ml/printtyp.ml", line 42, characters 6-38

PR #856, opened 2023-12-04
fix: add _ to charaters allowed in modules name

WhyThat:
Hello,
since #844 we don't have our modules listed in completion 

This PR fix that


zth:
Good catch!

PR #865, opened 2023-12-12
If interface exists, ask if it should be overwritten

fhammerschmidt:
Small QoL feature

image


zth (approved):
Great stuff! Changelog?

Issue #863, opened 2023-12-12
Uncurried needs to default to true for ReScript 11


cknitt:
In ReScript 11, "uncurried" is true by default (i.e., if not set explicitly in rescript.json).
However, it seems the VSCode plugin still assumes that the default is false.

To test: Create a new project using `npm create rescript-app@next` with the "Basic" template and add "let f = () => ()" to Demo.res.

Observe CodeLens / tool tips showing the function type as `(. unit) => unit`.

<img width="280" alt="Bildschirmfoto 2023-12-12 um 14 11 49" src="https://github.com/rescript-lang/rescript-vscode/assets/591384/2ec44966-89d6-4a03-8bc2-ca980c2c56d4">

(tested with the latest prerelease version 1.29.5)

PR #867, opened 2023-12-15
Look up and propagate ReScript version

zth:
So we can default uncurried by default properly.

Ideally this would be done entirely in the analysis bin, but given that I don't want to mess with the logic of finding and executing the ReScript binary, I think it's fine that it's propagated through an env variable for now, given that the logic lives in the LS server itself (so vim and other LS users won't have to have their own logic).

Closes https://github.com/rescript-lang/rescript-vscode/issues/863

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 #### :bug: Bug Fix
 
+- Proper default for `"uncurried"` in V11 projects. https://github.com/rescript-lang/rescript-vscode/pull/867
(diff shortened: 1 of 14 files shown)


cristianoc (approved):
Curious why an env variable is used. Seems similar to other project config info which I think goes into some config state.

zth:
> Curious why an env variable is used. Seems similar to other project config info which I think goes into some config state.

I added some justification in the original description. But I'm not terribly happy with it.

cknitt (approved):

PR #868, opened 2023-12-17
Tools: enable doc extraction from compiler

aspeddro:
This PR has one workaround to extract docs from the compiler repo and a fix

1. Introduce an env var `FROM_COMPILER` to analysis find the correct path to `cmt` files. In standalone packages, the `cmt` files are located in `lib/bs/`, but in the compiler repo they are in `lib/ocaml`
2. Add `stdio: "inherit"`. Node was truncating output when I run a separate process in `rescript-tools doc path/to/rescript-compiler/lib/ocaml/js.ml` https://github.com/nodejs/node/issues/2360

### DIFF
--- a/analysis/src/BuildSystem.ml
+++ b/analysis/src/BuildSystem.ml
@@ -6,15 +6,18 @@ let namespacedName namespace name =
 let ( /+ ) = Filename.concat
 
 let getBsPlatformDir rootPath =
-  let result =
-    ModuleResolution.resolveNodeModulePath ~startPath:rootPath "rescript"
-  in
-  match result with
-  | Some path -> Some path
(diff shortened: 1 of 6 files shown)


zth (commented):
What about doing it as a regular variable to the cli command instead of an env variable?

aspeddro:
This is just to generate the compiler docs. It is an internal and exclusive use for extracting compiler docstrings in the `rescript-lang` repo via the `rescript-tools` package

I don't think it would make sense to other users

zth:
@aspeddro ok, sounds good! Is this ready to go then?

aspeddro:
yup. CHANGELOG updated

PR #869, opened 2023-12-17
Track scope properly when inferring values

zth:
This fixes an infinite loop bug that was cause by not tracking the scope properly as we infer values from unsaved content. All inference worked on the same, final scope, which led to infinite loops when adding a new binding to scope with a name that was already in scope.

Now, each value in scope tracks its own scope properly, and this fixes the bug.

### DIFF
--- a/analysis/src/Scope.ml
+++ b/analysis/src/Scope.ml
@@ -1,12 +1,8 @@
-type item =
-  | Constructor of string * Location.t
-  | Field of string * Location.t
-  | Module of string * Location.t
-  | Open of string list
-  | Type of string * Location.t
-  | Value of string * Location.t * SharedTypes.Completable.contextPath option
+type item = SharedTypes.ScopeTypes.item
 
 type t = item list
+
+open SharedTypes.ScopeTypes
 
 let itemToString item =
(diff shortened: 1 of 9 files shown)


analysis/src/Scope.ml
```diff
@@ -1,13 +1,9 @@
...
-  | Field of string * Location.t
-  | Module of string * Location.t
-  | Open of string list
-  | Type of string * Location.t
-  | Value of string * Location.t * SharedTypes.Completable.contextPath option
+type item = SharedTypes.ScopeTypes.item
 
 type t = item list
 
+open SharedTypes.ScopeTypes
```
cristianoc, line 5: where is this used?

cristianoc: Ah in the pattern matching below.

cristianoc (approved):
sweet

no chances of infinite loops left?

zth:
> sweet
> 
> no chances of infinite loops left?

Oh no idea lol. Probably. But at least not from this issue.

analysis/src/Scope.ml
zth: Yeah exactly, I opened it just to have it function as it did before.

PR #874, opened 2023-12-18
handle completion for polyvars without leading #

zth:
Fixes so that you don't need to explicitly type a leading `#` to get completion when the underlying value is a polyvar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## master
 
+#### :bug: Bug Fix
+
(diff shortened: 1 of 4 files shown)


cristianoc (approved):

DZakh:
It works 😍

It's not critical, but maybe it would be nice not to add curly brackets around variants.

image

zth:
@DZakh the curly braces are intentional

DZakh:
Ok

PR #877, opened 2023-12-19
Extend quoting functionality to include polyvariant keywords

WhyThat:
Hey,
Following up #870, we have now extended the functionality to also wrap polyvariant keywords with quotes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 #### :bug: Bug Fix
 
 - Fix so that you don't need a leading `#` to complete for polyvariant constructors. https://github.com/rescript-lang/rescript-vscode/pull/874
+- Print keyword polyvariant constructors with quotes when doing completions. https://github.com/rescript-lang/rescript-vscode/pull/877
 
 ## 1.30.0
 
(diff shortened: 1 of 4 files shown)


WhyThat:
Maybe we could use https://github.com/rescript-lang/rescript-vscode/blob/0d8fbc3b49e9f3dedd119968781b29ee925ee103/analysis/src/Utils.ml#L225

zth:
Great stuff! Can you add a changelog?

WhyThat:
@zth Done :)
Just thinking about it, we could use `fileNameHasUnallowedChars` to detect if we have to wrap exotic polyvariant, wyt ?

zth:
> @zth Done :) Just thinking about it, we could use `fileNameHasUnallowedChars` to detect if we have to wrap exotic polyvariant, wyt ?

I don't think that's equivalent to what we do, right? Doesn't that check that the string is capitalized etc?

WhyThat:
> I don't think that's equivalent to what we do, right? Doesn't that check that the string is capitalized etc?

Seems like it match the same things after testing these cases : 
```ocaml
let printMaybeExoticIdent txt =
  (* fileNameHasUnallowedChars "not good" (* => true *); *)
  (* fileNameHasUnallowedChars "not-good" (* => true *); *)
  (* fileNameHasUnallowedChars "good"(* => false *); *)
  (* fileNameHasUnallowedChars "Good"(* => false *); *)
  (* fileNameHasUnallowedChars "123"(* => false *); *)
  match (Res_token.isKeywordTxt txt, fileNameHasUnallowedChars txt) with
  | true, _ | _, true -> "\"" ^ txt ^ "\""
  | _ -> txt
```
test are unchanged with this, but nevermind

PR #878, opened 2023-12-29
fix(tools): get docstrings from module file

aspeddro:
```bash
FROM_COMPILER=true ./analysis/rescript-editor-analysis.exe extractDocs ~/Desktop/projects/rescript-compiler/jscomp/others/belt.res | jq '{items}[] | map(select(.id == "Belt.Array")) | .[0] | .docstrings'
```

Before:

```
[
  "[`Belt.Array`]()\n\n  **mutable array**: Utilities functions"
]
```

After:

```
[
  "Utililites for `Array` functions.\n\n### Note about index syntax\n\nCode like `arr[0]` does *not* compile to JavaScript `arr[0]`. Reason transforms\nthe `[]` index synta
x into a function: `Array.get(arr, 0)`. By default, this\nuses the default standard library's `Array.get` function, which may raise an\nexception if the index isn't found. 
If you `open Belt`, it will use the\n`Belt.Array.get` function which returns options instead of raising exceptions. \n[See this for more information](../belt.mdx#array-acce
ss-runtime-safety)."
]
```

### DIFF
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -294,7 +294,7 @@ let extractDocs ~path ~debug =
                    let id =
                      (modulePath |> List.rev |> List.hd) ^ "." ^ item.name
                    in
-                   let items, docstring =
+                   let items, internalDocstrings =
                      match
                        ProcessCmt.fileForModule ~package:full.package
                          aliasToModule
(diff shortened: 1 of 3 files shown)


zth:
Should we get both rather than dropping one of them?

aspeddro:
of course, it's an array

zth (approved):

PR #880, opened 2024-01-06
Do not crash entire extension when analysis fails

zth:
Instead report the error as a message, but only every 15 minutes.

We want people to report when things crash, but having the entire extension crash because of some potentially minor thing (code actions failing to resolve for example) is a bad experience. This makes sure that the extension does not crash when the analysis fails, but rather reports a message with the error (and not too often).

This is a good balance imo between getting people to report and preserving the UX of the extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ## master
 
+- Better error recovery when analysis fails. https://github.com/rescript-lang/rescript-vscode/pull/880
+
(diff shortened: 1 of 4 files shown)


cristianoc (approved):
Is there any additional info shown when it actually crashes?
Or is this equivalent in terms of info reported.

zth:
Yeah this now shows some information around what crashed. Still quite bare bones, but more than before.

PR #882, opened 2024-01-06
Include fields when completing for identifiers in a braced expr that is likely a module path

zth:
This is a QoL improvement intended to simplify the following scenario where you'd constructing a record without a type definition, but the expression is ambiguous and we don't know whether you're constructing a record yet or just typing some other expression:

```rescript
let someRecord = {
  SomeModuleWithTheRecordDefinition.<com>
}
```

Previously you'd only get values from `SomeModuleWithTheRecordDefinition`, but with this change you'll also get any fields visible in that module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,10 @@
 ## master
 
 - Better error recovery when analysis fails. https://github.com/rescript-lang/rescript-vscode/pull/880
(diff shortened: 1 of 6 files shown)


cristianoc (approved):
Is there an analogue in pattern matching?

zth:
> Is there an analogue in pattern matching?

I don't think so, I think pattern matching already does the right thing.

PR #883, opened 2024-01-06
Complete lowercase JSX labels from the domProps type

zth:
Change completion of DOM props of lowercase elements to try to look up `ReactDOM.domProps` first, before falling back to the current method of completing from a pre-defined list of DOM props (that was there because of JSXv3).

This is an adaption to JSXv4, and also a preparation for our upcoming generic JSX transform where we'll want to be able to swap out `domProps` to something custom for the generic transform.


cristianoc (approved):
nice touch

PR #884, opened 2024-01-06
Complete creator/maker functions for type t

zth:
Sometime we reach a `type t` from a module when completing expressions, and cannot go any further because `type t` might be abstract, or something we cannot complete. This is pretty common when using `type t` to hide implementation.

In those scenarios, it's a convention to provide `make` or other creator functions if the type can be constructed. This PR adds completions for all maker-style functions whenever constructing an expression where the expected type is `type t` and we can't go any further in the resolution.

This is going to make APIs where the type is abstract and hidden behind `type t` _a lot_ more discoverable.

Example from a classic situation where the value we're supposed to pass is an abstract `type t` that hides the implementation of `t` and has its own `make` function (quite common scenario):
image

Another example where `type t` is abstract and hidden behind different maker functions (fairly common in bindings when binding to something very polymorphic that cannot be represented in the type system):
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Include fields when completing a braced expr that's an ID, where it the path likely starts with a module. https://github.com/rescript-lang/rescript-vscode/pull/882
 - Complete domProps for lowercase JSX components from `ReactDOM.domProps` if possible. https://github.com/rescript-lang/rescript-vscode/pull/883
 - Do not emit `_` when completing in patterns. https://github.com/rescript-lang/rescript-vscode/pull/885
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Better error recovery when analysis fails. https://github.com/rescript-lang/rescript-vscode/pull/880
 - Expand type aliases in hovers. https://github.com/rescript-lang/rescript-vscode/pull/881
 - Include fields when completing a braced expr that's an ID, where it the path likely starts with a module. https://github.com/rescript-lang/rescript-vscode/pull/882
+- Complete for maker-style functions when encountering a `type t` in relevant scenarios. https://github.com/rescript-lang/rescript-vscode/pull/884
```
cristianoc, line 18: describe maker-style function: I guess, function that returns `t`.

cristianoc (approved):

Issue #887, opened 2024-01-08
doc extracted signature of variant with inline-record incomplete?


woeps:
Given the following rescript [code](https://github.com/rescript-lang/rescript-vscode/blob/d0c842a3042a3904b1f9b45907008d46bc839d20/analysis/tests/src/DocExtractionRes.res#L28):

```rescript
type status =
    | /** If this is started or not */ Started(t) | /** Stopped? */ Stopped | /** Now idle.*/ Idle
```
is supposed to be extracted as the following [(partial) json](https://github.com/rescript-lang/rescript-vscode/blob/d0c842a3042a3904b1f9b45907008d46bc839d20/analysis/tests/src/expected/DocExtractionRes.res.txt#L49C4-L75C8):

```json
 {
      "id": "DocExtractionRes.SomeInnerModule.status",
      "kind": "type",
      "name": "status",
      "signature": "type status = Started(t) | Stopped | Idle",
      "docstrings": [],
      "detail": 
      {
        "kind": "variant",
        "items": [
        {
          "name": "Started",
          "docstrings": ["If this is started or not"],
          "signature": "Started(t)"        <-- HERE
        }, 
        {
          "name": "Stopped",
          "docstrings": ["Stopped?"],
          "signature": "Stopped"
        }, 
        {
          "name": "Idle",
          "docstrings": ["Now idle."],
          "signature": "Idle"
        }]
      }
    }, 
```

Especially mind the `Started` constructor having it's payload present in the json details' `signature` field!

Whereas it seems that the payload is not exported into it's json `signature` field, if it is an inline-record:
e.g. the following [code](https://github.com/rescript-lang/rescript-vscode/blob/d0c842a3042a3904b1f9b45907008d46bc839d20/analysis/tests/src/DocExtractionRes.res#L49):

```rescript
type someVariantWithInlineRecords = | /** This has inline records...*/ SomeStuff({offline: bool})
```

is currently tested to result in the following [(partial) json](https://github.com/rescript-lang/rescript-vscode/blob/d0c842a3042a3904b1f9b45907008d46bc839d20/analysis/tests/src/expected/DocExtractionRes.res.txt#L117-L133):

```json
  {
      "id": "DocExtractionRes.AnotherModule.someVariantWithInlineRecords",
      "kind": "type",
      "name": "someVariantWithInlineRecords",
      "signature": "type someVariantWithInlineRecords =\\n  | SomeStuff({offline: bool})",
[...]

PR #889, opened 2024-01-09
Support inline record fields in doc extraction

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/887

I've also removed all of the JSON decoding code, because with untagged variants we can just map to the result zero cost, which is a lot easier to maintain.

cc @aspeddro

### DIFF
--- a/analysis/tests/src/expected/DocExtractionRes.res.txt
+++ b/analysis/tests/src/expected/DocExtractionRes.res.txt
@@ -52,6 +52,7 @@ extracting docs for src/DocExtractionRes.res
     "id": "DocExtractionRes.SomeInnerModule",
     "name": "SomeInnerModule",
     "kind": "module",
+    "docstrings": ["Another module level docstring here."],
     "items": [
     {
       "id": "DocExtractionRes.SomeInnerModule.status",
@@ -99,6 +100,7 @@ extracting docs for src/DocExtractionRes.res
     "id": "DocExtractionRes.AnotherModule",
     "name": "AnotherModule",
     "kind": "module",
+    "docstrings": ["Mighty fine module here too!"],
     "items": [
     {
       "id": "DocExtractionRes.LinkedModule",
@@ -161,6 +163,7 @@ extracting docs for src/DocExtractionRes.res
     "id": "DocExtractionRes.ModuleWithThingsThatShouldNotBeExported",
     "name": "ModuleWithThingsThatShouldNotBeExported",
     "kind": "module",
+    "docstrings": [],
     "items": [
     {
       "id": "DocExtractionRes.ModuleWithThingsThatShouldNotBeExported.t",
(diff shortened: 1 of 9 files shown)


analysis/tests/src/expected/DocExtractionRes.res.txt
```diff
@@ -127,7 +127,18 @@ extracting docs for src/DocExtractionRes.res
...
           "name": "SomeStuff",
           "docstrings": ["This has inline records..."],
-          "signature": "SomeStuff"
+          "signature": "SomeStuff({offline: bool, online?: bool})",
+          "inlineRecordFields": [{
+            "name": "offline",
+            "optional": false,
+            "docstrings": [],
+            "signature": "bool"
+          }, {
```
aspeddro, line 136: I think is better for the field to be named `items`. A variant payload can be a tuple and we use `items` elsewhere

```jsonc
{
  "name": "SomeStuff",
  "items": [{...}],
}
```

woeps:
Removing the JSON decoding code let a bug surface: [`type item` in Tools_Docgen.resi](https://github.com/rescript-lang/rescript-vscode/blob/c369db02457051c95cbddc698a72db27efd78f05/tools/src/Tools_Docgen.resi#L43) defines
```rescript
Module({
  id: 
  docstrings: array<string>,
  deprecated?: string,
  name: string,
  items: array<item>,
})
```
having docstrings not being optional, but the generated json (by analysis binary) does not contain the docstrings field for module items.

I believe, I found the issue in [DocExtraction.ml#L156-166](https://github.com/rescript-lang/rescript-vscode/blob/c369db02457051c95cbddc698a72db27efd78f05/analysis/src/DocExtraction.ml#L153-L165):
```ml
  | Module m ->
    stringifyObject ~startOnNewline:true ~indentation
      [
        ("id", Some (wrapInQuotes m.id));
        ("name", Some (wrapInQuotes m.name));
        ("kind", Some (wrapInQuotes "module"));
        ("docstrings", Some (wrapInQuotes (stringifyDocstrings m.docstring)));  (* THIS LINE IS MISSING! *)
        ( "items",
          Some
            (m.items
            |> List.map
                 (stringifyDocItem ~originalEnv ~indentation:(indentation + 1))
            |> array) );
      ]
```
which doesn't stringified the docstring field yet.

zth:
@woeps perfect, I'll fix it!

zth:
@woeps can you test the latest commits?

tools/src/Tools_Docgen.resi
```diff
@@ -5,11 +5,16 @@ type field = {
...
+
+@tag("kind")
+type constructorPayload = | @as("inlineRecord") InlineRecord({fields: array<field>})
+
 type constructor = {
   name: string,
   docstrings: array<string>,
   signature: string,
   deprecated?: string,
+  constructorPayload?: constructorPayload,
```
woeps, line 17: @zth This field name (`constructorPayload`) is not matching the field name given in [`stringifyDetail in DocExtraction.ml`](https://github.com/rescript-lang/rescript-vscode/blob/719f8a29f0691d7e7ad3029ae4d64acb6459b4ed/analysis/src/DocExtraction.ml#L121) (`items`).
If I'm not mistaken, both names need to be identical: either `items` or `constructorPayload`.

aspeddro: @zth changed `items` from array to object then better rename field to `payload` or `params`?

aspeddro:
Gabriel, can we remove JSON decode soon (maybe 0.5 release)? At the moment, I use the decode function on the website.

PR #891, opened 2024-01-10
fix extracted docs of types include escaped linebreaks in signature

### LINKED ISSUE #888: extracted docs of types include escaped linebreaks in signature
This rescript code:
```rescript
type variant = FirstButVeryLongAndMeaninglessVariantConstructor

type t = {thisIsSomeReallyLongFieldWhichIsJustHereForDemonstration: bool}

```

gets extracted as:
```json
{
  "name": "MinimalExample",
  "docstrings": [],
  "items": [
  {
    "id": "MinimalExample.variant",
    "kind": "type",
    "name": "variant",
    "signature": "type variant =\\n  | FirstButVeryLongAndMeaninglessVariantConstructor", <-- HERE
    "docstrings": [],
    "detail": 
    {
      "kind": "variant",
      "items": [
      {
        "name": "FirstButVeryLongAndMeaninglessVariantConstructor",
        "docstrings": [],
        "signature": "FirstButVeryLongAndMeaninglessVariantConstructor"
      }]
    }
  }, 
  {
    "id": "MinimalExample.t",
    "kind": "type",
    "name": "t",
    "signature": "type t = {\\n  thisIsSomeReallyLongFieldWhichIsJustHereForDemonstration: bool,\\n}", <-- HERE
    "docstrings": [],
    "detail": 
    {
      "kind": "record",
      "items": [{
        "name": "thisIsSomeReallyLongFieldWhichIsJustHereForDemonstration",
        "optional": false,
        "docstrings": [],
        "signature": "bool"
      }]
    }
  }]
}
```
Note the escaped linebreaks (`\\n`) in the signature.  
This seems to happen, if a signature reaches some minimal length.

I would expect the line breaks to not be escaped.

*edit*:  
I believe, this is due to [calling `Json.escape`](https://github.com/rescript-lang/rescript-vscode/blob/d48ee3373794750d4a3874eeae2aa2d4387bc1a6/analysis/vendor/json/Json.ml#L140).

Not sure, what's best. But I guess, it's ether not escaping those special characters or "unescaping" them, when actually parsing the json in [Tools_DocGen.res](https://github.com/rescript-lang/rescript-vscode/blob/23bcade78a95fd9c354379000fe6204f6b35c5cf/tools/src/Tools_Docgen.res#L150)?

woeps:
Closes #888

`Json.escape` was called twice:

- once explicitly
- twice in `Protocol.wrapInQuotes`

I also updated the test, to not contain e.g. `\\n`

Since this is my first PR in this repo, I hope this is fine?

### DIFF
--- a/analysis/src/DocExtraction.ml
+++ b/analysis/src/DocExtraction.ml
@@ -112,13 +112,13 @@ let rec stringifyDocItem ?(indentation = 0) ~originalEnv (item : docItem) =
       [
         ("id", Some (wrapInQuotes id));
         ("kind", Some (wrapInQuotes "value"));
-        ("name", Some (name |> Json.escape |> wrapInQuotes));
+        ("name", Some (name |> wrapInQuotes));
         ( "deprecated",
           match deprecated with
           | Some d -> Some (wrapInQuotes d)
           | None -> None );
         ( "signature",
-          Some (signature |> String.trim |> Json.escape |> wrapInQuotes) );
+          Some (signature |> String.trim |> wrapInQuotes) );
         ("docstrings", Some (stringifyDocstrings docstring));
       ]
   | Type {id; docstring; signature; name; deprecated; detail} ->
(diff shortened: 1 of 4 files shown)


aspeddro:
You can update `tools/CHANGELOG.md`?

zth:
Awesome! I'm a bit worried about exotic identifiers if we don't escape. @woeps could you add a test case with a value like `let \"SomeConstant\" = 12` and just make sure that the output doesn't break?

woeps:
> Awesome! I'm a bit worried about exotic identifiers if we don't escape. @woeps could you add a test case with a value like `let \"SomeConstant\" = 12` and just make sure that the output doesn't break?

We still escape, because `Protocol.wrapInQuotes` calls `Json.escape`.
Before this PR, there were 4 places with `someVar |> Json.escape |> Protocol.wrapInQuotes`. This caused the strings being escapes twice.

Of cause, I'll add an additional test case and a note to the changelog.

zth:
@woeps thanks for the explanation. This is ready to merge after that's settled. Could you test the other in progress PR too? I can release a new tools version after both of these are merged then.

zth:
@woeps looks good, although there's a weird escape of `:` but maybe that doesn't matter. Ready to go?

woeps:
@zth not sure which escape you mean?

Now (after updating changelog) it's good to go.

PR #892, opened 2024-01-10
Bugfixes for type resolutions

zth:
This clears up a number of issues for resolving types across/within files, moving through type manifests, and what not. It'll make pattern completion and expression completion a lot more stable.

This introduces a fair bit of duplication and so on. There's _a lot_ of technical debt now with duplicated functions and what not. I plan on taking a larger grip at fixing that and improving the code in general after this has landed. Just need to find a way to do it incrementally.

However, I want to land this before cleaning things up, because this improves the completion situation a lot as is, and cleaning up won't give any new benefits to users more than us having more maintainable code. So, what I'm saying is bear with me with the extra things added here and it'll be cleaned up soon.

### DIFF
--- a/analysis/src/TypeUtils.ml
+++ b/analysis/src/TypeUtils.ml
@@ -266,70 +266,7 @@ let rec extractFunctionType2 ?typeArgContext ~env ~package typ =
   in
   loop ?typeArgContext ~env [] typ
 
-(** Pulls out a type we can complete from a type expr. *)
-let rec extractType ~env ~package (t : Types.type_expr) =
-  match t.desc with
-  | Tlink t1 | Tsubst t1 | Tpoly (t1, []) -> extractType ~env ~package t1
-  | Tconstr (Path.Pident {name = "option"}, [payloadTypeExpr], _) ->
-    Some (Toption (env, TypeExpr payloadTypeExpr))
-  | Tconstr (Path.Pident {name = "promise"}, [payloadTypeExpr], _) ->
-    Some (Tpromise (env, payloadTypeExpr))
-  | Tconstr (Path.Pident {name = "array"}, [payloadTypeExpr], _) ->
-    Some (Tarray (env, TypeExpr payloadTypeExpr))
-  | Tconstr (Path.Pident {name = "result"}, [okType; errorType], _) ->
-    Some (Tresult {env; okType; errorType})
-  | Tconstr (Path.Pident {name = "bool"}, [], _) -> Some (Tbool env)
-  | Tconstr (Path.Pident {name = "string"}, [], _) -> Some (Tstring env)
-  | Tconstr (Path.Pident {name = "exn"}, [], _) -> Some (Texn env)
-  | Tconstr (Pident {name = "function$"}, [t; _], _) -> (
-    (* Uncurried functions. *)
-    match extractFunctionType t ~env ~package with
-    | args, tRet when args <> [] ->
-      Some (Tfunction {env; args; typ = t; uncurried = true; returnType = tRet})
-    | _args, _tRet -> None)
-  | Tconstr (path, typeArgs, _) -> (
-    match References.digConstructor ~env ~package path with
-    | Some (env, {item = {decl = {type_manifest = Some t1; type_params}}}) ->
-      t1
-      |> instantiateType ~typeParams:type_params ~typeArgs
-      |> extractType ~env ~package
-    | Some (env, {name; item = {decl; kind = Type.Variant constructors}}) ->
-      Some
-        (Tvariant
-           {env; constructors; variantName = name.txt; variantDecl = decl})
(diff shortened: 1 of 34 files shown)


analysis/src/TypeUtils.ml
```diff
@@ -1,5 +1,69 @@
 open SharedTypes
 
+let debugLogTypeArgContext {env; typeArgs; typeParams} =
+  Printf.sprintf "Type arg context. env: %s, typeArgs: %s, typeParams: %s\n"
+    (Debug.debugPrintEnv env)
+    (typeArgs |> List.map Shared.typeToString |> String.concat ", ")
+    (typeParams |> List.map Shared.typeToString |> String.concat ", ")
+
+let rec pathFromTypeExpr (t : Types.type_expr) =
```
zth, line 9: Just moved

analysis/src/TypeUtils.ml
```diff
@@ -1,5 +1,69 @@
...
+  match t.desc with
+  | Tconstr (Pident {name = "function$"}, [t; _], _) -> pathFromTypeExpr t
+  | Tconstr (path, _typeArgs, _)
+  | Tlink {desc = Tconstr (path, _typeArgs, _)}
+  | Tsubst {desc = Tconstr (path, _typeArgs, _)}
+  | Tpoly ({desc = Tconstr (path, _typeArgs, _)}, []) ->
+    Some path
+  | _ -> None
+
+let printRecordFromFields ?name (fields : field list) =
```
zth, line 19: Just moved

analysis/src/TypeUtils.ml
```diff
@@ -1,5 +1,69 @@
...
+  (match name with
+  | None -> ""
+  | Some name -> "type " ^ name ^ " = ")
+  ^ "{"
+  ^ (fields
+    |> List.map (fun f -> f.fname.txt ^ ": " ^ Shared.typeToString f.typ)
+    |> String.concat ", ")
+  ^ "}"
+
+let rec extractedTypeToString ?(inner = false) = function
```
zth, line 29: Just moved

analysis/src/TypeUtils.ml
```diff
@@ -48,6 +112,59 @@ let instantiateType ~typeParams ~typeArgs (t : Types.type_expr) =
     in
     loop t
 
+let instantiateType2 ?(typeArgContext : typeArgContext option)
```
zth, line 115: Slightly different to the existing `instantiateType` - this instantiates via comparing `Tvar` name only, while the existing one structurally checks whether the `Tvar` types are the same. In this new solution they won't be the exact same object since we're moving across type manifests and what not.

analysis/src/TypeUtils.ml
```diff
@@ -109,6 +226,45 @@ let rec extractFunctionType ~env ~package typ =
...
+    in
+    (match typeArgContext with
+    | None -> ()
+    | Some typeArgContext ->
+      if Debug.verbose () then
+        Printf.printf "[#type_arg_ctx]--> setting new type arg ctx: %s"
+          (debugLogTypeArgContext typeArgContext));
+    typeArgContext
+
+let rec extractFunctionType2 ?typeArgContext ~env ~package typ =
```
zth, line 246: Same as existing one but doesn't instantiate. Could potentially remove this one all together in the larger refactor as it's only special because it can do lookup of types.

analysis/src/TypeUtils.ml
```diff
@@ -178,6 +328,171 @@ let rec extractType ~env ~package (t : Types.type_expr) =
     | _args, _tRet -> None)
   | _ -> None
 
+let rec extractType2 ?(printOpeningDebug = true)
```
zth, line 331: This can't fully replace `extractType` yet because that fn is used internally here some. Will need to be fully replaced in separate refactor.

analysis/src/TypeUtils.ml
```diff
@@ -48,6 +112,59 @@ let instantiateType ~typeParams ~typeArgs (t : Types.type_expr) =
...
+  | Some {typeArgs; typeParams} ->
+    let rec applySub tp ta name =
+      match (tp, ta) with
+      | {Types.desc = Tvar (Some varName)} :: tRest1, t2 :: tRest2 ->
+        if varName = name then t2 else applySub tRest1 tRest2 name
+      | _ :: tRest1, _ :: tRest2 -> applySub tRest1 tRest2 name
+      | [], _ | _, [] -> t
+    in
+
+    let rec loop (t : Types.type_expr) =
```
cristianoc, line 128: any concerns with termination here, or are the cases clear?

cristianoc (approved):
Impressive set of changes.

analysis/src/TypeUtils.ml
zth: I think the cases are clear. Or at least this has been around forever and worked so far AFAICT (the code is just copied).

PR #894, opened 2024-01-11
Fix ambigious case with JSX brace wraps and record body

zth:
Fix ambigious case where JSX brace wraps in JSX props are interpreted as a record body, but we dont know whether it actually is a record body or just wraps for the prop value itselfix ambigious case where JSX brace wraps in JSX props are interpreted as a record body, but we dont know whether it actually is a record body or just wraps for the prop value itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 #### :bug: Bug Fix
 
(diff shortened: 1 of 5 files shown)


zth:
Side note: Life is so much easier with proper debug statements. Will continue adding these as I fix things.

cristianoc (approved):
I think the description in the change log should say that in case of ambiguity, both completions are provided.

PR #900, opened 2024-01-27
tools: add `source` field (Loc)

aspeddro:
Emit location of type/value/module

```jsonc
{filepath: "src/A.resi", line: 2, col: 2}
```

In general, you can run the command from anywhere. The `filepath` property is relative to the package root.


zth:
Missing ReScript bindings here, right? Should add them to the ReScript files.

aspeddro:
ReScript files updated!

tools/src/tools.ml
```diff
@@ -257,7 +280,40 @@ let typeDetail typ ~env ~full =
...
+    match dirname dirPath = realPath with
+    | true -> basename dirPath :: acc |> String.concat "/"
+    | false -> (
+      match dirPath with
+      | "/" -> failwith "Failed to find relative path of package"
+      | "." -> path
+      | _ -> loop (dirname dirPath) (basename dirPath :: acc))
+  in
+
+  loop path []
```
zth, line 304: No built in fn that does this I take it...?

aspeddro: thank you. Found in `Analysis.Files.realpath`

5396f79b

zth: Great! Now we just need to figure out the tests for Windows. Paths being different is obviously expected.

zth (approved):

PR #903, opened 2024-01-28
Enhance regexp completion

zth:
This enhances completion where the type is a regexp a bit, by:
- Preferring Core's `RegExp` module when Core is opened
- Adding a `%re("")` item automatically to the completion list in any place where a regexp value is expected

Small things, but should add some QoL.


cristianoc (approved):
Sweet

PR #905, opened 2024-01-29
up tests to ReScript v11

zth:
We now default to v11 in tests.


analysis/src/Packages.ml
```diff
@@ -13,7 +13,7 @@ let makePathsForModule ~projectFilesAndPaths ~dependenciesFilesAndPaths =
 
 let getReScriptVersion () =
   (* TODO: Include patch stuff when needed *)
-  let defaultVersion = (10, 1) in
+  let defaultVersion = (11, 0) in
```
zth, line 16: This is intentional. We can easily roll back to defaulting to 10.1 if this causes troubles (it shouldn't).

cristianoc (approved):

cristianoc:
Notice CI is red.

PR #913, opened 2024-02-07
Completion for import attributes in @module

zth:
Adds completion for the new import attributes in `@module`.


cristianoc:
looks like CI is red
leftover of the previous CI issues which were not fully solved?

zth:
@cristianoc look again

cristianoc (approved):

PR #917, opened 2024-02-07
Latest version of parser

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/915

### DIFF
--- a/analysis/tests/src/expected/CompletionJsx.res.txt
+++ b/analysis/tests/src/expected/CompletionJsx.res.txt
@@ -429,17 +429,17 @@ posCursor:[30:12] posNoWhite:[30:11] Found expr:[11:4->32:10]
 posCursor:[30:12] posNoWhite:[30:11] Found expr:[12:4->32:10]
 posCursor:[30:12] posNoWhite:[30:11] Found expr:[15:5->32:10]
 JSX <div:[15:5->15:8] > _children:15:8
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[15:8->32:4]
-Pexp_construct :::[16:7->32:4] [16:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[16:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[17:7->32:4]
-Pexp_construct :::[17:7->32:4] [17:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[17:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->32:4]
-Pexp_construct :::[30:10->32:4] [30:10->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->30:12]
-JSX <di:[30:10->30:12] > _children:None
+posCursor:[30:12] posNoWhite:[30:11] Found expr:[15:8->33:2]
+Pexp_construct :::[16:7->33:2] [16:7->33:2]
+posCursor:[30:12] posNoWhite:[30:11] Found expr:[16:7->33:2]
+posCursor:[30:12] posNoWhite:[30:11] Found expr:[17:7->33:2]
+Pexp_construct :::[17:7->33:2] [17:7->33:2]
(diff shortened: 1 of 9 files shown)


analysis/tests/src/expected/CompletionJsx.res.txt
```diff
@@ -429,17 +429,17 @@ posCursor:[30:12] posNoWhite:[30:11] Found expr:[11:4->32:10]
...
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[16:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[17:7->32:4]
-Pexp_construct :::[17:7->32:4] [17:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[17:7->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->32:4]
-Pexp_construct :::[30:10->32:4] [30:10->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->32:4]
-posCursor:[30:12] posNoWhite:[30:11] Found expr:[30:10->30:12]
-JSX <di:[30:10->30:12] > _children:None
+posCursor:[30:12] posNoWhite:[30:11] Found expr:[15:8->33:2]
```
cristianoc, line 432: which location is correct?

cristianoc (approved):

analysis/tests/src/expected/CompletionJsx.res.txt
zth: Not sure actually. Think it's a problem?

cristianoc: Depends on whether they're wrong or fixed now.

cristianoc: Not sure why the output is different, but it's this test:
```res
    <div>
      {React.string(someProp)}
      <div> {React.null} </div>
      // {someString->st}
      //                ^com
      // {"Some string"->st}
      //                   ^com
      // {"Some string"->Js.String2.trim->st}
      //                                    ^com
      // {someInt->}
      //           ^com
      // {12->}
      //      ^com
      // {someArr->a}
      //            ^com
      // <di
      //    ^com
    </div>
```

I think when completing `di`, it finds the following `/div` and now considers it as a prop of `di`, while before it did not.

The one thing that comes to mind is that the treatment of nested jsx could have changed with the jsx mode pops etc in this:
  https://github.com/rescript-lang/rescript-compiler/pull/6609
  
@tsnobip any thoughts?
Not sure there's anything to do, just double checking if anything comes to mind. Otherwise we can just go ahead.

tsnobip: it is totally possible that my PR accidentally created some regressions, do you have an example in mind of syntax that should not be recognized but is now? I could add tests. I'm not really happy about working with the syntax parser right now, especially around mode, I think it can introduce bugs that are hard to spot.

cristianoc: The only thing that comes to mind: if inside a jsx one finds a nested jsx, when the nested one is closed, is the mode again in jsx mode for the outer one?

tsnobip: Hmm, we should likely add tests around this. I'm not sure what's the best way to do that.
Basically, before, jsx mode tended to "leak" outside of jsx, which created unwanted behavior like accepting hyphens in identifiers outside of jsx. I tightened the jsx mode to avoid that, but I might have restricted it too much. I might have popped jsx mode one time too many.

cristianoc: It could be that the difference here is because jsx mode is now correctly restored while before it was not.
In which case that's great.

zth: I've tested this for a bit and I can't find any obvious issues with it. Not tests fail either from what I see. I'd like to merge this, because 2 other PRs depend on this, but I'd also not want to drop the ball here on if this change actually is problematic. How can we investigate it further? Add tests in the syntax for nested JSX mode?

cristianoc: This can be merged.
One can add a couple of tests of nested jsx separately in the compiler.

zth: @tsnobip would you be up for adding tests for various forms of nested JSX in the compiler repo?

PR #918, opened 2024-02-07
Enhance module attribute completion some more

zth:
- Relax filter for what local files that come up in `from` and regular string completion in `@module`
- Make `from` completion trigger for expr hole so we get a nice experience when completing `{from: <com>}`.


cristianoc:
Any idea what's up with CI? It seems to report on `.js` and `.ml` files, which is strange.

PR #919, opened 2024-02-07
Generic JSX transform completion

zth:
This adapts the JSX lowercase element completion (both for the attribute names and values) to look up the proper type if using a generic JSX transform, instead of relying on hard coded paths to React and ReactDOM.

I set up a new test project for this because that felt the least invasive compared to changing the main test project's JSX config.

Closes https://github.com/rescript-lang/rescript-vscode/issues/914


.gitignore
```diff
@@ -7,6 +7,10 @@ analysis/tests/lib
 analysis/tests/.bsb.lock
 analysis/tests/.merlin
 
+analysis/tests-generic-jsx-transform/lib
+analysis/tests-generic-jsx-transform/.bsb.lock
+analysis/tests-generic-jsx-transform/.merlin
```
cristianoc, line 12: I think `.merlin` files are not generated anymore.

cristianoc (approved):

PR #920, opened 2024-02-13
Highlight tagged template fn as fn

zth:
This highlights tagged template functions used as a function.

Before:
image

After:
image


DZakh:
@zth @fhammerschmidt I wonder how correct it is to highlight it as a function, when we usually treat functions as regular variables in other places

image

zth:
Other variables we don't know whether they're functions or not, because we'd need to do potentially expensive lookups. This one we always know is a function because only a function can be used as a tagged template literal function.

DZakh:
I'm just thinking that it might be better to be consistent in not highlighting functions

zth:
I don't think that's a problem, this is also a pretty isolated context from other functions so it's fine.

DZakh:
> Other variables we don't know whether they're functions or not, because we'd need to do potentially expensive lookups.

By the way, do you know how TS does it?

zth:
> > Other variables we don't know whether they're functions or not, because we'd need to do potentially expensive lookups.
> 
> By the way, do you know how TS does it?

They use semantic highlighting, and look up the actual type definition for each identifier to see if it's a function.

We could do the same, but we'd need to benchmark it because it _might_ be taxing on performance. So far we haven't used any of the type information for highlighting (IIRC). But we could probably try it out as an experiment, should be easy enough. Would be good to have benchmarks for it.

DZakh:
I see. But I actually like functions of the same color as variables 😁

Issue #902, opened 2024-01-28
handling of module types (in doc extraction)


woeps:
Currently, it seems to me there is no distinction made between a `module` and a `module type` in doc extraction (I believe, this is true for the LSP as well?). The extracted docs (json) gives no hint, that it's a module type and not an actual implementation.

Furthermore, any implementing module explicitly annotated with the module type is being ignored in doc extraction:

The following rescript code:

```rescript
// ModuleType.res

/***
extracted docs don't give a hint, that Example is not an implementation, but a module type

module M is missing in extracted docs
*/

module type Example = {
  /***
  this is an example module type 
  */

  /**
  main type of this module 
  */
  type t

  /**
  function from t to t
  */
  let f: t => t
}

module M: Example = {
  /***
  implementation of Example module type
  */

  /**
  main type 
  */
  type t = int

  /**
  identity function
  */
  let f = (x: int) => x
}

```

produces this json:
```json
{
  "name": "ModuleType",
  "docstrings": ["extracted docs don't give a hint, that Example is not an implementation, but a module type\n\nmodule M is missing in extracted docs"],
  "items": [
  {
    "id": "ModuleType.Example",
    "name": "Example",
    "kind": "module",
    "docstrings": [],
    "items": [
    {
      "id": "ModuleType.Example.t",
      "kind": "type",
      "name": "t",
      "signature": "type t",
      "docstrings": ["main type of this module"]
    }, 
    {
      "id": "ModuleType.Example.f",
      "kind": "value",
      "name": "f",
      "signature": "let f: (. t) => t",
      "docstrings": ["function from t to t"]
    }]
  }]
}
```

I'm honestly not sure how I'd expect it to be represented in the extracted json. (`kind` currently being one of `Value`, `Type`, `Module`)

1. [x] I believe the minimal requirement is to distinguish an actual module implementation from a module type, since I won't be able to actually call the lather one. [addressed in #925]
2. [ ] Any (Sub-) Module (not hidden by a `*.resi` file) should be present in the docs.
3. [ ] There is no autocompletion help by the LSP to help with annotating the module type with a module type being a *submodule in the same file*. (The LSP helps annotating the module type, when the module type from another file is used.)
4. [x] The LSP doesn't distinguish between `module` and `module type` as well:  
  Typing `let x = ModuleType.Example.`, LSP offers completion for `ModuleType.Example.f`. - Which is just defined in the module type. Therefore this will yield a compilation error. [addressed in #925]

[...]

PR #925, opened 2024-02-25
distinguish regular modules from module types

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/902

@woeps please look through this and say what you think. A few things I can think of right away:
- We don't indicate when a module is implementing a module type. Not sure if this would be relevant for docgen.
- We can have id clashes now I guess, because I think a regular module and a module type with the same name can co-exist in the same module scope. So we need to account for that in the id generation.

### DIFF
--- a/tools/tests/src/expected/DocExtractionRes.res.json
+++ b/tools/tests/src/expected/DocExtractionRes.res.json
@@ -60,10 +60,10 @@
     }
   }, 
   {
-    "id": "DocExtractionRes.SomeConstant\\",
+    "id": "DocExtractionRes.SomeConstant",
     "kind": "value",
-    "name": "SomeConstant\\",
-    "signature": "let SomeConstant\\: int",
+    "name": "SomeConstant",
+    "signature": "let SomeConstant: int",
     "docstrings": ["exotic identifier"],
     "source": {
       "filepath": "src/DocExtractionRes.res",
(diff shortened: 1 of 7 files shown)


tools/tests/src/expected/DocExtractionRes.res.json
```diff
@@ -270,5 +270,41 @@
...
+      "kind": "value",
+      "name": "f",
+      "signature": "let f: t => t",
+      "docstrings": ["function from t to t"],
+      "source": {
+        "filepath": "src/DocExtractionRes.res",
+        "line": 109,
+        "col": 3
+      }
+    }]
```
woeps, line 308: I'd also expect `"id": "DocExtractionRes.M"` to be present. - Which is missing here?

tools/tests/src/expected/DocExtractionRes.res.json
```diff
@@ -270,5 +270,41 @@
         "col": 3
       }
     }]
+  }, 
+  {
+    "id": "DocExtractionRes.Example",
+    "name": "Example",
+    "kind": "moduleType",
+    "docstrings": [],
```
woeps, line 278: docstrings shouldn't be empty, since the module type contains:
```rescript
module type Example = {
  /***
  this is an example module type 
  */
...
}
```

woeps (commented):
Wow, you were fast!

* What about adding the signature field to the Module specific records (like in non-module related `tools.docItem` constructors? Have it show e.g. `"signature": "module M: Example"` or for modules not using a previously defined module type `"signature": "module AnotherModule: { type t; let x: string; ..."`.
* Do you refer to @aspeddro s work? He introduced the `source` record `{filepath: string, line: int, col: int}`, which should be different in your described case.
* Ideally, I'd like to have a `source` record also available for the refered module type. The record type `tools.docsForModule` includes a source field, which describes, where the module was specified. But if the module uses a module type, there is currently no source for that module type. Having such info would be very handy for linking up the different files in a site generator.
* Am I right to assume, these changes do not affect the language server's auto-complete suggestions? - I only tested tools locally.

tools/tests/src/expected/DocExtractionRes.res.json
zth: I believe `/***` is for the top level only. If you want a comment on an in-file module you need to use `/**` just above the module definition, like regular docstrings.

woeps: You confused me for a minute there.. ;) 
I don't think so.
See my simple example:
https://github.com/woeps/rescript-tools-doc-md/blob/54b93c3da527074be767227c057e5b10dbdf7b85/test/ExampleModule.res#L6

https://github.com/woeps/rescript-tools-doc-md/blob/54b93c3da527074be767227c057e5b10dbdf7b85/test/md/Example.md?plain=1#L166

woeps:
Any further updates/thoughts on this one?

zth:
@woeps not really, but I'd be happy to try and fix it quickly if I understood exactly what needs to be done. Will try and re-read the instructions.

woeps:
@zth I previously created several comments in this PR regarding overall expectations of docgen.
The PR does, what's stated in the title.

I'd suggest to merge this as is and I'll create follow up issues for next steps.

zth:
@woeps ok, merging this and we'll continue in separate issues for the other things.

PR #926, opened 2024-02-27
grammar: Remove punctuation.terminator since it doesn't exist in ReScript

DZakh:


zth:
Hmm, `;` does exist in ReScript, but the formatter removes it. Not sure if we should remove it.

PR #936, opened 2024-02-29
Fix issue where completion inside of switch expr would not work

zth:
Fixes an issue where completion inside of switch exprs wouldn't work because we iterate the cases instead when we shouldn't. 

Adds a bunch of debug logging as well.


zth:
CI fails for unrelated reasons.

cristianoc (approved):

PR #947, opened 2024-03-09
Extend signature help to also work on variant constructor payloads

zth:
Made it possible to turn off if wanted. This can be quite useful - figure out what arguments a constructor takes (and where you're at with the arguments) via signature help. 

Examples:
image
image
image


cristianoc (approved):
Something makes CI unhappy

PR #950, opened 2024-03-10
Signature help fixes

zth:
* Ensures that locations are taken into account, so that the deepest signature help item with the cursor is the one that triggers.
* Refactors the function call lookup so that we try using typed information first before resorting to using the completion enginge for unsaved content. This will yield better results with incremental type checking because type params etc will be instantiated.


cristianoc (approved):
Any changes when the incremental compilation is off?

zth:
> Any changes when the incremental compilation is off?
Should not be, no.

PR #951, opened 2024-03-10
Refactor: completion creation fns to just one

zth:
A long overdue refactor. No need for 2 functions to exist. 

Also assumes all clients support snippets, which they do.


cristianoc (approved):
So clean

PR #952, opened 2024-03-10
Cleanup completion details docs

zth:
Closes https://github.com/rescript-lang/rescript-vscode/issues/469 and https://github.com/rescript-lang/rescript-vscode/pull/837

This does 2 things:
- Does a major clean up of the details and docs generated for completion items.
- Implements the LSP `completionResolve` request so we can resolve (some) completion item docs async. Then via that implement file module docs resolution. 

This is a big diff test wise, because I've changed quite a few things wrt details and docs. It might not all be perfect yet, but it's going to be much much better than what we have today already, given what mess the details and docs for completion items are today.

Example of file module level docs:
image

### DIFF
--- a/analysis/tests/src/expected/CompletionAttributes.res.txt
+++ b/analysis/tests/src/expected/CompletionAttributes.res.txt
@@ -74,20 +74,20 @@ Resolved opens 1 pervasives
     "label": "version",
     "kind": 5,
     "tags": [],
-    "detail": "version?: int\n\ntype jsxConfig = {version: int, module_: string, mode: string}",
-    "documentation": null
+    "detail": "int",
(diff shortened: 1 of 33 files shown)


analysis/tests/src/expected/CompletionAttributes.res.txt
```diff
@@ -74,20 +74,20 @@ Resolved opens 1 pervasives
     "label": "version",
     "kind": 5,
     "tags": [],
-    "detail": "version?: int\n\ntype jsxConfig = {version: int, module_: string, mode: string}",
-    "documentation": null
+    "detail": "int",
```
cristianoc, line 77: change in detail intentional?

cristianoc (approved):

analysis/tests/src/expected/CompletionAttributes.res.txt
zth: Yes, intentional.

PR #955, opened 2024-03-13
Enable signature help for option and result

zth:
Doesn't need a changelog.


cristianoc (approved):
Why only on these 3 variant cases?
Do they have something special in common?

zth:
> Why only on these 3 variant cases? Do they have something special in common?

Since option and result are now builtins, they're not covered by the generic variant stuff (which is what was used before). So they need to be handled separately like this.

PR #974, opened 2024-04-28
Add experimental command for extracting (string) contents from extension points

zth:
Internal for now. Using it to experiment with the codegen approach I'm working on for ReScript.


zth:
@cristianoc merging this. It's internal/experimental, so it'll be easy to remove later on if we don't want it.

PR #976, opened 2024-05-01
Build static linux binaries, including arm64

cknitt:
* Build static binaries for Linux which fixes glibc compatibility issues. Uses same Docker container as in the rescript-compiler repo.
* Build Linux ARM binaries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,16 +12,59 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test:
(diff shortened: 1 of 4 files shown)


zth (approved):
Absolutely fantastic! Do we need to update any logic around selecting what binary to use...?

cknitt:
The existing logic should be fine:
https://github.com/rescript-lang/rescript-vscode/blob/3b5824b3957e95acbb79592ce8ea4e16d04dee5e/client/src/utils.ts#L6

But I haven't really tested it on Linux ARM64 yet. Could you try it?

PR #977, opened 2024-05-01
Build with OCaml 5.1.1

cknitt:
Build with OCaml 5.1.1, like on the master branch of the rescript-compiler repo.

So that we can also use the same matching version of the docker container for the static Linux build which I'll add in a separate PR.


zth:
Awesome! 😁

PR #982, opened 2024-05-24
Use canonicalized URIs/paths for jump to definition

fhammerschmidt:
This fixes the issue when you have an internal package in a monorepo and you click on a module from it in another package, you end up at `node_modules/@yourNamespace/yourPackageName`. Now it uses the real path and not the symlink anymore, so end up at the correct location.


zth (approved):
Mind adding a changelog?

zth (approved):

zth:
@fhammerschmidt just a rebase and we're ready to go.

fhammerschmidt:
rebased and added a changelog now!

Issue #962, opened 2024-03-19
Incorrect autocomplete for jsx props after a variant


DZakh:
<img width="1041" alt="image" src="https://github.com/rescript-lang/rescript-vscode/assets/49292466/ba35ceb0-6b37-43f8-bbaf-c01fb4a1d55a">

I expect to get an autocomplete for the rest `Info` props here.

zth:
Could you investigate if the type of the prop matters here, or if it's the same regardless of type?

zth:
Also if you could PR a failing test then that'd be great.

PR #984, opened 2024-05-25
Fix JSX prop autocomplete special case

zth:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/962


cristianoc (approved):
good catch

PR #987, opened 2024-05-25
Code action: Expand catch all variant

zth:
code-action-expand-catch-all

Code action for expanding catch-all patterns. Useful when you want to go from a catch all to get exhaustive matching when you add new variant members, for example.

Works on:
- Variant
- Polyvariant
- Option of variant
- Option of polyvariant

Those are the ones I thought made sense to support without this snowballing in complexity.


cristianoc (approved):
wow

PR #989, opened 2024-05-27
Code actions for removing dead code

zth:
- For the dead item itself
- Removing all dead items in the current file

Later on it'd be nice to have a command to run that can remove all of the dead code in a project. This might be better to do in `reanalyze` itself though? Or we can have an action in the editor that uses the editor to apply the edits. That would be easy to do as a first step. Although it'd only work in VSCode at first.

### DIFF
--- a/client/src/commands/code_analysis.ts
+++ b/client/src/commands/code_analysis.ts
@@ -118,7 +118,12 @@ let resultsToDiagnostics = (
         }
       }
 
-      if (item.message.endsWith(" is never used")) {
+      // This heuristic below helps only target dead code that can be removed
+      // safely by just removing its text.
+      if (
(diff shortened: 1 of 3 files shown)


client/src/commands/code_analysis.ts
```diff
@@ -119,6 +119,42 @@ let resultsToDiagnostics = (
           }
         }
       }
+
+      if (item.message.endsWith(" is never used")) {
```
zth, line 123: Staying safe here and only targeting this message since these items seem to be the only ones you can just straight up remove via the reported range, without risking breaking the syntax (like if you'd remove a single variant constructor but not `|`, etc).

cristianoc:
Removing types is much more involved, and unclear how to do automatically. E.g. one variant case.

PR #990, opened 2024-05-27
preserve URI format in `canonicalizeUri`

WhyThat:
This PR addresses a bug in the `canonicalizeUri` function that was causing issues with the `go to definition` feature in Neovim.


fhammerschmidt:
thanks, this was what I was actually missing, now I can do a little cleanup

PR #992, opened 2024-05-27
Remove workaround for canonicalize function in tests

fhammerschmidt:
With the fix from #990 tests run now fine without the workaround


zth:
@fhammerschmidt doesn't compile, a reference that needs to be removed.

fhammerschmidt:
woops, just wondered why it works for me but I missed to add that file to staging, now it should work

zth:
Awesome, thanks!

zth (approved):

PR #999, opened 2024-05-31
Fix issue where pattern completion would not be triggered

zth:
Forgot to check for the cursor position when initiating pattern completion in a certain case.


zth:
Merging this before review as its a simple small fix and I need it to continue.

PR #1000, opened 2024-05-31
Cache project config on demand

zth:
Experimental support for caching the project config to speed up analysis.

It should be good enough to merge and get started testing, and we can then polish before making it the default.

You enable it via the `rescript.settings.cache.projectConfig.enabled` extension config option. It'll then watch `build.ninja` and build a cache of (some of) the project config, that the analysis can then read as needed and avoid doing a bunch of work.

In larger repos this has shown latency improvements of +8-9x. So a massive improvement.

Running it as experimental for a while to get some feedback, and then hopefully something that'll be easy and risk free to enable for everyone.

### DIFF
--- a/analysis/bin/main.ml
+++ b/analysis/bin/main.ml
@@ -110,6 +110,11 @@ let main () =
         path line col
   in
   match args with
+  | [_; "cache-project"; rootPath] -> (
+    let uri = Uri.fromPath rootPath in
+    match Packages.getPackage ~uri with
+    | Some package -> Cache.cacheProject package
+    | None -> ())
   | [_; "completion"; path; line; col; currentFile] ->
     printHeaderInfo path line col;
     Commands.completion ~debug ~path
--- a/analysis/bin/main.ml
+++ b/analysis/bin/main.ml
@@ -111,13 +111,13 @@ let main () =
   in
   match args with
   | [_; "cache-project"; rootPath] -> (
-    Cfg.useProjectConfigCache := false;
+    Cfg.readProjectConfigCache := false;
     let uri = Uri.fromPath rootPath in
     match Packages.getPackage ~uri with
     | Some package -> Cache.cacheProject package
     | None -> print_endline "\"ERR\"")
   | [_; "cache-delete"; rootPath] -> (
-    Cfg.useProjectConfigCache := false;
+    Cfg.readProjectConfigCache := false;
     let uri = Uri.fromPath rootPath in
     match Packages.findRoot ~uri (Hashtbl.create 0) with
     | Some (`Bs rootPath) -> (
--- a/analysis/bin/main.ml
+++ b/analysis/bin/main.ml
@@ -111,10 +111,22 @@ let main () =
   in
   match args with
   | [_; "cache-project"; rootPath] -> (
+    Cfg.useProjectConfigCache := false;
     let uri = Uri.fromPath rootPath in
     match Packages.getPackage ~uri with
     | Some package -> Cache.cacheProject package
-    | None -> ())
+    | None -> print_endline "\"ERR\"")
+  | [_; "cache-delete"; rootPath] -> (
+    Cfg.useProjectConfigCache := false;
+    let uri = Uri.fromPath rootPath in
+    match Packages.findRoot ~uri (Hashtbl.create 0) with
+    | Some (`Bs rootPath) -> (
+      match BuildSystem.getLibBs rootPath with
+      | None -> print_endline "\"ERR\""
+      | Some libBs ->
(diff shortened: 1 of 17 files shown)


server/src/server.ts
```diff
@@ -205,20 +205,49 @@ let sendCompilationFinishedMessage = () => {
...
-    if (config.extensionConfiguration.inlayHints?.enable === true) {
-      sendInlayHintsRefresh();
-    }
-    if (config.extensionConfiguration.codeLens === true) {
-      sendCodeLensRefresh();
+  .on("all", (_e, changedPath) => {
+    if (changedPath.includes("build.ninja")) {
+      let projectRoot = utils.findProjectRootOfFile(changedPath);
+      if (projectRoot != null) {
+        syncProjectConfigCache(projectRoot);
```
jfrolich, line 240: maybe make this conditional on the setting, so potential issues of writing this file would be contained to having the setting on.

server/src/server.ts
```diff
@@ -205,20 +205,49 @@ let sendCompilationFinishedMessage = () => {
...
-  .on("all", (_e, _changedPath) => {
-    sendUpdatedDiagnostics();
-    sendCompilationFinishedMessage();
-    if (config.extensionConfiguration.inlayHints?.enable === true) {
-      sendInlayHintsRefresh();
-    }
-    if (config.extensionConfiguration.codeLens === true) {
-      sendCodeLensRefresh();
+  .on("all", (_e, changedPath) => {
+    if (changedPath.includes("build.ninja")) {
```
jfrolich, line 237: I think we can make it only listen to `.compiler.log` this file will change for every compile run

server/src/server.ts
zth: Definitely!

zth: Although the watch wouldn't be set up unless the setting is on.

jfrolich: Fair point :)

analysis/src/Packages.ml
```diff
@@ -44,165 +44,167 @@ let newBsPackage ~rootPath =
...
+            in
+            (projectFiles, dependenciesFiles, pathsForModule)
+        in
+        Some
+          (let opens_from_namespace =
+             match namespace with
+             | None -> []
+             | Some namespace ->
+               let cmt = Filename.concat libBs namespace ^ ".cmt" in
+               Hashtbl.add pathsForModule namespace (Namespace {cmt});
```
cristianoc, line 114: When the cache file is read, `pathsForModule` will already have been populated when this line was executed when creating the cache, and here it is added again.
No consequence I can think of, but good practice to make sure things are not mutated after caching them.

zth: This should be fine because this is in the non-cached case, right?

zth: Good catch. Let me think about how to make that better.

zth: Maybe it's fine to leave in for now. I can't think of a consequence either. Maybe doing a Hashtbl.replace instead of add makes sense though.

analysis/bin/main.ml
```diff
@@ -110,6 +110,23 @@ let main () =
         path line col
   in
   match args with
+  | [_; "cache-project"; rootPath] -> (
+    Cfg.useProjectConfigCache := false;
```
cristianoc, line 114: Nit: this can perhaps be called `readProjectConfigCache`.

analysis/bin/main.ml
```diff
@@ -110,6 +110,23 @@ let main () =
...
+    let uri = Uri.fromPath rootPath in
+    match Packages.getPackage ~uri with
+    | Some package -> Cache.cacheProject package
+    | None -> print_endline "\"ERR\"")
+  | [_; "cache-delete"; rootPath] -> (
+    Cfg.useProjectConfigCache := false;
```
cristianoc, line 120: This does not seem to be used?

cristianoc (approved):
Seems good to go for people to try as an experimental feature.
Perhaps I'd add a short description in the PR's description that this watches `build.ninja` and uses it to update a cache of files in the project sources and dependencies.

analysis/bin/main.ml
zth: It's called here: https://github.com/rescript-lang/rescript-vscode/pull/1000/files#diff-4398e3b2ab3e7e7e70a93409fa893abf68608d45a5ad1ac7faaf888dac77005dR220-R228

cristianoc: Sorry I meant to say the line `Cfg.useProjectConfigCache := false` does not seem to be necessary, as none of the code below is affected by it.

Issue #961, opened 2024-03-18
autocomplete slow


tshemeng:
VSCdoe ReScript extension autocomplete very slow.

![vscode-rescript-autocomplete-slow](https://github.com/rescript-lang/rescript-vscode/assets/8343797/8157a90e-6742-412c-a2c6-af5122de883c)

env:
- VSCode 1.87.2 (system setup)
- ReScript extension: v1.50.0
- Windows 10 x64

PR #1001, opened 2024-06-04
perf(language-server): memoize possibly heavy IO utils

cometkim:
This could fixes the LSP latency issue

I noticed that caused by mainly spawning multiple `rescript -v` processes on every single requests, but it could be extreamly slow if there is some process monitor (e.g. antivirus) on the environment.

Ideally, we would create a context object with the LSP server lifespan, but I added simple memoization to make fix easier.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -10,6 +10,7 @@
 			"license": "MIT",
 			"dependencies": {
 				"chokidar": "^3.5.1",
+				"memize": "^2.1.0",
 				"vscode-jsonrpc": "^8.0.1",
 				"vscode-languageserver": "^8.0.1",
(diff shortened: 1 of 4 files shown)


cometkim:
Maybe it fixes #961 too

zth:
@cometkim great find, thank you! Here's the other part of the story - reducing latency in the analysis binary itself: https://github.com/rescript-lang/rescript-vscode/pull/1000

This is definitely what we need to do. But, I think we need to think another lap on how and where to memoize. I'll spend some time soon to think about where to do this.

zth:
Testing alternative approach here: https://github.com/rescript-lang/rescript-vscode/pull/1003

PR #1002, opened 2024-06-04
Fix special case in switch expr completion

zth:
Fixes a special case.


analysis/src/CompletionFrontEnd.ml
```diff
@@ -587,6 +587,14 @@ let completionWithParser1 ~currentFile ~debug ~offset ~path ~posCursor
         if Debug.verbose () && debugTypedCompletionExpr then
           print_endline "[typedCompletionExpr] No cases, rest";
         ()
+      | Pexp_match (expr, [{pc_lhs; pc_rhs}])
```
cristianoc, line 590: Perhaps add a comment here as the case is similar to what's in the comment for the next case.

cristianoc (approved):

zth:
Note to self: This logic can probably be cleaned up a little bit in the future.

PR #1003, opened 2024-06-04
Reduce language server latency

zth:
This reduces latency in the language server by caching a few project config related things, so that we most often don't need to do any potentially expensive lookups before we call the analysis, or invoke incremental typechecking.

Known issues with this approach: We need to recompute the ReScript version if it changes. I however don't believe we're watching anything already that can help us figure out when to recompute it. Will need to think a bit about how to do that. It's not a big problem though, because changing the ReScript version in a project does not happen often. And a restart of the language server (or closing all project files) will fix it.

### DIFF
--- a/server/src/projectFiles.ts
+++ b/server/src/projectFiles.ts
@@ -10,6 +10,8 @@ interface projectFiles {
   filesWithDiagnostics: Set<string>;
   filesDiagnostics: filesDiagnostics;
   rescriptVersion: string | undefined;
+  bscBinaryLocation: string | null;
+  namespaceName: string | null;
 
   bsbWatcherByEditor: null | cp.ChildProcess;
 
--- a/server/src/projectFiles.ts
+++ b/server/src/projectFiles.ts
@@ -1,0 +1,25 @@
+import * as cp from "node:child_process";
+import * as p from "vscode-languageserver-protocol";
+
+export type filesDiagnostics = {
+  [key: string]: p.Diagnostic[];
+};
+
+interface projectFiles {
+  openFiles: Set<string>;
+  filesWithDiagnostics: Set<string>;
+  filesDiagnostics: filesDiagnostics;
+  rescriptVersion: string | undefined;
+
+  bsbWatcherByEditor: null | cp.ChildProcess;
+
+  // This keeps track of whether we've prompted the user to start a build
+  // automatically, if there's no build currently running for the project. We
+  // only want to prompt the user about this once, or it becomes
+  // annoying.
+  // The type `never` means that we won't show the prompt if the project is inside node_modules
+  hasPromptedToStartBuild: boolean | "never";
+}
+
+export let projectsFiles: Map<string, projectFiles> = // project root path
+  new Map();
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -58,15 +58,19 @@ export let findProjectRootOfFile = (
       if (
         foundRootFromProjectFiles == null ||
         rootPath.length > foundRootFromProjectFiles.length
-      )
+      ) {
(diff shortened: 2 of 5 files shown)


zth:
cc @cometkim

zth:
Cc @cristianoc

server/src/projectFiles.ts
```diff
@@ -0,0 +1,27 @@
...
+  [key: string]: p.Diagnostic[];
+};
+
+interface projectFiles {
+  openFiles: Set<string>;
+  filesWithDiagnostics: Set<string>;
+  filesDiagnostics: filesDiagnostics;
+  rescriptVersion: string | undefined;
+  bscBinaryLocation: string | null;
+  namespaceName: string | null;
```
cristianoc, line 14: I guess these 3 are the added cached values.

cristianoc (commented):
What's the perf difference in practice?

In case we're worried about things changing infrequently such as compiler version, one possibility would be to refresh the data from time to time. Which should have no impact on latency.

Anything else to pay attention to in this PR?

zth:
> What's the perf difference in practice?
> 

It seems to vary wildly depending on how the system handles frequently spawning child processes. For me, there's no noticeable difference, but for for example @cometkim and @jonoprest, there's a pretty massive difference at times.

> In case we're worried about things changing infrequently such as compiler version, one possibility would be to refresh the data from time to time (e.g. based on an interval). Which should have no impact on latency.
> 

Yeah this is a good approach. I think what I want us to do soon is what you mentioned in another thread - figure out and write down all of the inputs that might change which the LS needs to react to, and then figure out which of those are worth caring about, and what then to do about them.

> Anything else to pay attention to in this PR?

Not really no, this should be fairly straight forward.

JonoPrest:
I certainly experience delays. I've increased my lsp timeout to 5 seconds and still often hit timeouts with rescript ls. It could be a configuration or setup issue but I don't get these delays on other lsps. Rust analyzer for eg also has to index a project before it can be used with the editor (which can take minutes) but it doesn't freeze the editor or hit any timeouts.

fhammerschmidt:
@JonoPrest did you try with the VSIX of this PR?
https://github.com/rescript-lang/rescript-vscode/actions/runs/9397264258

JonoPrest:
> @JonoPrest did you try with the VSIX of this PR?
> 
> https://github.com/rescript-lang/rescript-vscode/actions/runs/9397264258

I did yes 👍🏼 we've been discussing it on discord. I think there was an improvement but I haven't been able to try the project cache configuration yet.

server/src/utils.ts
```diff
@@ -40,11 +39,41 @@ export let findProjectRootOfFile = (
...
 
+// TODO: races here?
+// TODO: this doesn't handle file:/// scheme
+export let findProjectRootOfFile = (
+  source: p.DocumentUri
+): null | p.DocumentUri => {
+  // First look in project files
+  let foundRootFromProjectFiles: string | null = null;
+
+  for (const rootPath of projectsFiles.keys()) {
```
cometkim, line 55: If I understand correctly here, the most common operation is to find the project info associated with the current open file.

I see you make a `{ [project root path]: [project info] }` index above, but it seems like it's missing the `{ [open file]: [project root path] }` or `{ [open file]: [project info] }` inverted index.

I think we don't need this loop if we're using an index for it in the same way

```ts
export const projectRoots = new Map<p.DocumentUri, string>();
export const projects = new Map<string, projectFiles>();
```

zth: We could do that, but this feels like a good trade off to me. That loop is cheap and it'll also be resilient to if roots change for some reason while the file is open (unlikely, but still).

cristianoc (approved):

PR #1013, opened 2024-06-08
rescript-tools: do not override flags for static linking

cknitt:
Fixes #1006.

PR #1007, opened 2024-06-06
add system requirements section for linux to tools readme

woeps:
put insights of #1006 into tools readme
@zth

Issue #1006, opened 2024-06-05
tools linux binary dynamically links musl


woeps:
Running `npx --package @rescript/tools@0.6.2 -- rescript-tools -v` exits with exit code `0` but gives no output. The same is true for any other tools command.

Runnin `npx --package @rescript/tools@0.5.0 -- rescript-tools -v` exits with exit coce `0` but outputs the expected string `0.5.0`.

I'm running on `linux x86_64` and node 20.14.0.
(My colleague ran the commands on mac with success. It seams like a linux issue...)

Any ideas?

zth:
We recently worked on arm64 support for Linux, maybe we messed something up with the binaries then?

woeps:
I'm trying to find out what's happening:

AFAIK: `npm i @rescript/tools@0.6.2` puts a link into `<project>/node_modules/.bin/rescript-tools` which links to `<project>/node_modules/@rescript/tools/npm/cli.js`. `cli.js` just calls `getBinaryPath` (which in my case returns `<project>/node_modules/@rescript/tools/binaries/linux/rescript-tools.exe` and spwans a process with this binary.
Right?

Trying to run the binary directly in the terminal yields the following error message: `zsh: no such file or directory: <project>/node_modules/@rescript/tools/binaries/linux/rescript-tools.exe`
Running `file <project>/node_modules/@rescript/tools/binaries/linux/rescript-tools.exe` yields
```
<project>/node_modules/@rescript/tools/binaries/linux/rescript-tools.exe: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, BuildID[sha1]=b1b311c6abafbede111905ae90396f8d5caede7d, with debug_info, not stripped
```

zth:
> Alright, the `file` call (see above comment) gave me the final clue to solve my issue: My linux system was **missing** the **`musl` package**.
> 
> After installing this package, tools cli binary runs as intended! 🥳
> 
> Maybe we should state the dependency on musl in the readme?

Yes definitely! That's a terrible experience/error 🙈 would you mind sending a PR? Also, I wonder if there's an alternative to that, and why the compiler doesn't require the same thing...

woeps:
It seems like the alpine docker image uses musl instead of libc (as it is more lightweight) which is dynamically linked and therefore needs to be present on the system.
[...]

PR #1014, opened 2024-06-08
Revert "add system requirements section for linux to tools readme"

woeps:
Reverts rescript-lang/rescript-vscode#1007 because it's obsolete due to #1013 .
releated to #1006


zth:
Thanks!

PR #1023, opened 2024-06-17
Set GITHUB_OUTPUT instead of GITHUB_ENV in bump-version script

fhammerschmidt:
At least that is also how GitHub's own set output action does it, so it's worth a try.
https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L193


zth:
Let's try it.

PR #1035, opened 2024-09-03
:ambulance: - Monorepo hotfix

jfrolich:


server/package.json
```diff
@@ -1,38 +1,38 @@
...
-	"bugs": {
-		"url": "https://github.com/rescript-lang/rescript-vscode/issues"
-	},
-	"dependencies": {
-		"chokidar": "^3.5.1",
-		"vscode-jsonrpc": "^8.0.1",
-		"vscode-languageserver": "^8.0.1",
-		"vscode-languageserver-protocol": "^3.17.1"
-	}
+  "name": "@rescript/language-server",
```
jfrolich, line 2: for some reason this was tab formatted (and in the server it's spaces - I think spaces is more standard)

server/src/utils.ts
```diff
@@ -70,7 +69,7 @@ export let findProjectRootOfFile = (
   } else {
     const isDir = path.extname(source) === "";
     return findProjectRootOfFileInDir(
-      isDir ? path.join(source, "dummy.res") : source
+      isDir && !allowDir ? path.join(source, "dummy.res") : source
```
jfrolich, line 72: This caused the fix to not work :|

zth:
You need to run `npm install` in both folders as well so the version is updated in the lockfile.

nojaf:
Hello @jfrolich, could you describe your folder structure a bit where this originally went wrong. What kind of setup/folder structure did you have where the root could not be found?
I'm looking at `findProjectRootOfFile` and would like to learn why this seems more complicated than I would expect it to be.

PR #1043, opened 2024-10-19
Add additional signature information to doc json

nojaf:
Subset of https://github.com/rescript-lang/rescript-vscode/pull/1041

### DIFF
--- a/tools/src/tools.ml
+++ b/tools/src/tools.ml
@@ -381,7 +381,7 @@ let path_to_string path =
 let valueDetail (typ : Types.type_expr) =
   let rec collectSignatureTypes (typ_desc : Types.type_desc) =
     match typ_desc with
-    | Tlink t -> collectSignatureTypes t.desc
+    | Tlink t | Tsubst t -> collectSignatureTypes t.desc
     | Tconstr (Path.Pident {name = "function$"}, [t; _], _) ->
       collectSignatureTypes t.desc
     | Tconstr (path, ts, _) -> (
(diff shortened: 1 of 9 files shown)


tools/src/tools.ml
```diff
@@ -310,6 +352,60 @@ let typeDetail typ ~env ~full =
...
+      aux p2;
+      Buffer.add_char buf ')'
+  in
+  aux path;
+  Buffer.contents buf
+
+let valueDetail (typ : Types.type_expr) =
+  let rec collectSignatureTypes (typ_desc : Types.type_desc) =
+    match typ_desc with
+    | Tlink t -> collectSignatureTypes t.desc
```
zth, line 384: Should probably do the same as this for `Tsubst` and `Tpoly` as well.

zth (approved):
Let's merge this and try it out as it's purely additive.

nojaf:
Thank you for the review, @zth!

This leaves the PR in a bit of a puzzled state. What is the next action, and who should take it? There is both a remark and an approval. 

My goal is to have a new version of the tool on npm if possible. What does it take to reach that level? As a contributor, I'm left to speculate about what comes next.

zth:
> Thank you for the review, @zth!
> 
> This leaves the PR in a bit of a puzzled state. What is the next action, and who should take it? There is both a remark and an approval.
> 
> My goal is to have a new version of the tool on npm if possible. What does it take to reach that level? As a contributor, I'm left to speculate about what comes next.

Well, I left a comment that I'd like to at least have acknowledged before we can merge.

tools/src/tools.ml
```diff
@@ -310,6 +352,69 @@ let typeDetail typ ~env ~full =
...
+        let ts =
+          ts
+          |> List.concat_map (fun (t : Types.type_expr) ->
+                 collectSignatureTypes t.desc)
+        in
+        [{path = p; genericParameters = ts}])
+    | Tarrow (_, t1, t2, _) ->
+      collectSignatureTypes t1.desc @ collectSignatureTypes t2.desc
+    | Tvar None -> [{path = "_"; genericParameters = []}]
+    | Tpoly (t, ts) -> (
```
nojaf, line 401: @zth I'm not sure about `Tpoly` here.
Could you give an example of what this could be? Is that the thing with the `#type`?

tools/src/tools.ml
nojaf: Try adding those, but unsure about `Tpoly`. Please take a look.

tools/src/tools.ml
nojaf: Thanks!

zth:
Great stuff! Releasing tonight.

PR #1047, opened 2024-10-24
Enable incremental typechecking by default

zth:
Enables incremental typechecking by default by changing the name of the prop (so old configs are invalidated) and setting a new default.

I'm sure there are more things we could do with incremental typechecking to make it a better experience, but at this point it's such a magnitudes better experience even with the small issues it might have. Enabling by default will make the experience better for everyone by default, and if issues surface we can fix them.

This also enables the `projectConfig` cache options, that dramatically reduce latency in certain scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## master
 
+#### :rocket: New Feature
+
+- Enable incremental typechecking by default. https://github.com/rescript-lang/rescript-vscode/pull/1047
+
 ## 1.58.0
 
 #### :bug: Bug fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 #### :rocket: New Feature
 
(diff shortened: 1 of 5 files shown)


zth:
Actually, let's enable project cache config too by default. Updating the PR.

fhammerschmidt:
I never tested the project cache config apparently. Is it stable enough to activate by default already?

zth:
I think so. But it'd be good to give it a spin in your project(s) too of course.

cristianoc (approved):
Great!
Are there any downsides, or unknowns, about turning on the cache?

fhammerschmidt:
I mean that's why we have pre-releases anyway, right? Or do you want to release this as stable soon?

fhammerschmidt (approved):

zth:
We can do pre-release for a while, sure. Maybe a week or so. But then the big benefit comes from shipping this to stable.

PR #1059, opened 2024-12-01
Allow `esmodule` as a "module" option when compiling path

cbowling:
Updates `ModuleFormat` to understand latest `rescript.json` `"module:" "esmodule"` config. This allows a user to use the `ReScript: Open the compiled JS file for this implementation file` option without it default looking for `/lib/js/src` instead of `/lib/es6/src`.

### DIFF
--- a/server/src/buildSchema.ts
+++ b/server/src/buildSchema.ts
@@ -15,6 +15,7 @@ export enum ModuleFormat {
   Commonjs = "commonjs",
   Es6 = "es6",
   Es6Global = "es6-global",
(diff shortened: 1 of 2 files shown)


zth:
@cbowling thank you very much for this! @fhammerschmidt I think you originally implemented this, could you have a look?

fhammerschmidt:
Looks fine and I am not sure why I missed this one two weeks ago, sorry!

fhammerschmidt (approved):

PR #1062, opened 2024-12-19
Additional completion tests

nojaf:
Hi @zth, I've added a few more test cases based on my last project I'm working on:

- `RxjsCompletion.res` it should complete from the parent module
- `CompletionPipeProperty.res` already works, could be interested to add for regression sake.
- `CompletionMultipleEditorCompleteFrom.res` didn't work a single attribute either. Seems interesting to have regression-wise.

### DIFF
--- a/analysis/tests/src/CompletionMultipleEditorCompleteFrom.res
+++ b/analysis/tests/src/CompletionMultipleEditorCompleteFrom.res
@@ -1,0 +1,22 @@
+@@warning("-26")
+@@warning("-27")
+@@warning("-110")
+
+module A = {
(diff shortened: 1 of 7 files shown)


cristianoc:
These have moved to the compiler.
@zth should we delete this code from this repo now?

zth:
No we need to keep it around for the foreseeable future, the extension will still ship this code and we need to be able to make fixes to it too since it affects all currently released versions of ReScript.

This particular PR is about the dot-completion feature, which I'll port to the compiler repo once we've merged it. So it's fine.

PR #1069, opened 2025-01-22
Build without docker, use GitHub Linux ARM runner

cknitt:
* No need for Docker anymore
* Get rid of BuildJet runner

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,57 +11,7 @@ concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


cknitt:
> What is like the tl;dr of BuildJet?

We were just using it because it was offering hosted Linux ARM 64 GitHub action runners which GitHub themselves did not offer at the time.

It is a paid service (but not really very expensive for our usage). I see no reason to continue using it. The new GitHub Linux ARM runner is free, and seems faster than the BuildJet runner we were using, too.

zth (approved):

PR #1078, opened 2025-02-26
Try to fix latest-master release

fhammerschmidt:


fhammerschmidt:
Wait that seems not to be an issue actually. The artifacts update fine at least, it is just the date that is old.

cknitt (approved):

PR #1080, opened 2025-03-28
Use fs api to check if string is a directory.

nojaf:
I found out why the incremental build wasn't working yesterday.
Turns out I have a dot in my project folder name, this was mistaken for a file.
That file is not part of the project, thus the rescript version could not be found for the root.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -268,7 +268,7 @@ let openedFile = (fileUri: string, fileContent: string) => {
         filesDiagnostics: {},
         namespaceName:
           namespaceName.kind === "success" ? namespaceName.result : null,
-        rescriptVersion: utils.findReScriptVersion(projectRootPath),
+        rescriptVersion: utils.findReScriptVersionForProjectRoot(projectRootPath),
         bsbWatcherByEditor: null,
         bscBinaryLocation: utils.findBscExeBinary(projectRootPath),
         editorAnalysisLocation: utils.findEditorAnalysisBinary(projectRootPath),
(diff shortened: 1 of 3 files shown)


nojaf:
Ping @zth

server/src/utils.ts
```diff
@@ -67,7 +67,8 @@ export let findProjectRootOfFile = (
   if (foundRootFromProjectFiles != null) {
     return foundRootFromProjectFiles;
   } else {
-    const isDir = path.extname(source) === "";
+    const dirStat = fs.statSync(source);
+    const isDir = dirStat.isDirectory();
```
zth, line 71: IIRC, this is in a hot path where we shouldn't use `fs` functions (which is why this fn looks the way it looks now). So we need to figure out another way to do this, outside of the hot path. OTOH maybe it could be aggressively cached in some sane way (unlikely), or moved outside of the hot path entirely to some sort of setup function somewhere. Would need to be explored.

zth (changes requested):

server/src/server.ts
```diff
@@ -268,7 +268,7 @@ let openedFile = (fileUri: string, fileContent: string) => {
         filesDiagnostics: {},
         namespaceName:
           namespaceName.kind === "success" ? namespaceName.result : null,
-        rescriptVersion: utils.findReScriptVersion(projectRootPath),
+        rescriptVersion: utils.findReScriptVersionForProjectRoot(projectRootPath),
```
nojaf, line 271: This is the root of the problem really.
Here my project is called `ronnies.be`, which is mistaken for a file.
At this point, we know it is a folder, so we can shortcut things I think.

zth:
@nojaf ready for a new review...?

nojaf:
@zth yup, go for it

zth (approved):
Looks fine to me! Although I have not tested it.

nojaf:
Works like a charm in the prerelease channel!

PR #1082, opened 2025-04-08
Fix local compilation when in a workspace

jfrolich:
If you compile a package inside a workspace, that package needs to be considered the root instead of the whole workspace when using rewatch.


zth (approved):
Looking forward to cleaning all of this up once Rewatch is the main build system.

Issue #1079, opened 2025-03-12
Doesn't work with standalone bun 1.2.5


iwdt:
When I use bun 1.2.5, it gives an error related to the absence of node.js

- OS: Debian 12
  ```bash
  $ cat /etc/os-release 
  PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
  NAME="Debian GNU/Linux"
  VERSION_ID="12"
  VERSION="12 (bookworm)"
  VERSION_CODENAME=bookworm
  ID=debian
  HOME_URL="https://www.debian.org/"
  SUPPORT_URL="https://www.debian.org/support"
  BUG_REPORT_URL="https://bugs.debian.org/"
  ```

- Bun: 1.2.5, installed via
  ```bash
  $ curl -fsSL https://bun.sh/install | bash -s "bun-v1.2.5"
  ```
- Extension: chenglou92.rescript-vscode 1.62.0
- Node.js not installed 
- Package.json:
```json
{
  "name": "my_app",
  "private": true,
  "scripts": {
    "build:res": "NODE_ENV=production rescript",
    "build:js": "NODE_ENV=production bun config/bun.config.js",
    "build": "bun run build:res && bun run build:js",
    "build:css": "NODE_ENV=production postcss --config ./config/postcss.config.js ./app/assets/stylesheets/application.css -o ./app/assets/builds/application.css",
    "dev:res": "bun run build:res -w",
    "dev:js": "bun run build:js -w",
    "dev:css": "bun run build:css -w"
  },
  "dependencies": {
    "react": "^19.0.0",
    "react-dom": "^19.0.0"
  },
  "devDependencies": {
    "@rescript/core": "^1.6.1",
    "@rescript/react": "^0.13.1",
    "reanalyze": "^2.23.0",
    "rescript": "^11.1.4",
    "autoprefixer": "^10.4.21",
    "cssnano": "^7.0.6",
    "postcss": "^8.5.3",
    "postcss-cli": "^11.0.0",
    "postcss-import": "^16.1.0",
    "postcss-nesting": "^13.0.1"
  },
  "trustedDependencies": [
    "reanalyze",
    "rescript"
  ]
}

```

Got error:

<img width="549" alt="Image" src="https://github.com/user-attachments/assets/b99d1f95-292e-4e6f-831d-e8b486d2eb28" />

But if I fix it manually in the file node_modules/.bin/rescript first line:

from
```sh
#!/usr/bin/env node
```
to

[...]

PR #1083, opened 2025-04-14
Use `bsc.exe` to figure out rescript version

nojaf:
Fixes https://github.com/rescript-lang/rescript-vscode/issues/1079

I'm also in a little weird position where I use `nvm` on Mac, and somehow node is not always on my path anymore. This worked in the past, but is somehow broken for me, so I took a look in how I can get the LSP server working without depending on Node being present.

Turns out, the only thing I bump into is the initial rescript version call.
That will run `node_modules/.bin/rescript` which in the latest alpha has

```
#!/usr/bin/env node
...
```

For all the other stuff where we shell out to `bsc`, `rescript-editor-analysis`, we target the platform specific executable.

So, in this PR, I'm changing the version check to do the same. `bsc.exe -v` also produces the version we are after, prepended with `ReScript`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 - Fix: bug where incremental compilation crashes when rewatch is being run in a specific package vs the root of the monorepo. https://github.com/rescript-lang/rescript-vscode/pull/1082
 
+- Fix: Absence of Node.js does not hinder LSP server. https://github.com/rescript-lang/rescript-vscode/pull/1083
+
 ## 1.62.0
 
 #### :nail_care: Polish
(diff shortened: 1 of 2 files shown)


zth:
I have a hunch there might be a reason we used the `rescript` bin instead, but can't fully remember. Could you check what older ReScript versions gives you when you call that command? ReScript 9 & 10.

nojaf:
```
ReScript 11.1.4
ReScript 10.1.4
ReScript 9.1.4
```

all works out

zth:
Great, thanks for checking! Then just a changelog and we're good to go.

PR #1085, opened 2025-04-28
Require whitespace before `as` in `to`/`downto` labels

mediremi:
The grammar rule `RE_TO_DOWNTO_AS_LABELS` contains two patterns for ensuring that labelled arguments with names `to` and `downto` are highlighted as variables and not keywords.

The second pattern, `(to|downto)\s*(as)`, is supposed to match cases like `(~to as to2) => ...`, but was also matching e.g. `~toast`.

I've fixed this by changing `\s*` to `\s+`.

| Before | After |
| -- | -- |
| image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 #### :bug: Bug fix
 
+- Fix: incorrect highlighting of `as` inside labelled arguments like `toast` https://github.com/rescript-lang/rescript-vscode/pull/1085
+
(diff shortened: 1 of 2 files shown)


mediremi:
I've also prefixed these patterns with `~` to make it clearer that they're targeting labelled arguments as well as to narrow their scope.

DZakh (approved):
Nice catch 👍

mediremi:
@zth changelog entry added

zth:
@mediremi awesome. Thank you for your contribution! 🎉

PR #1086, opened 2025-04-29
As part of incremental type checking, when filtering out potentially unwanted parser errors, don't ignore the multiple definitions type error

mediremi:
When using incremental type checking, if the compiler reports the multiple definitions type error, the user will see a warning notification:

image

The reason why this happens is as follows:

- `incrementalCompilation.compileContents` filters out diagnostic messages parsed from the compiler’s `stderr` that either:
  i) start with `Uninterpreted extension 'rescript.` or,
  ii) include the source file name

https://github.com/rescript-lang/rescript-vscode/blob/1bf87b386e6331ca85459df3bc03b8e4c3fc50ab/server/src/incrementalCompilation.ts#L677-L687

- If all the diagnostic messages extracted from the compiler’s output are filtered out due to matching at least one of the conditions above, then `res.length == 0` and `hasIgnoredErrorMessages = false`
- This then leads to a notification being shown telling the user that something may have gone wrong with incremental type checking
- Unfortunately, the logic for this doesn’t handle one edge case: [the multiple definitions type error](https://github.com/rescript-lang/rescript/blob/155ab6ee2a2609ce19234f032ee88df3109a17ec/compiler/ml/typemod.ml#L1851-L1855) (`Multiple definition of the <kind> name <name>`) [prints out the LOC of the repeated definition](https://github.com/rescript-lang/rescript/blob/155ab6ee2a2609ce19234f032ee88df3109a17ec/compiler/ml/typemod.ml#L1853-L1855)
- By printing out the full LOC, including source file name, this triggers condition number 2 of the filter: ``!d.message.includes(`/${INCREMENTAL_FOLDER_NAME}/${entry.file.sourceFileName}`)``
- Which then triggers the error notification

This PR fixes this issue by adding an exception to the filter for error messages that include the string `"Multiple definition of the"`.


nojaf:
Great work on these last PRs @mediremi!

mediremi:
Thank you!

PR #1092, opened 2025-05-11
Find per-platform binaries for v12.0.0-alpha.13+

mediremi:
https://github.com/rescript-lang/rescript/pull/7395 split binaries into optional platform-specific dependencies (e.g. `@rescript/linux-x64`).

I've updated `utils.findPlatformPath` (used by `utils.findBscExeBinary` and `utils.findEditorAnalysisBinary`) to look for these platform-specific dependencies if `node_modules/rescript/{platform}` doesn't exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## master
 
+#### :rocket: New Feature
+
+- Find `bsc.exe` and `rescript-code-editor-analysis.exe` from platform-specific packages used by ReScript `v12.0.0-alpha.13`+.https://github.com/rescript-lang/rescript-vscode/pull/1092
+
 #### :bug: Bug fix
 
 - Fix: bug where we incorrectly showed a warning notification about something going wrong with incremental type checking, when in fact the compiler was reporting module-related type errors https://github.com/rescript-lang/rescript-vscode/pull/1090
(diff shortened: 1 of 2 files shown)


zth:
Thank you! 😄 

@cometkim can you have a quick look at this one?

cometkim:
Can't we use `rescript` as an API?

If we already know the directory, we can use it as an API like:

```js
const { binPaths } = await import('../path/to/rescript package')
binPaths.bsc_exe // should point to the binary path
```

mediremi:
The reason why I didn't go with the dynamic import approach is that `findPlatformPath` and its upstreams (e.g. `utils.findBscExeBinary->server.openedFile->server.onMessage`) are all synchronous functions and that making them async would likely be an extensive refactor.

I think that copying the import path logic of `@rescript/${process.platform}-${process.arch}` from `cli/common/bins.js` is a good pragmatic short-term solution, and a refactor to make the relevant functions async can be done later.

zth:
@mediremi would it be possible for you to do a quick assessment on how much work it'd be to do the async refactor? At some point we'll need to bite the bullet for sure. But whether it's best to do it now or later is not clear to me right now.

mediremi:
I'll look into it at some point today :+1:

nojaf:
Let's merge this one and open a follow-up issue?

cometkim:
VSCode extension finally supports ESM. I just noticed it.

If it needs a huge refactoring, let me know 😉

zth:
> Let's merge this one and open a follow-up issue?

Yeah, we can merge this first for sure.

mediremi:
Here's my attempt at an async rewrite: https://github.com/rescript-lang/rescript-vscode/pull/1093

PR #1093, opened 2025-05-15
[WIP] async find binary

mediremi:


server/src/utils.ts
```diff
@@ -74,21 +76,72 @@ export let findProjectRootOfFile = (
...
+  if (binary == "rescript") {
+    // Can't use the native bsb/rescript since we might need the watcher -w
+    // flag, which is only in the JS wrapper
+    binaryPath = path.join(rescriptDir, rescriptJSWrapperPath)
+  } else if (semver.gte(rescriptVersion, "12.0.0-alpha.13")) {
+    // TODO: export `binPaths` from `rescript` package so that we don't need to
+    // copy the logic for figuring out `target`.
+    const target = `${process.platform}-${process.arch}`;
+    const targetPackagePath = path.join(rescriptDir, "..", `@rescript/${target}/bin.js`)
+    const { binPaths } = await import(targetPackagePath);
```
mediremi, line 119: Unfortunately I couldn't just do `import(rescriptDir)` to get bin paths as the `rescript` package only exports `./lib/` and `./package.json` in the `exports` field of its `package.json`.

We probably either want to export `./common/bins.js` or have a default export that reexports `binPaths`.

As for `@rescript/${target}`, I had to append `/bin.js` to the import path for it to work as, like `rescript`, the per-platform binary packages have no default export.

server/src/utils.ts
```diff
@@ -7,7 +7,9 @@ import {
   ResponseMessage,
 } from "vscode-languageserver-protocol";
 import fs from "fs";
+import fsAsync from "fs/promises";
```
mediremi, line 10: In a later PR we might want to replace sync `fs` functions with their promise-based alternatives.

package.json
```diff
@@ -255,9 +255,10 @@
   },
   "devDependencies": {
     "@types/node": "^14.14.41",
+    "@types/semver": "^7.7.0",
```
mediremi, line 258: I added `semver` to make parsing the rescript version easier.

If adding this dependency is fine, then I can also refactor [`incrementalCompilation.figureOutBscArgs`](https://github.com/rescript-lang/rescript-vscode/blob/f4ba82955050e66bd678139f0a6e7888251219e1/server/src/incrementalCompilation.ts#L576) and [`utils.runAnalysisAfterSanityCheck`](https://github.com/rescript-lang/rescript-vscode/blob/f4ba82955050e66bd678139f0a6e7888251219e1/server/src/utils.ts#L211-L220) to use it.

https://github.com/rescript-lang/rescript-vscode/blob/f4ba82955050e66bd678139f0a6e7888251219e1/server/src/incrementalCompilation.ts#L576

https://github.com/rescript-lang/rescript-vscode/blob/f4ba82955050e66bd678139f0a6e7888251219e1/server/src/utils.ts#L211-L220

server/src/utils.ts
cometkim: Oh, you're right. I didn't re-export it because I wasn't sure the API from the compiler (it will be the first programmatic API, so...)

cometkim: > As for @rescript/${target}, I had to append /bin.js to the import path for it to work as, like rescript, the per-platform binary packages have no default export.

There is the `exports` field (which always has precedence over the `main` field). Importing the binary package without the `/bin.js` suffix should also work.

cometkim:
VSCode supports ESM-based extensions since very recently.
https://code.visualstudio.com/updates/v1_100#_esm-support-for-extensions

If we bump the target version and use fewer legacy dependencies, we can entirely switch to ESM.

Which accepts a fully asynchronous program, including top-level await, without bundler magic or TDZ-related bugs.

server/src/utils.ts
```diff
@@ -74,21 +76,72 @@ export let findProjectRootOfFile = (
...
+  } catch (error) {
+    return null
+  }
+
+  let binaryPath: string | null = null
+  if (binary == "rescript") {
+    // Can't use the native bsb/rescript since we might need the watcher -w
+    // flag, which is only in the JS wrapper
+    binaryPath = path.join(rescriptDir, rescriptJSWrapperPath)
+  } else if (semver.gte(rescriptVersion, "12.0.0-alpha.13")) {
```
cometkim, line 114: AFAIK, npm's default semver comparison doesn't work intuitively with pre-releases. This might need the `includePrelease` flag.

https://www.npmjs.com/package/semver#prerelease-tags

server/src/utils.ts
mediremi: > Don't forget to use the right compilerOptions.nodeResolution option in the tsconfig.json. Use NodeNext as long as possible, or Bundler.

Ah I didn't realise we were bundling as CJS, this explains the error messages I was getting when trying to import without `bin.js`. I'll have a go at tweaking `compilerOptions.nodeResolution` to make this work :+1:

server/src/utils.ts
mediremi: You're right that we'll need that flag for range matching, so I added it in another place where we're checking major version numbers:

https://github.com/rescript-lang/rescript-vscode/blob/22bff6d02b2c3ac35f8a6f86bd169ed16e8d3f87/server/src/incrementalCompilation.ts#L580

For `semver.gte(rescriptVersion, "12.0.0-alpha.13")` it won't be needed as we're not doing a range comparison.

package.json
mediremi: I've done the proposed refactoring in this commit: https://github.com/rescript-lang/rescript-vscode/pull/1093/commits/d4cc19db8509a9f0273600fc955a839cfa7b0712

server/src/utils.ts
mediremi: I tried the following:

- `compilerOptions.moduleResolution` values `NodeNext` and `Bundler`
- Updating typescript to v5.8.3
- ES2022 instead of ES1029

but I still can't get it to work :(

package.json
zth: Great! 👍

server/src/utils.ts
```diff
@@ -74,21 +76,72 @@ export let findProjectRootOfFile = (
...
 
-// Check if binaryName exists inside binaryDirPath and return the joined path.
-export let findBinary = (
-  binaryDirPath: p.DocumentUri | null,
-  binaryName: string
-): p.DocumentUri | null => {
-  if (binaryDirPath == null) {
+// If ReScript < 12.0.0-alpha.13, then we want `{project_root}/node_modules/rescript/{c.platformDir}/{binary}`.
+// Otherwise, we want to dynamically import `{project_root}/node_modules/rescript` and from `binPaths` get the relevant binary.
+// We won't know which version is in the project root until we read and parse `{project_root}/node_modules/rescript/package.json`
```
zth, line 81: 👍 Good comment!

server/src/utils.ts
zth: This is still WIP, right? Or could we merge this and do the above in a follow up?

zth (approved):
Looks great, thank you very much! Let me know if this is good to merge as is or if there are ongoing things that need to be handled first.

mediremi:
Thanks :smile: I'm happy for this to be merged as-is. We can try to figure out how to improve the dynamic import path in a followup PR :+1:

server/src/utils.ts
mediremi: We can improve the dynamic import path in a followup PR 👍

mediremi:
Changelog entry added :+1:

zth:
Amazing work @mediremi!

PR #1103, opened 2025-06-19
Show existing compiler warnings and errors on file open

mediremi:
Currently, opening a ReScript file in VS Code does not show existing compiler warnings and errors until a file change triggers `onWorkspaceDidChangeWatchedFiles`, which in turn calls `sendUpdatedDiagnostics`.

This fix calls `sendUpdatedDiagnostics` on file open, ensuring that users will see existing warnings & errors immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 
 #### :bug: Bug fix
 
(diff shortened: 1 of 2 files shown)


zth (approved):

nojaf:
Good catch, I think I've experienced this. Thanks

PR #1112, opened 2025-07-30
Prevent infinite recursion in `TypeUtils.resolveTypeForPipeCompletion` by not looping over type variables

mediremi:
Backport of https://github.com/rescript-lang/rescript/pull/7731

I've tested this fix on a local v11 project where `analysis codeAction` will hang without the fix and exists fine with the fix applied.


mediremi:
ping @zth

PR #1114, opened 2025-08-02
Dont create empty diagnostic messages

mediremi:
VS Code throws a `TypeError: message not sent` error if a diagnostic message is blank:

https://github.com/microsoft/vscode/blob/3aef8be5e46aa7e97e8b0ed115f6147cd0b24634/src/vs/workbench/api/common/extHostTypes.ts#L1183-L1185

We create diagnostics with empty messages if `diagnosticMessage` is an empty array, which happens if `parsedDiagnostic.content` is empty or has only 1 element:

https://github.com/rescript-lang/rescript-vscode/blob/ef518a9ae69caf1b398204eaddf13517aa82ae3c/server/src/utils.ts#L699-L707

The compiler log message `FAILED: cannot make progress due to previous errors.` is converted to a diagnostic message with `content: [line]`, which triggers this condition:

https://github.com/rescript-lang/rescript-vscode/blob/5b76419e99288a7a19cc57a3228ec99a6e398782/server/src/utils.ts#L571-L578

In addition to adding a check for `message != ""` before creating a diagnostic message, I've also changed the compiler log output parser to skip the following errors messages:
- `FAILED: cannot make progress due to previous errors`
- `FAILED: dependency cycle`

These errors don't currently contain filepath/range information and so we cannot create diagnostic messages for them anyway (though with some tweaking on the compiler side, we might be able to output filepaths for dependency cycle errors.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 - Fix: `rescript-editor-analysis.exe semanticTokens` sometimes returned invalid JSON, which affected syntax highlighting. https://github.com/rescript-lang/rescript-vscode/pull/1113
 
- - Fix: hang in `rescript-editor-analysis.exe codeAction` that sometimes prevented ReScript files from being saved in VS Code. https://github.com/rescript-lang/rescript-vscode/pull/1112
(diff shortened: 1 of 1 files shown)


zth:
Side note - outputting dependency issues in the compiler would be a massive (and necessary) improvement.

PR #1116, opened 2025-08-11
Try/catch on trying to read compiler log

zth:
This sometimes blows up, if the compiler log does not exist for some reason. Simple try/catch to protect.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -60,14 +60,20 @@ let codeActionsFromDiagnostics: codeActions.filesCodeActions = {};
 // will be properly defined later depending on the mode (stdio/node-rpc)
 let send: (msg: p.Message) => void = (_) => {};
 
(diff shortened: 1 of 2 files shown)


nojaf (approved):

server/src/server.ts
```diff
@@ -99,7 +99,12 @@ let sendUpdatedDiagnostics = async () => {
   for (const [projectRootPath, projectFile] of projectsFiles) {
     let { filesWithDiagnostics } = projectFile;
     let compilerLogPath = path.join(projectRootPath, c.compilerLogPartialPath);
-    let content = fs.readFileSync(compilerLogPath, { encoding: "utf-8" });
+    let content = "";
+    try {
+      content = fs.readFileSync(compilerLogPath, { encoding: "utf-8" });
```
nojaf, line 104: Is an existence check worth it here?

Issue #1056, opened 2024-11-17
The module or file RescriptCore can't be found.


BIYUEHU:
Report details:

```txt
The module or file RescriptCore can't be found.
- If it's a third-party dependency:
  - Did you add it to the "bs-dependencies" or "bs-dev-dependencies" in bsconfig.json?
- Did you include the file's directory to the "sources" in bsconfig.json?
```

- Extension version: `v1.58.0`
- I created a basic project by `create-rescript-app` and already installed all packages.
- There isn't non-english Characters at the project path: `E:\repos\my-res`

Part of `Package.json`:

```json
  "dependencies": {
    "@rescript/core": "^1.3.0",
    "rescript": "^11.1.0"
  }
```

`Visual Studio Code` information:

```txt
Version: 1.85.2 (user setup)
Commit: 8b3775030ed1a69b13e4f4c628c612102e30a681
Date: 2024-01-18T06:40:10.514Z(10 mos ago)
Electron: 25.9.7
ElectronBuildld:26354273
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
Os: Windows NT x64 10.0.22621
```

Screenboot:

![QQ截图20241117160728](https://github.com/user-attachments/assets/67db23af-9379-4bb2-a405-bb963bf1daed)

nojaf:
Hello, could you share your `rescript.json`?

BIYUEHU:
> Hello, could you share your `rescript.json`?

Sorry, I forgot. Here:

```json
{
  "name": "rescript-project-template",
  "sources": {
    "dir": "src",
    "subdirs": true
  },
  "package-specs": {
    "module": "commonjs",
    "in-source": true
  },
  "suffix": ".res.js",
  "bs-dependencies": ["@rescript/core"],
  "bsc-flags": ["-open RescriptCore"]
}
```

Everything was generated automatically through create-rescript-app, and I didn't make any changes.

fhammerschmidt:
I just created a create-rescript-app basic project with pnpm and do not have this issue. I hope it's not windows-specific.

I guess you also tried to restart the language server from VSCode's command palette?

Other than that what does the OUTPUT tab say when you select "ReScript Language Server" from the dropdown menu?

BIYUEHU:
> I just created a create-rescript-app basic project with pnpm and do not have this issue. I hope it's not windows-specific.
> 
[...]

PR #1120, opened 2025-08-27
Detect quoted path from ninja on Windows

nojaf:
I believe this fixes https://github.com/rescript-lang/rescript-vscode/issues/1056 on Windows.


zth (approved):
Awesome! Changelog? Then merge away!

nojaf:
I asked folks to try out the vsix and confirm it.
Will merge once fix is confirmed.

PR #1129, opened 2025-09-19
Use compiler-info.json to find runtime and bsc

nojaf:
Using the new information found in https://github.com/rescript-lang/rescript/pull/7889


nojaf:
Good questions:
- In case of the runtime (which is on the hot path), `findRuntimeCached` is used and that discovery logic is only ever ran once. Including the file read.
- For `bsc.exe`, it is only called when `projectRootState` is not available in `openedFile` so it will only read the json file there once.
- The old code is still in place as a fallback.

PR #1131, opened 2025-09-25
Use the correct build-schema.json for the installed compiler version

### LINKED ISSUE #1110: Build schema is always taken from the compiler's latest master
Only users of the latest version actually get correct autocompletions for their rescript.json, e.g. I now get 
"Property pinned-dependencies is not allowed." in a ReScript 11 project.

We should find some way for 
https://github.com/rescript-lang/rescript-vscode/blob/master/package.json#L54
to be aware of the version

fhammerschmidt:
Fixes #1110

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "vscode-languageclient": "8.1.0-next.5"
+        "vscode-languageclient": "9.0.1"
       }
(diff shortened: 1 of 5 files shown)


fhammerschmidt:
will try a different approach (the same as TS LSP)

cknitt:
@codex review

fhammerschmidt:
Of course it only works in the jsx property and not in the more advanced cases, adding proper tests now.

PR #1142, opened 2025-11-17
Externalize oxc-parser

nojaf:
## Problem

The extension fails to activate in the published version (v1.67.1) with the error:

```
TypeError: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received undefined
at createRequire (node:internal/modules/cjs/loader:1919:11)
```

This was introduced in commit e371ab2894c177752ea9b053d754bc071316be16 which added `oxc-parser` and `oxc-walker` dependencies for the paste-as-JSX feature.

## Root Cause

`oxc-parser` is a native module with platform-specific bindings that uses `createRequire(__filename)` internally to load the correct binding for the current platform. When esbuild bundles these modules, `__filename` and `__dirname` become `undefined`, causing `createRequire` to fail.

## Solution

Externalize `oxc-parser` in the esbuild bundle configuration so it is loaded as external dependencies at runtime instead of being bundled. This allows them to correctly resolve `__filename`/`__dirname` and load their native bindings.


tsnobip (approved):
nice catch!

PR #1154, opened 2025-11-29
fix: resolve symlinks when finding platform binaries

han-tyumi:
This fixes an issue where the extension fails to find ReScript platform binaries (like bsc.exe) when using package managers that use symlinked node_modules structures, such as Deno and pnpm.

The issue occurs because `path.join(rescriptDir, "..")` performs string manipulation rather than following the symlink. When `node_modules/rescript` is a symlink (e.g., to `.deno/rescript@12.0.0/node_modules/rescript`), the parent directory resolves to `node_modules/` instead of the actual parent where `@rescript/darwin-arm64` etc. are located.

Using `fs.realpathSync()` to resolve the symlink first ensures the correct parent directory is found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 > - :house: [Internal]
 > - :nail_care: [Polish]
 
+## [Unreleased]
+
+#### :nail_care: Polish
+
+- Resolve symlinks when finding platform binaries. https://github.com/rescript-lang/rescript-vscode/pull/1154
+
 ## 1.70.0
 
 #### :bug: Bug fix


han-tyumi:
Yes, I'm on version 1.70.0 of the extension.

You're right that when `compiler-info.json` exists, the extension uses `bsc_path` directly and everything works. My fix is for the fallback code path when `compiler-info.json` doesn't exist yet (e.g., fresh project before first build).

To reproduce, I reverted my fix locally, deleted `lib/bs/compiler-info.json`, and reloaded VS Code:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/han-tyumi/Code/javascript/rescript/gracefully-hooked/node_modules/@rescript/darwin-arm64/bin.js'
```

The extension tries to import from `node_modules/@rescript/darwin-arm64/bin.js`, but with Deno's symlinked structure the actual location is `node_modules/.deno/@rescript+darwin-arm64@12.0.0/node_modules/@rescript/darwin-arm64/bin.js`.

Since `node_modules/rescript` is a symlink (`-> .deno/rescript@12.0.0/node_modules/rescript`), `path.join(rescriptDir, "..")` performs string manipulation rather than resolving through the symlink, resulting in the wrong parent directory.

Using `fs.realpathSync(rescriptDir)` first resolves the symlink to its actual location, so the `..` traversal finds the correct directory. This only affects the fallback path—once the project is built and `compiler-info.json` exists, that path is used instead.

Issue #1155, opened 2025-12-01
The ReScript Language Server server crashed


taigua:
# Problem

```
ode:internal/modules/esm/load:188
    throw new ERR_UNSUPPORTED_ESM_URL_SCHEME(parsed, schemes);
          ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'e:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:188:11)
    at defaultLoad (node:internal/modules/esm/load:82:3)
    at ModuleLoader.load (node:internal/modules/esm/loader:815:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:594:31)
    at #createModuleJob (node:internal/modules/esm/loader:624:36)
    at #getJobFromResolveResult (node:internal/modules/esm/loader:343:34)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:311:41)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:664:25) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}

Node.js v22.20.0
[Error - 10:10:19] The ReScript Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

# Platform
OS: Windows 11
VSCode: 1.106.3
Rescript: 12.0
Rescript-Vscode: 1.70.0

# Reproduce Step

1. Installed NodeJS v24.11.1
2. Create a rescript project `hello` through `npm create rescript-app@latest`.  The selected template is `Vite` and the ReScript version is `12.0.0`.
3. Use `code hello` to open the project. (Rescript-Vscode v1.70.0 is already installed)
4. Open `App.res` in `src`.
5. The Rescript-Language-Server complains:
```
node:internal/modules/esm/load:188
    throw new ERR_UNSUPPORTED_ESM_URL_SCHEME(parsed, schemes);
          ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'e:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:188:11)
[...]

PR #1160, opened 2025-12-13
Fix rewatch lockfile detection on Windows

nojaf:
Refactor findProjectRootOfFile to use separate functions for file vs directory lookups, eliminating confusing boolean parameter and fixing the case where exact directory matches were incorrectly excluded.

This fixed https://github.com/rescript-lang/rescript-vscode/issues/1155 for me on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## [Unreleased]
 
+#### :bug: Bug fix
+
+- Fix rewatch lockfile detection on Windows. https://github.com/rescript-lang/rescript-vscode/pull/1160
+
 #### :nail_care: Polish
 
 - Resolve symlinks when finding platform binaries. https://github.com/rescript-lang/rescript-vscode/pull/1154


cristianoc:
@nojaf this is great -- much cleaner
Is this root finding documented somewhere? I have the impression that we have now settled on precise rules to determine monorepo structure, so it might be possible to document how a monorepo should be set up.

nojaf:
> Is this root finding documented somewhere? I have the impression that we have now settled on precise rules to determine monorepo structure, so it might be possible to document how a monorepo should be set up.

I would not say that we nailed down everything related to searching monorepo structures.
It is more like changes upon changes I think.
We have some detection but not on all levels:
- Incremental compile will check for a lock file in a potential parent folder.
- Runtime detection will take it into account as well.
- But in findBinary parts of the fallback logic doesn't take it into account.

For the runtime and binary, in a best case a `compiler-info.json` file was found that gives an accurate result of where these things are.

We also don't quite take anything npm/pnpm/bun specific into account of a monorepo.
Not saying that we should, but it is something we don't do.

PR #1162, opened 2025-12-17
Add LSP logger

nojaf:
Writing to the console does not show up in all LSP clients.
I was pretty blind in Zed when something did not work as expected.

Turned out my problem was that we do not override the default settings when the client provides them (see 7334e1e2a86bfdf06ac7da42b19e21525826d47e).

`configuration.incrementalTypechecking?.debugLogging` enables the highest log level.
This is to semantically close enough to what we have today. Existing config won't break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,12 @@
 #### :bug: Bug fix
 
 - Fix rewatch lockfile detection on Windows. https://github.com/rescript-lang/rescript-vscode/pull/1160
+- Override default `initialConfiguration` with user specific config. https://github.com/rescript-lang/rescript-vscode/pull/1162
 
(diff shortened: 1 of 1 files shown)


server/src/server.ts
```diff
@@ -26,9 +26,39 @@ import { assert } from "console";
...
+  );
+
+  const debugLoggingEnabled =
+    configuration.incrementalTypechecking?.debugLogging === true;
+
+  if (debugLoggingEnabled) {
+    // incrementalTypechecking.debugLogging overrides logLevel to retain legacy behavior
+    setLogLevel("log");
+    return;
+  }
```
zth: I don't think we need to retain legacy behavior here. I can't imagine anyone else but us using that property anyway.

PR #1171, opened 2026-01-04
fix Start Code Analysis: use correct binary paths for ReScript v12+

cristianoc:
More in style with the existing codebase.
Just add one more path to try, on top of all the others already tried.

### DIFF
--- a/client/src/utils.ts
+++ b/client/src/utils.ts
@@ -2,6 +2,8 @@ import * as path from "path";
 import * as fs from "fs";
 import * as os from "os";
 import { DocumentUri } from "vscode-languageclient";
+import * as semver from "semver";
+import { getBinaryPathLegacy } from "./utils-legacy";
 
 /*
  * Much of the code in here is duplicated from the server code.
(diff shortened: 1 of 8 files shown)


server/src/utils.ts
```diff
@@ -272,42 +330,31 @@ let findBinary = async (
...
   }
 
-  if (binaryPath != null && fs.existsSync(binaryPath)) {
-    return normalizePath(binaryPath);
-  } else {
+  // Top-level separation: v12+ or legacy
+  if (rescriptVersion == null) {
     return null;
   }
+  const isReScript12OrHigher = semver.gte(rescriptVersion, "12.0.0");
```
cknitt, line 348: Nit: capitalization is inconsistent here. We usually use "Rescript" in variable names in this repo, not "ReScript".

client/src/utils.ts
```diff
@@ -32,42 +34,110 @@ export function normalizePath(filePath: string | null): NormalizedPath | null {
...
+export const findReScriptVersion = (
+  projectRootPath: NormalizedPath | null,
+): string | null => {
+  if (projectRootPath == null) {
+    return null;
+  }
+  try {
+    const packageJsonPath = path.join(
+      projectRootPath,
+      "node_modules",
```
nojaf, line 56: This is already quite the assumption you make.
You expect the projectRootPath to have `node_modules`. It might be in a subfolder for all you know.

Also, something like Deno would even have an `node_modules` folder anymore.

Nor will everyone add rescript as regular dependency package.
Some people will add rescript as devDep and in isolated installs that folder won't even be there.

client/src/utils.ts
```diff
@@ -32,42 +34,110 @@ export function normalizePath(filePath: string | null): NormalizedPath | null {
...
+    "bin.js",
   );
 
+  if (!fs.existsSync(binJsPath)) {
+    return null;
+  }
+
+  // Read bin.js and extract the binary path
+  // bin.js exports binPaths object with paths to binaries
+  const binDir = path.join(
```
nojaf, line 97: Same remark, this is quite the assumption and does not always hold.

nojaf: Example path of where `bsc.exe` is located in my isolated install project:

```
{
  "bsc_path": "/Users/nojaf/Projects/intervallo/node_modules/.bun/@rescript+darwin-arm64@12.0.2/node_modules/@rescript/darwin-arm64/bin/bsc.exe",
}
```

server/src/utils-legacy.ts
```diff
@@ -0,0 +1,58 @@
...
+import * as c from "./constants";
+import { normalizePath, NormalizedPath } from "./utils";
+
+const fsAsync = fs.promises;
+
+/**
+ * Finds binaries for ReScript < 12 using the old path structure.
+ * Checks compiler-info.json first, then falls back to node_modules/rescript/${platformDir}/.
+ * NOTE: This preserves the original behavior exactly - do not add existence checks
+ * to the compiler-info.json branch as the original code returned immediately.
```
nojaf, line 17: **compiler-info.json** is **NOT** at all legacy.
It is a ReScript (Rewatch) 12 feature that has a significant higher accuracy to determine the bsc_exe value.

It gets written after a `rescript build` and has significant accurate information about the compiler and runtime. Respecting environment variables that can override these locations which is crucial for local testing.

https://github.com/rescript-lang/rescript/blob/66ba3f8d7769c9d21084f8e3e3acd1c622d56514/rewatch/src/build.rs#L410

server/src/utils.ts
```diff
@@ -211,18 +212,93 @@ export let getProjectFile = (
...
+  // Use realpathSync to resolve symlinks, which is necessary for package
+  // managers like Deno and pnpm that use symlinked node_modules structures.
+  const targetPackagePath = path.join(
+    fs.realpathSync(rescriptDir),
+    "..",
+    `@rescript/${target}/bin.js`,
+  );
+
+  try {
+    const { binPaths } = await import(targetPackagePath);
```
nojaf, line 242: Though this is correct, it does not take `RESCRIPT_BSC_EXE` into account:

https://github.com/rescript-lang/rescript/blob/66ba3f8d7769c9d21084f8e3e3acd1c622d56514/rewatch/src/helpers.rs#L364-L374

nojaf (changes requested):
I'm not going to claim what we have today is the crown jewel or can't be improved, but this PR makes too many assumptions and misses a lot of recent developments.

Makes me wonder whether we should add a `rescript binaries` command that outputs all the info as JSON to stdout. That would essentially be what `compiler-info.json` contains today, but without invoking a full build.

Can you elaborate what wasn't working for you? This feels more like a rewrite for the fun of it, presented as "v12 never worked so here is the fix." We fixed a lot over the summer, so I'm curious which setups failed and what exactly went wrong.

cristianoc:
> I'm not going to claim what we have today is the crown jewel or can't be improved, but this PR makes too many assumptions and misses a lot of recent developments.
> 
> Makes me wonder whether we should add a `rescript binaries` command that outputs all the info as JSON to stdout. That would essentially be what `compiler-info.json` contains today, but without invoking a full build.
> 
> Can you elaborate what wasn't working for you? This feels more like a rewrite for the fun of it, presented as "v12 never worked so here is the fix." We fixed a lot over the summer, so I'm curious which setups failed and what exactly went wrong.

The original intent is actually quite simple.
"Start Code Analyzer" does not find the correct binary path (and it looks for the wrong binary anyway), falls back into some "use what's included in the extension", then crashes. Instead of simply reporting that the binaries were not found.

The original intent is to at least eliminate the v12 vs other versions complexity. And fix code analyzer binary lookup.
Clearly the result simplifies too much at the moment.

Is there a spec anywhere of what binary resolution should look like? Is there a reason why it should be different than what the compiler does (if not, then the compiler code could be the spec). Even a few tiny example projects checked in that "should not break" could work as substitute for a proper spec.

Otherwise, "change the existing code as little as possible or something will break" is the only thing that can be done at the moment. Which is one way to do the PR if there's no alternative.

cristianoc:
moved to different PR for simplicty.

nojaf:
> Is there a spec anywhere of what binary resolution should look like?

Not really, this is mostly the consequence of situation x or y was not working.

The 
```js
    const targetPackagePath = path.join(
      fs.realpathSync(rescriptDir),
      "..",
      `@rescript/${target}/bin.js`,
    );
```
part was a recent addition.

If we had that sooner, we probably would not have done the `compiler-info.json` lookup.

PR #1174, opened 2026-01-14
Fix/monorepo root compilerlog watch

cristianoc:
Fix monorepo build detection: watch only root .compiler.log; run reanalyze from root

- move logic for finding binaries to be shared between server and client
- Server: restrict .compiler.log watching to ROOT/lib/bs/.compiler.log per workspace folder to avoid event floods from packages/** and node_modules/**.
- Client: Start Code Analysis now uses shared logic to locate the binary, and starts the analysis from outside "node_modules" where the binary is

### DIFF
--- a/client/src/commands/code_analysis.ts
+++ b/client/src/commands/code_analysis.ts
@@ -15,11 +15,12 @@ import {
   OutputChannel,
   StatusBarItem,
 } from "vscode";
+import { getBinaryPath, NormalizedPath, normalizePath } from "../utils";
 import {
-  findProjectRootOfFileInDir,
-  getBinaryPath,
-  NormalizedPath,
-} from "../utils";
+  findBinary,
+  findBinary as findSharedBinary,
+} from "../../../shared/src/findBinary";
+import { findProjectRootOfFile } from "../../../shared/src/projectRoots";
 
 export let statusBarItem = {
   setToStopText: (codeAnalysisRunningStatusBarItem: StatusBarItem) => {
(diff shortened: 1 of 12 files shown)


client/src/commands/code_analysis.ts
```diff
@@ -213,9 +213,14 @@ export const runCodeAnalysisWithReanalyze = (
   let currentDocument = window.activeTextEditor.document;
   let cwd = targetDir ?? path.dirname(currentDocument.uri.fsPath);
 
+  // Resolve the project root from `cwd` (which is the workspace root when code analysis is started),
+  // rather than from the currently-open file (which may be in a subpackage).
   let projectRootPath: NormalizedPath | null = findProjectRootOfFileInDir(
-    currentDocument.uri.fsPath,
+    path.join(cwd, "bsconfig.json"),
```
cknitt: Why bsconfig.json and not rescript.json / both?

cknitt: In extension.ts, you are using workspace.getWorkspaceFolder(currentDocument.uri) getting the workspace root.
Shouldn't this be unified/extracted?

server/src/server.ts
```diff
@@ -118,16 +118,28 @@ let findRescriptBinary = async (
 ): Promise<utils.NormalizedPath | null> => {
   if (
     config.extensionConfiguration.binaryPath != null &&
-    fs.existsSync(
-      path.join(config.extensionConfiguration.binaryPath, "rescript"),
-    )
+    (fs.existsSync(
+      path.join(config.extensionConfiguration.binaryPath, "rescript.exe"),
```
cknitt: This is weird - if I remember correctly, the binaries always had the .exe extension on all platforms.

client/src/commands/code_analysis.ts
cristianoc: Good catch will fix.

server/src/server.ts
cristianoc: Then it's simple and we can just always look for .exe.

client/src/commands/code_analysis.ts
cristianoc: Btw the "bsconfig.json" is likely just dead code and I can remove it, as in the projects I've tested on there wasn't any.
So it should rely entirely on finding the project root.
Now, whether the project root is the workspace folder is an entirely different question. I'm not convinced and will try to open the workspace from a subproject.

cristianoc: As expected, the workspace root is not what we want.
And the analysis is in the client. And the other monorepo-compatible logic for finding binaries is in the server.
Maybe there's a way to share files between the two.

cristianoc:
@cknitt changed the PR to move the logic which finds binaries to be shared between client and server.
We find the root of the project based on where the binary is found. So whatever folder the workspace is started from does not matter as long as it is inside the monorepo.

PR #1177, opened 2026-02-02
fix: wrap binary path resolution with fileURLToPath

jderochervlk:
The extension was failing to correctly resolve the path on Windows.
```
node:internal/modules/esm/load:188
    throw new ERR_UNSUPPORTED_ESM_URL_SCHEME(parsed, schemes);
          ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, node, and electron are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
```

Adding [fileURLToPath](https://nodejs.org/api/url.html#urlfileurltopathurl-options) uses node's built in way to normalize the file URL so it works across platforms.

> ...as well as ensuring a cross-platform valid absolute path string

### DIFF
--- a/shared/src/findBinary.ts
+++ b/shared/src/findBinary.ts
@@ -2,6 +2,7 @@ import * as fs from "fs";
 import * as fsAsync from "fs/promises";
 import * as path from "path";
 import * as semver from "semver";
+import * as url from "url";
 
 export type BinaryName =
   | "bsc.exe"
@@ -115,7 +116,7 @@ export const findBinary = async ({
       "..",
       `@rescript/${target}/bin.js`,
     );
-    const { binPaths } = await import(targetPackagePath);
+    const { binPaths } = await import(url.fileURLToPath(targetPackagePath));
 
     if (binary === "bsc.exe") {
       binaryPath = binPaths.bsc_exe;


jderochervlk:
I tested this out on Windows and the extension now works.

nojaf (approved):

zeeksphere:
Thanks for the fix! I had the same problem. But in my case it seems to be pathToFileURL, not fileURLToPath, that solves the problem. The issue persists with fileURLToPath.

davkats:
> Thanks for the fix! I had the same problem. But in my case it seems to be pathToFileURL, not fileURLToPath, that solves the problem. The issue persists with fileURLToPath.

Yeah, I agree.

Looking back at the forum thread, my guess is that `pathToFileURL()` and `fileURLToPath()` just got mixed up.

https://forum.rescript-lang.org/t/help-vscode-tooling-is-broken-for-me-on-windows/7112

There was a lot of trial and error during the investigation, so I'm not sure the whole code path was ever verified with logging. It's also possible that once a `compiler-info` file existed, the original error simply stopped happening by coincidence, making it look like this change was the fix.

That said, I think the idea was right. The path does need to be converted to a file URL.

PR #1178, opened 2026-02-02
fix: update globs to work on windows with WSL

jderochervlk:
This was mostly trial and error to figure out, but eventually I got it working.

Paths for WSL resolve differently so the simple glob pattern wasn't working.

Instead of setting `globPattern` as a string I changed them to define `baseUri` and `pattern` as separate values.

This only works when you have `incrementalTypechecking` enabled, but I like that anyway. We might want to document that for windows users.

## Before
https://github.com/user-attachments/assets/db9c1be6-10ca-4549-a43b-0ca9fd687b84

## After
https://github.com/user-attachments/assets/c92b8e83-6c7c-4df5-a68f-009f0be7378a

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -1528,20 +1528,27 @@ async function onMessage(msg: p.Message) {
             // Only watch the root compiler log for each workspace folder.
             // In monorepos, `**/lib/bs/.compiler.log` matches every package and dependency,
             // causing a burst of events per save.
-            globPattern: path.join(projectRootPath, c.compilerLogPartialPath),
+            globPattern: {
+              baseUri: utils.pathToURI(projectRootPath),
+              pattern: c.compilerLogPartialPath,
+            },
             kind: p.WatchKind.Change | p.WatchKind.Create | p.WatchKind.Delete,
           },
           {
-            globPattern: path.join(
-              projectRootPath,
-              "**",
-              c.buildNinjaPartialPath,
-            ),
+            // Watch ninja output
+            globPattern: {
+              baseUri: utils.pathToURI(projectRootPath),
+              pattern: path.join("**", c.buildNinjaPartialPath),
+            },
             kind: p.WatchKind.Change | p.WatchKind.Create | p.WatchKind.Delete,
           },
           {
-            globPattern: `${path.join(projectRootPath, "**", c.compilerDirPartialPath)}/**/*.{cmt,cmi}`,
-            kind: p.WatchKind.Change | p.WatchKind.Delete,
+            // Watch build artifacts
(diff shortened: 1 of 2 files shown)


server/src/server.ts
```diff
@@ -1528,23 +1528,43 @@ async function onMessage(msg: p.Message) {
...
+            },
             kind: p.WatchKind.Change | p.WatchKind.Create | p.WatchKind.Delete,
           },
           {
-            globPattern: path.join(
-              projectRootPath,
-              "**",
-              c.buildNinjaPartialPath,
-            ),
+            // Watch source files
```
nojaf: This is feels very wrong, these watchers are for files the user would not edit.
If this is your problem, the LSP is not picking up `textDocument/didChange` notifications I would say. If that is the case, that is the WSL problem you are having. It makes no sense to start having an extra watch on the source files.

Can you check if https://github.com/rescript-lang/rescript-vscode/blob/0326ae73e7a59e0985e5e7590b10d56665155d23/server/src/server.ts#L1593-L1611 is hit at all? If not, you need to find out why instead.

jderochervlk: I was able to remove the source files. Can you take another look?

cristianoc (approved):
Thanks for the fix.
Nit: there are more lines changed than needed, kind of makes it difficult to see the actual differences.

cristianoc:
Also, any thoughts to prevent regressions here?
Seems like it will very easily happen again as soon as the code is edited.

zth:
@cristianoc can the recent e2e test setup help here?

cristianoc:
> @cristianoc can the recent e2e test setup help here?

It can if tests work on windows. Anyone with a win machine can try to run this:
```shell
cd client
npm test
```

nojaf (approved):
Still works for me on Mac.
@jderochervlk can you add some entries in the CHANGELOG.md please.

jderochervlk:
On windows, no WSL, running `npm test` fails on master and on this branch.

```
[main 2026-02-04T00:20:14.166Z] Error: Error mutex already exists
    at qu.S (file:///F:/Dev/rescript-vscode/client/.vscode-test/vscode-win32-x64-archive-1.108.2/resources/app/out/main.js:142:18405)
```

jderochervlk:
> Thanks for the fix. Nit: there are more lines changed than needed, kind of makes it difficult to see the actual differences.

I removed some of the extra changes to minimize the difference.

jderochervlk:
> Still works for me on Mac. @jderochervlk can you add some entries in the CHANGELOG.md please.

Added a note to the changelog.

